PR #54, opened 2017-06-04
Optimize all the PNGs

balcsida:
4 MB saving


Rampastring:
Nice work. From a quick look I can't see a difference in image quality.

The reference resources from Dawn of the Tiberium Age (DTA) are updated every once in a while however, so it's likely that at least some of these savings will eventually be lost unless they're also incorporated into DTA. That being said, DTA could possibly benefit slightly from the space-saving as well.

@Bittah ^

balcsida:
Yaay, thank you!
Yes, I only did lossless optimization on the images.
I'm gonna check out that project, too. Are they hosting their code on GitHub, too?

Rampastring:
Dawn of the Tiberium Age is a mod of Tiberian Sun, and the main project that I originally developed this client for. Basically all the code for it is "here" on this repository. The resources (including the PNGs that you optimized) have been directly copied from DTA and are included in the repository just as an example of what the client needs to work.

http://www.moddb.com/mods/the-dawn-of-the-tiberium-age

Rampastring:
Bittah has included the optimized textures with Dawn of the Tiberium Age now, so these optimized reference resources won't be overwritten in the future if they're updated :)

PR #65, opened 2018-01-24
Changes to statistics handling for Ares mods & option to allow spectators on co-op based maps.

Starkku:
The three commits included in this pull request contain two separate improvements to the statistics handling, particularly for Yuri's Revenge mods that use Ares (which in context of this client primarily means Mental Omega) and one new feature for maps using the co-op game mode logic.

- **Removed recording of statistics for games where only one player was present** - Ares allows sandbox games to be played with just one player, and recording stats from these games seems counterproductive. This check is omitted for co-op maps due to possible scenarios where co-op logic is used for gamemodes that allow singleplayer, as they wouldn't get their statistics logged as a result as predetermined houses from co-op maps are not taken into consideration for this check.
- **Changed client to record & display the 'total number of objects built' statistic for MO instead of the unused 'economy' stat** - Since 'economy' statistic does not exist at all in Yuri's Revenge, it made sense to change this to the 'total number of objects built during game' statistic that is available. This only changes the labels displayed in statistics viewer and the statistics parsing from game's log files, no changes were made to writing or reading the database.
- **Added map property 'CoopAllowSpectators' to allow enabling of spectators on co-op maps** - It does make sense for spectators to be disabled for co-op missions. However, the co-op framework (with predetermined enemy & ally houses, briefings and so on) can just as easily be used for other types of gamemodes, including ones where allowing spectators is more justified. Which is why I made this customizable on per map basis. It still defaults to spectators disabled, however.

EDIT: Added another commit to include two YR hotkeys in the hotkey configuration that had previously been overlooked.

### DIFF
--- a/ClientCore/ClientConfiguration.cs
+++ b/ClientCore/ClientConfiguration.cs
@@ -411,6 +411,14 @@ namespace ClientCore
             get { return clientDefinitionsIni.GetIntValue(SETTINGS, "MapCellSizeY", 24); }
         }
 
+        public bool UseBuiltStatistic
+        {
+            get
+            {
+                return clientDefinitionsIni.GetBooleanValue(SETTINGS, "UseBuiltStatistic", false);
+            }
+        }
+
         public string[] GetThemeInfoFromIndex(int themeIndex)
         {
             string[] values = clientDefinitionsIni.GetStringValue("Themes", themeIndex.ToString(), ",").Split(',');
--- a/ClientCore/Statistics/GameParsers/DTAStatisticsParser.cs
+++ b/ClientCore/Statistics/GameParsers/DTAStatisticsParser.cs
@@ -26,14 +26,7 @@ namespace ClientCore.Statistics.GameParsers
         {
             this.fileName = fileName;
             this.loserString = loserString;
-            ParseStatistics(gamepath);
-        }
-
-        public void ParseStats(string gamepath, string fileName, string loserString, string economyString)
-        {
-            this.fileName = fileName;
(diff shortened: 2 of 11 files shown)


Rampastring:
Sounds good. I'll review the changes in more detail later on.

Starkku:
There were significant oversights in implementation of CoopAllowSpectators that I should've caught in first place (particularly how spectators wouldn't count towards maximum number of players allowed in the game which combined with the preplaced houses caused some rather weird results), which could only really be addressed by adding some rather unintuitive safeguards. My verdict is that it is not worth salvaging, hence I've reverted those changes.

Rampastring:
The code looks good aside from this part:

``if (ClientConfiguration.Instance.LocalGame.ToUpper() == "MO")
+            {
+                strLblEconomy = "BUILT";
+                strLblAvgEconomy = "Avg. number of objects built:";
+            }``

Checking for specifically MO makes this not work for other YR projects, so it'd be better to add a key to ClientDefinitions.ini that controls whether to use the Built or Economy stat instead.

There's also an issue in that the Economy stat in TS is always between 0 and 100, so it is written as a byte, meaning it can't hold values over 255. I can fix that, though.

Starkku:
I amended the decision on which statistic to parse, track and display to be based on ClientDefinitions boolean key 'UseBuiltStatistic'. Feel free to change the name of the key if you so wish.

I forgot to check the data type in the statistics database itself. Since the database parsing seems to already support multiple versions of the file format, amending that shouldn't be difficult but I'll leave it to you I suppose.

PR #72, opened 2018-05-03
A small random side selectors fix & couple of other changes.

Starkku:
- Due to an oversight, random side selectors were not properly unselected if changing to a map where they were unavailable. Now the interactions with disallowed side logic should work properly under all circumstances.
- Allow changing the text displayed on the item in client resolution dropdown that is designated as recommended resolution. By default this displays '(recommended)', which is way too big to fit in the dropdown with the larger font used in most YR mods - changing it to '(default)' leaves enough space. Issues like this can potentially crop up elsewhere in the client as well, and treating them all as special cases like this might not be the best course of action. Would probably warrant further discussion on how to approach the said problems, or if even should at all.
- A simple fix to cleaning up renderer config files that would have previously raised an error if directory names were included in config file name paths in renderer configuration.

### DIFF
--- a/ClientCore/ClientConfiguration.cs
+++ b/ClientCore/ClientConfiguration.cs
@@ -573,6 +573,14 @@ namespace ClientCore
             get { return clientDefinitionsIni.GetBooleanValue(SETTINGS, "DisplayPlayerCountInTopBar", false); }
         }
 
+        public string ClientDefaultResolutionText
+        {
+            get
+            {
+                return clientDefinitionsIni.GetStringValue(SETTINGS, "ClientDefaultResolutionText", "(recommended)");
+            }
+        }
+
         /// <summary>
         /// Returns the name of the game executable file that is used on
         /// Linux and macOS.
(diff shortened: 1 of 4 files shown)


Rampastring:
> Allow changing the text displayed on the item in client resolution dropdown that is designated as recommended resolution. By default this displays '(recommended)', which is way too big to fit in the dropdown with the larger font used in most YR mods - changing it to '(default)' leaves enough space. Issues like this can potentially crop up elsewhere in the client as well, and treating them all as special cases like this might not be the best course of action. Would probably warrant further discussion on how to approach the said problems, or if even should at all.

Adding them all as keys to ClientDefinitions.ini would indeed be too much. In this case I think one approach would be creating a new control that extends XNADropDown and has a field that describes the recommended resolution options as well as the recommended label. When adding items to the control, the control would check if the item matches any of recommended resolution options, and if it does, then the specified recommended label is added to the end of the item. The recommended resolution option and the recommended label could then be changed with INI code.

PR #74, opened 2018-05-27
Add first pass at checks for cheat engine usage

GrantBartlett:
Kills game instance if Cheat Engine is in use.

### DIFF
--- a/DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetLobby.cs
+++ b/DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetLobby.cs
@@ -15,7 +15,10 @@ using Rampastring.XNAUI;
 using Rampastring.XNAUI.XNAControls;
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
+using System.Diagnostics;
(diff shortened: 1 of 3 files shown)


GrantBartlett:
@Rampastring See latest commit and let me know what you think?

Rampastring:
Start the checker in CnCNetLobby.ConnectionManager_WelcomeMessageReceived instead of CnCNetLobby.Initialize.

Then stop the checker in CnCNetLobby.ConnectionManager_Disconnected.

Make sure that the service is re-usable so that it works properly if you connect, disconnect and then connect again.

Also btw, you should call `Dispose()` on all the Process object instances that you get from the `GetProcesses()` call, otherwise you might cause a memory leak.

GrantBartlett:
@Rampastring like so?

```
            foreach (Process process in processlist)
            {
                if (process.ProcessName.Contains("cheatengine") ||
                    process.MainWindowTitle.ToLower().Contains("cheat engine") ||
                    process.MainWindowHandle.ToString().ToLower().Contains("cheat engine")
                )
                {
                    KillGameInstance();
                }

                process.Dispose();
            }
```

PR #76, opened 2018-05-30
New announcement feature

GrantBartlett:
**Show a stand out announcement message** 

Fetches an announcement from a remote URL and shows the text in the CnCNet lobby upon joining.
- Inserts into the CnCNet Lobby as a new message.
- Customisation over the URL and colour of text with the following keys:
```
CnCNetAnnouncementURL=
CnCNetAnnouncementColorRGB=
```
Preview attached
announcement


tomsons26:
Isn't this a Message of the day not a annoncement

DXMainClient/Domain/Multiplayer/CnCNet/CnCNetAnnouncement.cs
```diff
@@ -0,0 +1,45 @@
...
+            {
+                Logger.Log("Error fetching Announcement text from remote url: " 
+                    + ClientCore.ClientConfiguration.Instance.CnCNetAnnouncementURL 
+                    + " Error: " + ex.Message);
+            }
+
+            return message;
+        }
+
+        public string[] Color { get; private set; }
```
Rampastring: This string array could be replaced with Microsoft.Xna.Framework.Color. Then you could use
AssetLoader.GetColorFromString for parsing the color.

You could also parse the color from the announcement by including a separator character between the actual announcement and the color data. Doing this instead of coding the color locally would make the feature much more flexible.

Rampastring: This also needs to be multi-threaded, because this can significantly increase client load time due to the web request during startup. Multi-threading is hard to do right though, so I could make that change if you're not very familiar with it.

GrantBartlett:
@Rampastring Yes I'd be more happy if you could add in the multi-threading for this, I can then see how it's added for future use.

PR #90, opened 2019-02-17
Bugfix for a client freeze introduced by the recent changes to map listing in game lobby.

Starkku:
Fixes a problem introduced in commit aa6faa8002575d7b0ee4132ca7ee0bdc937ea704 included in PR #87 that would sometimes cause the client to freeze when searching for maps.

Long story short, when doing code refactoring to get rid of the 'unnecessary' counter variable, I somehow completely failed to register that just flat-out changing it to use the loop counter completely changes how the whole thing works when searching maps, as it skips pretty much the entire loop if map's name doesn't contain the search string and previously the counter would only get incremented if it didn't skip. Trying to get rid of the counter variable turned out to be a mostly moot point too, since it ended up needing an extra one regardless.

I did keep the for-loop implementation rather than revert back, however, since as far as I know that's bit more efficient on lists and arrays. I don't know if there are concerns for code style / consistency that'd take priority over that, though.

### DIFF
--- a/DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
+++ b/DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
@@ -403,13 +403,17 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
             lbMapList.SelectedIndex = -1;
 
             int mapIndex = -1;
+            int skippedMapsCount = 0;
 
             for (int i = 0; i < GameMode.Maps.Count; i++)
             {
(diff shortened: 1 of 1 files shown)


Rampastring:
`I did keep the for-loop implementation rather than revert back, however, since as far as I know that's bit more efficient on lists and arrays. I don't know if there are concerns for code style / consistency that'd take priority over that, though.`

The client in general isn't so heavy that performance would be a concern, so it's best to go for code readability and ease of long-term maintenance. Only exceptions to this are rare cases where inefficient code is ran often and could result in a worse user experience or unacceptably high resource use (such as checks done every single frame, render loop / drawing code etc.).

PR #97, opened 2019-06-02
Auto-ready checkbox for multiplayer game lobbies.

Starkku:
This feature is based on a suggestion from players. Essentially an invisible-by-default (must be made visible through the UI INI system) checkbox that if toggled on, marks the player as with 'auto-ready' status which means that their actual ready status is not revoked under any circumstances until after the checkbox is unticked (unticking itself won't remove it, though). This piggybacks on existing messaging between players and game host, but since LAN game variant breaks compatibility and CnCNet one would otherwise be susceptible to sync problems, the protocol versions have been incremented as well. Default position of the checkbox is equivalent to where the 'Lock Game' button appears for game hosts.

Isn't marked as draft, but still open for discussion / scrutiny.

### DIFF
--- a/ClientCore/ProgramConstants.cs
+++ b/ClientCore/ProgramConstants.cs
@@ -20,8 +20,8 @@ namespace ClientCore
 
         public const string QRES_EXECUTABLE = "qres.dat";
 
-        public const string CNCNET_PROTOCOL_REVISION = "R6";
-        public const string LAN_PROTOCOL_REVISION = "RL4";
+        public const string CNCNET_PROTOCOL_REVISION = "R7";
+        public const string LAN_PROTOCOL_REVISION = "RL5";
         public const int LAN_PORT = 1234;
         public const int LAN_INGAME_PORT = 1234;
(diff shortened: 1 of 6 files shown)


Starkku:
I am not sure if clutter or lack of space is a concern here, since intention was for the checkbox to take the place of the 'Lock Game' button for non-host players. Hosts do not need this feature, after all.

The variable is necessary if the checkbox visibility is to be kept customizable through INI files, as otherwise the 'initial state' is overridden by host/non-host toggles. If not being able to hide it at all is acceptable, then that step could be eliminated entirely.

Rampastring:
Looking more at it, it should be possible to get rid of the attribute by calling `chkAutoReady.GetAttributes(ThemeIni)` in the non-host code part in `MultiplayerGameLobby.Refresh`, and just disabling it in the host-part of the code.

There is also a critical bug in the current code; the check-box is made invisible, but it is not disabled, meaning it still accepts user input. Use `chkAutoReady.Disable()` rather than `chkAutoReady.Enabled = false`.

PR #105, opened 2019-08-30
Various client additions

Metadorius:
- Added an optional lookup "array" for explicit declaration of country codes according to game INI section
- Added an optional spectator country override
- Added optional separation of multiplayer maps and modes INI files

Look in commit descriptions for further details.


Metadorius:
Reverted MPModes.ini and MPMaps.ini separation since Starkku reported it silently breaks some stuff with game mode aliases.

PR #112, opened 2019-10-26
Add "Change Tunnel" button to CnCNet game lobby

Metadorius:
Only available at CnCNet lobby. Does the same as the chat command.

### DIFF
--- a/DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
+++ b/DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
@@ -95,6 +95,7 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
         private TunnelHandler tunnelHandler;
         private CnCNetTunnel tunnel;
         private TunnelSelectionWindow tunnelSelectionWindow;
+        private XNAClientButton btnChangeTunnel;
 
         private Channel channel;
         private CnCNetManager connectionManager;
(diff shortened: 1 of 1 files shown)


DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
```diff
@@ -179,11 +196,15 @@ private void GameBroadcastTimer_TimeElapsed(object sender, EventArgs e)
             {
                 RandomSeed = new Random().Next();
                 RefreshMapSelectionUI();
+                btnChangeTunnel.Enabled = true;
+                btnChangeTunnel.Visible = true;
```
Rampastring: Use `btnChangeTunnel.Enable()` instead of setting the `Enabled` and `Visible` attributes manually.

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
```diff
@@ -179,11 +196,15 @@ private void GameBroadcastTimer_TimeElapsed(object sender, EventArgs e)
...
             }
             else
             {
                 channel.ChannelModesChanged += Channel_ChannelModesChanged;
                 AIPlayers.Clear();
+                btnChangeTunnel.Enabled = false;
+                btnChangeTunnel.Visible = false;
```
Rampastring: Use `btnChangeTunnel.Disable()`.

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
```diff
@@ -151,9 +152,25 @@ public override void Initialize()
...
+            btnChangeTunnel.Visible = false;
+
+            AddChild(btnChangeTunnel);
             WindowManager.AddAndInitializeControl(gameBroadcastTimer);
         }
 
+        private void BtnChangeTunnel_LeftClick(object sender, EventArgs e)
+        {
+            ShowTunnelSelectionWindow("Select tunnel server:");
+        }
```
Rampastring: Since this function is so simple, it could be better to embed its functionality in a lambda expression on line 160. Typing it with an expression body is another option.

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
```diff
@@ -151,9 +152,25 @@ public override void Initialize()
...
 
+            btnChangeTunnel = new XNAClientButton(WindowManager);
+            btnChangeTunnel.Name = "btnChangeTunnel";
+            btnChangeTunnel.ClientRectangle = new Rectangle(btnLeaveGame.Right - btnLeaveGame.Width - 145,
+                btnLeaveGame.Y, 133, 23);
+            btnChangeTunnel.Text = "Change Tunnel";
+            btnChangeTunnel.LeftClick += BtnChangeTunnel_LeftClick;
+
+            btnChangeTunnel.Enabled = false;
+            btnChangeTunnel.Visible = false;
```
Rampastring: Isn't making it disabled by default like this rather pointless since `SetUp` is always called when the game room is opened?

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
Metadorius: Literally the function next to the one I added (`GameBroadcastTimer_TimeElapsed`) is also very simple and not declared as lambda or via expression body, so I think what you said should not be done in this PR as it would make code not match the style of the rest.

Metadorius:
Resolved most of the issues in a commit as requested.

Metadorius:
Addressed the expression body usage issue along with giving expression bodies to some more methods in scope of the class.

PR #114, opened 2019-11-30
Misc. fixes & improvements

Starkku:
### Deprecated DisableUpdaterOptions & DisableComponentOptions. The tabs in options window are now automatically disabled (rather than removed) when appropriate.
These options were mostly just clutter in ClientDefinitions.ini, as the conditions under which you'd want the tabs in options disabled are usually the same and easily determined. It also now just makes the tabs unselectable rather than outright removing them (which can still be done anyway through the RemoveTabIndex# keys), however there is no visual indicator for this as of current - I have created a PR that addresses it however: Rampastring/Rampastring.XNAUI#3.

### Output info on physical memory alongside other hardware info
Allows people going through the log file to more easily narrow down potential edge-cases when debugging problems (f.ex machines with extremely low amount of system memory).

### Added names to controls missing them in UpdateQueryWindow
Exposing them to the UI INI system, allowing better customization of the window (primarily to address problems arising from using a larger font).

###  Added a message to game lobby chatbox that reminds players of how to view chatbox commands.
This might not be a desirable change, but the amount of people seemingly unaware of the chatbox commands existing, especially after the potentially very useful 'Change Tunnel' one was added (I am aware that #112 also touches this issue) made me think it might be necessary. Can obviously be tweaked/reworded or reverted entirely if deemed appropriate.

###  Correctly generate & check file hashes for GlobalCode.ini and map extra INI files
There was a discrepancy between the GlobalCode.ini filepath between FileHashCalculator and where it actually parses and merges the INI file, I opted to change former to match latter. INI files declared in maps' ExtraINIFilename were also not included in hash check at all. I recognize that a more straightforward option would have been to apply a similar directory-wide check to INI\Map Code as to INI\Game Options, but I opted to not change the existing course of logic and merely add on top to it. There might be performance differences between the two approaches, but I could not reliably gauge this myself.

### MORelease solution configuration now includes project configuration for DXMainClient that correctly defines the MO conditional compilation symbol
FileHashCalculator built-in list of files to check specific to the MORelease build of the client wasn't previously used at all. This addresses that problem.

###  Deprecated ClientDefaultResolutionText, replaced with a dropdown control that allows setting a preferred item index and a string label to be displayed next to the preferred item's text on the dropdown.
Back when ClientDefaultResolutionText option was implemented, this was mentioned as a more preferrable solution to the problem. Could also potentially be used elsewhere in the future, although current implementation does not allow customizing the default item through UI INI system, only the string displayed.


DXClient.sln
```diff
@@ -18,168 +18,168 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "DTAConfig", "DTAConfig\DTAC
...
 		{D517317D-8EA8-4225-AF9E-5385D8A7F047}.TSRelease|XNAFramework.ActiveCfg = TSRelease|XNAFramework
 		{D517317D-8EA8-4225-AF9E-5385D8A7F047}.TSRelease|XNAFramework.Build.0 = TIRelease|XNAFramework
-		{D517317D-8EA8-4225-AF9E-5385D8A7F047}.YRRelease|WindowsGL.ActiveCfg = YRRelease|WindowsGL
-		{D517317D-8EA8-4225-AF9E-5385D8A7F047}.YRRelease|WindowsGL.Build.0 = YRRelease|WindowsGL
 		{D517317D-8EA8-4225-AF9E-5385D8A7F047}.YRRelease|SharpDX.ActiveCfg = YRRelease|SharpDX
 		{D517317D-8EA8-4225-AF9E-5385D8A7F047}.YRRelease|SharpDX.Build.0 = YRRelease|SharpDX
+		{D517317D-8EA8-4225-AF9E-5385D8A7F047}.YRRelease|WindowsGL.ActiveCfg = YRRelease|WindowsGL
+		{D517317D-8EA8-4225-AF9E-5385D8A7F047}.YRRelease|WindowsGL.Build.0 = YRRelease|WindowsGL
 		{D517317D-8EA8-4225-AF9E-5385D8A7F047}.YRRelease|XNAFramework.ActiveCfg = YRRelease|XNAFramework
 		{D517317D-8EA8-4225-AF9E-5385D8A7F047}.YRRelease|XNAFramework.Build.0 = YRRelease|XNAFramework
```
Rampastring: The reordering here is weird, but as far as I can see, no information / configuration choices have been actually lost.

DXMainClient/DXGUI/Generic/UpdateQueryWindow.cs
```diff
@@ -56,11 +56,13 @@ public override void Initialize()
             btnYes.ClientRectangle = new Rectangle(12, 110, 75, 23);
             btnYes.Text = "Yes";
             btnYes.LeftClick += BtnYes_LeftClick;
+            btnYes.Name = "btnYes";
 
             var btnNo = new XNAClientButton(WindowManager);
             btnNo.ClientRectangle = new Rectangle(164, 110, 75, 23);
             btnNo.Text = "No";
             btnNo.LeftClick += BtnNo_LeftClick;
+            btnNo.Name = "btnNo";
```
Rampastring: Make use of `nameof`. For example, `btnNo.Name = nameof(btnNo)`.

(I'm aware the old client code doesn't make use of the operator, but that's no reason not to use it in new code.)

DXMainClient/DXMainClient.csproj
```diff
@@ -166,6 +166,34 @@
...
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'MORelease|XNAFramework'">
+    <OutputPath>bin\XNAFramework\Release\</OutputPath>
+    <DefineConstants>TRACE;XNA MO</DefineConstants>
+    <Optimize>true</Optimize>
+    <DebugType>pdbonly</DebugType>
+    <PlatformTarget>x86</PlatformTarget>
+    <ErrorReport>prompt</ErrorReport>
+    <CodeAnalysisRuleSet>MinimumRecommendedRules.ruleset</CodeAnalysisRuleSet>
+  </PropertyGroup>
```
Rampastring: These seem to have some differences compared to the other profiles.

1) WindowsGL MORelease lacks the `Optimize` setting.
2) The others don't have CodeAnalysisRuleSet
3) The others don't have Prefer32Bit
4) XNAFramework MORelease lacks `<TargetFrameworkVersion>v4.0</TargetFrameworkVersion>`

The first and fourth seem like mistakes, I'm not sure about the effects of 2) and 3).

DXMainClient/Online/FileHashCalculator.cs
```diff
@@ -94,11 +94,26 @@ public void CalculateHashes(List<GameMode> gameModes)
...
                     fh.INIHashes = AddToStringIfFileExists(fh.INIHashes, "INI\\Map Code\\" + gameMode.Name + ".ini");
-                    Logger.Log("Hash for INI\\Map Code\\" + gameMode.Name + ".ini :" +
+                    Logger.Log("Hash for INI\\Map Code\\" + gameMode.Name + ".ini: " +
                         Utilities.CalculateSHA1ForFile(ProgramConstants.GamePath + "INI\\Map Code\\" + gameMode.Name + ".ini"));
+
+                    foreach (Map map in gameMode.Maps)
+                    {
+                        if (!string.IsNullOrEmpty(map.ExtraININame))
+                            extraINIFilenames.Add("INI\\Map Code\\" + map.ExtraININame);
+                    }
```
Rampastring: I'm slightly worried about the performance hit this could have if a map exists in multiple game modes and the same extra INI filename gets added to the list multiple times. It's probably a non-issue in all current mods, but if some YR mods with tons of maps started making heavy use of the extra INIs, it could cause delays.

Starkku: I now changed this to a similar blanket file check as already existed for INI\Game Options so the runtime won't scale with game mode / map count. 

I am not sure why INI\Map Code check is omitted on YR release, so I didn't change that and left the GlobalCode.ini special case for YR only as a consequence.

I also took the liberty of adjusting how both the map code / game option INI hash log messages show filenames, previously it was the full filename, now it is relative to the game root like everything else.

DXMainClient/DXGUI/Generic/UpdateQueryWindow.cs
Starkku: UpdateQueryWindow has been changed to use this. Will keep in mind for future.

DXMainClient/DXMainClient.csproj
Starkku: 1. The `Optimize` setting is actually missing from the WindowsGL Release configuration as well, which is where it was copied from here. I have now corrected this.
2, 3 & 4: Changed to match the rest.

DXClient.sln
Starkku: Most likely a consequence of not doing this entirely by hand and letting Visual Studio add the new configurations.

PR #121, opened 2020-02-09
Custom settings overhaul

Metadorius:
Implement `RestartRequired` INI parameter support for FileSettingCheckBox. Also changed and fixed option panels to support the functionality. This branch will possibly have further improvements to custom file settings subsystem.


Metadorius:
Example of the new custom settings file controls:

Dropdown
Notice window

in `OptionsWindow.ini`:
```ini
[DisplayOptionsPanelExtraControls]
0=chkEnableCannonTracers:CustomSettingFileCheckBox
1=labelTest:XNALabel
2=ddTest:CustomSettingFileDropDown

[chkEnableCannonTracers]
Location=285,113
Text=Use Cannon Tracers
EnabledFile0=Resources\ECache91.mix,MIX\ECache91.mix
DefaultValue=true

[labelTest]
Text=Test Label:
Location=285,137

[ddTest]
Location=285,161
Size=97,21
Items=No Test,Test 1,Test 2
DefaultValue=0
CheckFilePresence=yes
ResetUnselectableItem=yes
ForceApplyUnselectableItem=no
RestartRequired=true

Item1File0=test1.txt,result.txt
Item2File0=test2.txt,result.txt
```

ClientCore/Settings/UserINISettings.cs
```diff
@@ -9,6 +9,13 @@ public class UserINISettings
     {
         private static UserINISettings _instance;
 
+        private const string VIDEO = "Video";
+        private const string MULTIPLAYER = "MultiPlayer";
+        private const string OPTIONS = "Options";
+        private const string AUDIO = "Audio";
+        private const string CUSTOM_SETTINGS = "Custom Settings";
```
Rampastring, line 16: Let's follow the WW convention of not having spaces in INI section identifiers.

DTAConfig/CustomSettings/CustomSettingFileCheckBox.cs
```diff
@@ -0,0 +1,161 @@
...
+                }
+
+                FileOperationOptions options = default;
+                if (parts.Length >= 3)
+                    Enum.TryParse(parts[2], out options);
+
+                disabledFiles.Add(new FileSourceDestinationInfo(parts[0], parts[1], options));
+
+                i++;
+            }
```
Rampastring, line 86: The code inside these two loops is identical aside from the target list and the name of the key. It should be moved into its own method unless there's a very good reason not to do so.

DTAConfig/CustomSettings/CustomSettingFileCheckBox.cs
```diff
@@ -0,0 +1,161 @@
...
+        private List<FileSourceDestinationInfo> disabledFiles = new List<FileSourceDestinationInfo>();
+
+        private bool EnabledFilesComplete => enabledFiles.All(f => File.Exists(f.SourcePath));
+        private bool DisabledFilesComplete => disabledFiles.All(f => File.Exists(f.SourcePath));
+
+        private bool defaultValue;
+        private bool originalState;
+        private bool restartRequired;
+        private bool checkFilePresence;
+        private bool resetUnselectableItem;
```
Rampastring, line 29: Some comments here on what `checkFilePresence` and `resetUnselectableItem` precisely do (or are meant to do) would be useful.

DTAConfig/CustomSettings/CustomSettingFileCheckBox.cs
```diff
@@ -0,0 +1,161 @@
...
+                enabledFiles.ForEach(f => f.Apply());
+            }
+            else if (!Checked && canBeUnchecked)
+            {
+                enabledFiles.ForEach(f => f.Revert());
+                disabledFiles.ForEach(f => f.Apply());
+            }
+            else // undefined state, the checkbox is bork, don't continue
+            {
+                return false;
```
Rampastring, line 153: Are there any situations where this could happen? If it happens, should the situation be fixable through the client or is the user left to mess with the files by themself?

This looks like something that should likely be logged.

DTAConfig/CustomSettings/FileSettingCheckBox.cs
```diff
@@ -0,0 +1,100 @@
...
+        {
+            if (reversed != Checked)
+                files.ForEach(f => f.Apply());
+            else
+                files.ForEach(f => f.Revert());
+
+            return restartRequired && (Checked != originalState);
+        }
+    }
+}
```
Rampastring, line 100: To prevent duplicate code, this class should be implemented as a `CustomSettingFileCheckBox`. Maybe implementing this as a derived class would be easy.

Is there a reason why the new class is called `CustomSettingFileCheckBox` instead of following the naming convention of this class with `CustomFileSettingCheckBox` btw?

DTAConfig/CustomSettings/FileSourceDestinationInfo.cs
```diff
@@ -0,0 +1,115 @@
...
+                    }
+
+                    break;
+
+                case FileOperationOptions.DontOverwrite:
+                    if (File.Exists(ProgramConstants.GamePath + DestinationPath))
+                        break;
+
+                    File.Copy(ProgramConstants.GamePath + SourcePath,
+                        ProgramConstants.GamePath + DestinationPath, false);
```
Rampastring, line 47: Looking at the examples above and below, it'd be better to revert the `if` condition here and have the operation in brackets, keeping the code consistent and leaving only one `break` statement and exit point from this case.

DTAConfig/CustomSettings/FileSourceDestinationInfo.cs
```diff
@@ -0,0 +1,115 @@
...
+                case FileOperationOptions.MoveFile:
+                    File.Move(ProgramConstants.GamePath + SourcePath,
+                        ProgramConstants.GamePath + DestinationPath);
+                    break;
+
+                case FileOperationOptions.AlwaysOverwrite:
+                default:
+                    File.Copy(ProgramConstants.GamePath + SourcePath,
+                        ProgramConstants.GamePath + DestinationPath, true);
+                    break;
```
Rampastring, line 78: Handle `AlwaysOverwrite` on its own and throw an exception in the `default` case. If someone adds a new `FileOperationOption` and forgets to actually implement it here, it's better to throw than have it confusingly work the same as `AlwaysOverwrite`.

DTAConfig/OptionPanels/XNAOptionsPanel.cs
```diff
@@ -54,35 +55,59 @@ protected override void ParseExtraControls(IniFile iniFile, string sectionName)
...
+                {
+                    case FileSettingCheckBox controlAsFileSettingCheckBox:
+                        customSettings.Add(controlAsFileSettingCheckBox);
+                        break;
+                    case CustomSettingFileCheckBox controlAsAdvancedFileCheckBox:
+                        customSettings.Add(controlAsAdvancedFileCheckBox);
+                        break;
+                    case CustomSettingFileDropDown controlAsFileSettingDropDown:
+                        customSettings.Add(controlAsFileSettingDropDown);
+                        break;
```
Rampastring, line 68: Why not check for the `ICustomSetting` interface instead?

Rampastring (changes requested):
Looks good, but there's some things to improve.

DTAConfig/CustomSettings/FileSettingCheckBox.cs
Metadorius: > Is there a reason why the new class is called `CustomSettingFileCheckBox` instead of following the naming convention of this class with `CustomFileSettingCheckBox` btw?

Yes, the interface for all custom settings is named as `ICustomSetting` so I followed that and prefixed the class name like so because there could be non-file settings and because of that I thought that it would make more sense to operate with custom settings, not settings that allow for custom files.

DTAConfig/OptionPanels/XNAOptionsPanel.cs
Metadorius: As far as I remember it refused to work if I checked for `ICustomSetting`, but I'll check again just to be sure.

DTAConfig/CustomSettings/CustomSettingFileCheckBox.cs
Metadorius: In current implementation this is would be either mod config maker's fault when there would be no valid states/items chosen by `Refresh()` when the option is reset or the installation is broken.

DTAConfig/OptionPanels/XNAOptionsPanel.cs
Metadorius: Hm, seems that I remembered it wrong. Works now.

DTAConfig/CustomSettings/FileSettingCheckBox.cs
Metadorius: Also this won't be possible without breaking compatibility with older builds and/or user configurations.

Metadorius:
`CheckFilePresence` is now `CheckAvailablility`, `ResetUnselectableItem` is now `ResetUnavailableValue`. Also broken compatibility with older userconfigs due to a changed user INI section name. `ForceApplyUnselectableItem` is removed and so is an option for FSDInfo to move files instead of copying as I see it as impossible to implement that nicely with the current code.

DTAConfig/CustomSettings/FileSettingCheckBox.cs
Starkku: Latest implementation of `FileSettingCheckBox` achieves full backwards compatibility with the old client configuration whilst using a minimal amount of duplicate or bespoke code & meshing in with the new / altered code in `XNAOptionsPanel`. There is room for improvement, such as looking into the possibility of implementing `FileSettingsCheckBox` as a class inherited from `CustomSettingFileCheckBox`, but there's probably further discussion warranted for the aim of such move (is the aim to keep 1:1 backwards compatibility or offer a transitional system to the new INI setting-based system?) that, along with any further code changes regarding this, are probably best deferred to a later date and outside this PR.

Starkku (approved):
Latest changes look OK and address all remaining issues, apart from one that I have touched upon separately in [this comment](https://github.com/CnCNet/xna-cncnet-client/pull/121#discussion_r551019888).

PR #127, opened 2020-02-26
Fix FinalSun issue with non-ASCII path

SadPencil:
FinalSun/FinalAlert regard the INI file as ANSI, but the INI file DTA client generated is UTF-8.
This caused a problem that FinalSun can't be started on a path with non-ASCII path.

Example:
`C:\DTAClient`
Okay.
`C:\DTAî`
European characters. Failed.
`C:\DTA测试`
Chinese characters. Failed.

This patch fixed the problem.
Also, see the comment on the code, so that this problem can be fixed in a nicer way.


Metadorius:
You can take a look and contribute to improve `IniFile` here:
https://github.com/Rampastring/Rampastring.Tools

SadPencil:
@Metadorius A new version of Rampastring.Tools and the FinalSun.cs is submitted.
The corresponding code is at https://github.com/Rampastring/Rampastring.Tools/pull/1

SadPencil:
Both DTA's `FinalSun_DTA.exe` and MO's `FinalAlert2MO.exe` have this issue. This pull request solves problems on this two map editors. And I believe other map editors based on FinalSun have the same issue.

DXMainClient/Domain/FinalSunSettings.cs
```diff
@@ -20,8 +21,12 @@ public static void WriteFinalSunIni()
                 {
                     Logger.Log("FinalSun settings file exists.");
 
-                    IniFile iniFile = new IniFile(ProgramConstants.GamePath + finalSunIniPath);
-
+                    IniFile iniFile = new IniFile();
+                    iniFile.FileName = ProgramConstants.GamePath + finalSunIniPath;
+                    // Be sure to use .NET Framework instead of .NET Core as the latter doesn't support ANSI          
+                    iniFile.Encoding = System.Text.Encoding.Default;
```
Rampastring: Using `System.Text.Encoding.Default` is problematic. It can be different on different computers, and isn't always (at least the same) ANSI. https://docs.microsoft.com/en-us/dotnet/api/system.text.encoding.default?view=netframework-4.8

Instead use `Encoding.GetEncoding(1252)`, it reliably returns the Windows-1252 encoding that is used by the game and probably also FinalSun / FinalAlert. Make that 1252 a constant.

SadPencil: No. The encoding Final Sun/Final Alert uses is not 1252. It is exactly the system's default non-Unicode code page, which can be changed on Control Panel, denoted as `ANSI`. The default setting depends on the system's language, such as 1033 for en-US, 2052 for zh-CN. 

`System.Text.Encoding.Default` is(are) exactly the same encoding(s) the Final Sun/Final Alert uses, and it's the most common way on .NET to specify ANSI encoding.

PR #128, opened 2020-02-26
Fix High DPI issue

SadPencil:
When High DPI is set, e.g. 125%
both DTA client and the game will be resized to a bad size, which sometimes causes the game failed to be started.

Example:
1920x1080 monitor with 125% DPI settings, which is the default setting on a 15.6 inch laptop.
The game will be resized to 1536x768 and then failed to start because the graphic card doesn't support this strange resolution.

By changing the manifest file this issue can be solved.

Note: although user can change the compatibility settings on Windows, but that is not a nice way. The issue should be solved directly, i.e this pull request.


SadPencil:
Also, I find that `Screen.PrimaryScreen.Bounds.Width` is used in the code.
This value requires DPI awareness to be turned on, i.e. just like this pull request,
otherwise, can't get screen resolution through this value.
So, as a conclusion, this pull request is necessary.

Metadorius:
Rampastring is quite busy with real life now, but he will get to that sooner or later, just need to wait some time for him to review the changes.

PR #129, opened 2020-02-27
Add Windows Firewall settings + Restart UAC

SadPencil:
No one likes the Windows Firewall pop-up window.
This pull request implement a new feature to check and set up firewall settings to avoid the firewall pop-up window.
If the firewall settings are there, nothing happens.
Otherwise, the client will ask for user would he/she want to restart the client and gain administrator privileges. Once user clicked 'yes', the client will restart and set up firewall settings.
XNAMessagebox is used to ask user.

A new key in `ClientDefination.ini` is set as `BypassWindowsFirewall` with default value `true`.

Tested on DTA and MO.

### DIFF
--- a/DXMainClient/DXGUI/Generic/MainMenu.cs
+++ b/DXMainClient/DXGUI/Generic/MainMenu.cs
@@ -440,7 +440,8 @@ namespace DTAClient.DXGUI.Generic
 
             PlayMusic();
 
-            BypassWindowsFirewall();
+            if (ClientConfiguration.Instance.BypassWindowsFirewall)
+                BypassWindowsFirewall();
 
             if (!ClientConfiguration.Instance.ModMode)
             {
--- a/DXMainClient/DXGUI/Generic/MainMenu.cs
+++ b/DXMainClient/DXGUI/Generic/MainMenu.cs
@@ -14,6 +14,7 @@ using Rampastring.Tools;
 using Rampastring.XNAUI;
 using Rampastring.XNAUI.XNAControls;
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Threading;
 using Updater;
@@ -439,6 +440,8 @@ namespace DTAClient.DXGUI.Generic
 
             PlayMusic();
 
+            BypassWindowsFirewall();
+
             if (!ClientConfiguration.Instance.ModMode)
             {
                 if (UserINISettings.Instance.CheckForUpdates)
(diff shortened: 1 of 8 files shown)


Metadorius:
A screenshot of the window would be handy :wink:

SadPencil:
@Metadorius 
https://i.postimg.cc/xCLLbwZ4/001.png
https://i.postimg.cc/tJ2Z3pN2/002.png

SadPencil:
Update:
If two XNAMessageBoxes are shown, i.e. the first run messagebox and firewall messagebox at the same time, there is chaos.
So currently, fallback to the common MessageBox which shows and blocks the thread.

DXMainClient/DXGUI/Generic/MainMenu.cs
```diff
@@ -351,28 +352,26 @@ private void CheckIfFirstRun()
...
+                xnaMessageBox = XNAMessageBox.ShowYesNoDialog(WindowManager, "Initial Installation",
                     string.Format("You have just installed {0}." + Environment.NewLine +
                     "It's highly recommended that you configure your settings before playing." +
                     Environment.NewLine + "Do you want to configure them now?", ClientConfiguration.Instance.LocalGame));
-                firstRunMessageBox.YesClickedAction = FirstRunMessageBox_YesClicked;
-                firstRunMessageBox.NoClickedAction = FirstRunMessageBox_NoClicked;
+                xnaMessageBox.YesClickedAction = (XNAMessageBox messageBox) =>
+                {
+                    optionsWindow.Open();
+                };
```
Rampastring: No need to have brackets in a single-line lambda function.

Rampastring:
Tbh I don't think it's the client's job to change firewall permissions.

DTA already shows the compatibility fix and first run dialog at the same time on the first run on a new system, and it works okay.

SadPencil:
Yes. I don't know where the dialog should be put on. 

It's true that the client should not change firewall settings. I just find the two Firewall pop-up windows are annoying, one of which players can only pause the game, return to Windows and deal with the Firewall pop-up window. So client asking the user's permission and then set-up Windows Firewall can reduce this annoying thing.

However, my pull request is indeed not an elegant way because it creates a new in-game pop-up window. I don't know what's the best way.

SadPencil:
There are more optimizations can be done. See https://github.com/CnCNet/xna-cncnet-client/issues/132

PR #134, opened 2020-03-04
Changed hotkey for Extras from E to X

MahBoiDeveloper:
"E" was a common hotkey for Map Editor and Extras so i edit hotkey for Extras.


Rampastring:
Made redundant by #138

PR #135, opened 2020-03-05
Fixed a crash when no item in side dropdown was selected

Metadorius:
The issue was fixed by adding a null check to presence update method.


Rampastring:
Fixed by https://github.com/CnCNet/xna-cncnet-client/commit/2ebe432ff98a7d98b9e90359bac5d5c34b652eff

PR #138, opened 2020-03-27
Fixes to various UI-related problems

Starkku:
- Custom component button text now retains file size information even if download is canceled or unsuccessful.
- Options button in top bar will now be disabled if options window is open. This comes with minor graphical issues due to pressing the Options button altering its own AllowClick property. https://github.com/Rampastring/Rampastring.XNAUI/pull/5 contains a proposed fix to these graphical issues.
- Top bar buttons and main menu hotkeys are now disabled while updater window is open. Disabling hotkeys is not currently possible without just outright unsetting them, so there is potentially some room for improvement here. Also renders #134 incompatible and obsolete.
- Top bar status text won't be changed on disconnect from CnCNet if you have LAN (game) lobby currently open.
- Tooltips are now added as child controls of the most immediate XNAWindow-derived parent of their master control, instead of as a root-level controls. This fixes issues with tooltip display in certain places of the client such as the main menu. Also made XNALinkButton support multi-line tooltip text like checkboxes do.
- Saved client sound volume setting is now immediately restored upon exiting options without saving, instead of leaving the current sound volume active until the settings are refreshed by starting game or opening options window again.
- Main menu is initialized and extra controls are parsed and added before the MainMenuDarkeningPanel to ensure that none of the main menu elements can be interacted with through the darkening panel.

### DIFF
--- a/DXMainClient/DXGUI/Generic/MainMenu.cs
+++ b/DXMainClient/DXGUI/Generic/MainMenu.cs
@@ -100,16 +100,16 @@ namespace DTAClient.DXGUI.Generic
         private CancellationTokenSource cncnetPlayerCountCancellationSource;
 
         // Main Menu Buttons
-        XNAClientButton btnNewCampaign;
-        XNAClientButton btnLoadGame;
-        XNAClientButton btnSkirmish;
-        XNAClientButton btnCnCNet;
-        XNAClientButton btnLan;
-        XNAClientButton btnOptions;
-        XNAClientButton btnMapEditor;
-        XNAClientButton btnStatistics;
-        XNAClientButton btnCredits;
-        XNAClientButton btnExtras;
+        private XNAClientButton btnNewCampaign;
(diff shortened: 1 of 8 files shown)


DXMainClient/DXGUI/Generic/MainMenu.cs
```diff
@@ -87,6 +99,18 @@ class MainMenu : XNAWindow, ISwitchable
...
+        XNAClientButton btnNewCampaign;
+        XNAClientButton btnLoadGame;
+        XNAClientButton btnSkirmish;
+        XNAClientButton btnCnCNet;
+        XNAClientButton btnLan;
+        XNAClientButton btnOptions;
+        XNAClientButton btnMapEditor;
+        XNAClientButton btnStatistics;
+        XNAClientButton btnCredits;
+        XNAClientButton btnExtras;
```
Rampastring: Add the `private` visibility modifier to these.

Issue #88, opened 2018-12-28
Auto close inactive game rooms


GrantBartlett:
See it a lot, hosts go AFK people join the game and then get frustrated because they've waited only for the host to not be there. 

It would be good after a period of inactivity that the host's game is either closed, or removed.

PR #145, opened 2020-04-18
Warn inactive game hosts and eventually remove them from game listings

GrantBartlett:
## Removing inactive hosts from game lobbies
In relation to this: https://github.com/CnCNet/xna-cncnet-client/issues/88

image

- Checks the host hasn't been active for X seconds (default 180)
- Continues to check on past that time until X seconds (default 240) until the host is automatically removed from their game.  
- Disposes of checks when in game, and reactivates when back

## Improvements
- Wanted to check that everything is being disposed of correctly
- Probably needs to be listening to the global mouse handler (if that even exists, e.g window.addEventListener("mousemove",.. in JS land).
- Should probably also listen to Keyboard events


DXMainClient/DXMainClient.csproj
```diff
@@ -40,7 +40,7 @@
     <DebugType>full</DebugType>
     <Optimize>false</Optimize>
     <OutputPath>bin\Windows\Debug\</OutputPath>
-    <DefineConstants>TRACE;DEBUG;DTA</DefineConstants>
+    <DefineConstants>TRACE;DEBUG;YR</DefineConstants>
```
Rampastring, line 43: This looks like an accidental change.

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
```diff
@@ -213,6 +218,45 @@ public override void Initialize()
...
+            if (activityHandler != null)
+            {
+                activityHandler.Reset();
+            }
+        }
+
+        private void onPromptRemovalFromGame(object sender, EventArgs e)
+        {
+            LeaveGame();
+        }
```
Rampastring, line 253: Methods start with a capital letter. Also, you can make use of the expression body `=>` here.

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
```diff
@@ -213,6 +218,45 @@ public override void Initialize()
...
+
+        private void onPromptInactiveMessage(object sender, EventArgs e)
```
Rampastring, line 255: Start this with a capital letter too.

DXMainClient/DXGUI/Generic/ActivityHandler.cs
```diff
@@ -0,0 +1,100 @@
...
+        void Start();
+        void Reset();
+        void Dispose();
+        event EventHandler PromptInactiveMessage;
+        event EventHandler PromptRemoval;
+    }
+
+    public class ActivityHandler : IActivityHandler
+    {
+        private Timer activityTimer;
```
Rampastring, line 22: Instead of using the native timer, it'd be better to make this into a `XNATimerControl`, similar to the game broadcast timer. You could similarly use lifetime initialization for it, so there wouldn't be need to dispose of it and re-initialize it each time a game is hosted.

If you still need it to be disposable, implement the `IDisposable` interface for the class.

Rampastring (changes requested):
The overall code looks pretty good, but there's some things to fix.

GrantBartlett:
See here: https://github.com/CnCNet/xna-cncnet-client/pull/362 (closing this one)

PR #147, opened 2020-05-09
Allow to disable sides per game mode

Metadorius:
As requested by @Crimsonum you can now use `DisallowedPlayerSides` in game mode definitions. The usage is the same as with coop INI tag. Need investigation on how it would work with random selectors.

### DIFF
--- a/DXMainClient/Domain/Multiplayer/PlayerHouseInfo.cs
+++ b/DXMainClient/Domain/Multiplayer/PlayerHouseInfo.cs
@@ -19,7 +19,6 @@ namespace DTAClient.Domain.Multiplayer
         /// and randomizes it if necessary.
         /// </summary>
         /// <param name="pInfo">The PlayerInfo of the player.</param>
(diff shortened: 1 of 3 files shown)


DXMainClient/Domain/Multiplayer/PlayerHouseInfo.cs
```diff
@@ -23,7 +23,7 @@ public class PlayerHouseInfo
         /// <param name="sideCount">The number of sides in the game.</param>
         /// <param name="random">Random number generator.</param>
         /// <param name="disallowedSideArray">A bool array that determines which side indexes are disallowed by game options.</param>
-        public void RandomizeSide(PlayerInfo pInfo, Map map, int sideCount, Random random,
```
Starkku, line 26: Should probably also remove `map` param from method documentation.

Starkku:
None of the changes present alter interactions with random side selectors and they continue to function as before. Code overall looks fine too, just that one minor nitpick about documentation.

PR #150, opened 2020-05-18
Invite to game feature

yodathecoda:
When hosting a game, adds an Invite option to the context menu in friends list.  Allows "invite-only" games by creating a passworded game, as accepting an invitation bypasses password entry.


yodathecoda:
Note this implementation allows invite of anyone, even those who don't have the host on their friends list.  Happy to rework if that's not desired.

DXMainClient/DXGUI/Multiplayer/CnCNet/PrivateMessagingWindow.cs
```diff
@@ -18,6 +18,11 @@ namespace DTAClient.DXGUI.Multiplayer.CnCNet
 {
     internal class PrivateMessagingWindow : XNAWindow, ISwitchable
     {
+        // these are used by the "invite to game" feature in the
+        // context menu and are kept up-to-date by the lobby
+        public string inviteGameName;
+        public string inviteGamePassword;
```
Metadorius, line 24: Have not looked into that part of the code much yet, isn't it possible to directly check if the user is in game lobby?

DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetLobby.cs
```diff
@@ -1011,6 +1072,70 @@ private void ConnectionManager_WelcomeMessageReceived(object sender, EventArgs e
...
+                // note this is not reached for the rejection case
+                connectionManager.SendCustomMessage(new QueuedMessage("PRIVMSG " + sender + " :\u0001" +
+                    ProgramConstants.GAME_INVITATION_FAILED_CTCP_COMMAND + "\u0001",
+                    QueuedMessageType.CHAT_MESSAGE, 0));
+
+                return;
+            }
+
+            var inviteMessageBox = XNAMessageBox.ShowYesNoDialog(WindowManager, "Invitation received",
+                sender + " wants you to join their game (" + roomName + ").  Accept?");
```
Metadorius, line 1115: This will get annoying quickly but I've been informed that you're already working on a rework for that based on private messages notifications.

Screenshot I received from @dkeetonx

This one is pretty nice, my suggestion is to tweak margins so that it stays uniform for all elements and implement in that way that it would be reusable for more notifications of different types that could be implemented later (f.ex. "Ask to Join" notification), kind of like `XNAMessageBox.ShowYesNoDialog` is made.

DXMainClient/DXGUI/Multiplayer/CnCNet/PrivateMessagingWindow.cs
yodathecoda: Agree this would be nicer but didn't find a way to do this without making PrivateMessagingWindow aware of CnCNetLobby, which seems wrong to me.  Will have another look into this.

Metadorius:
Another concern - how does this work with locked games?

yodathecoda:
This implementation still displays the Invite option for the host when the game is locked, and the invitation is sent but silently rejected at the receiving end (see HandleGameInviteCommand and CanJoinGameByIndex).  At the host, a generic message box is displayed (see HandleGameInvitationFailedNotification).

DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetLobby.cs
yodathecoda: I've made the new ChoiceNotificationBox class (shows at top-left of the screen), mostly by copying code from PrivateMessageNotificationBox.  Considered adding a common NotificationBox base class for both but decided against as no controls are shared and pretty much all the functions are slightly different (PrivateMessageNotificationBox is used as single-instance, where ChoiceNotificationBox is stacked infinitely).

Invitation

yodathecoda: The width of the thing is a bit obnoxious but it looks like we don't have word-wrapping so I want to support both long player names and - given it's generic - long other "choice texts".

DXMainClient/DXGUI/Multiplayer/CnCNet/PrivateMessagingWindow.cs
yodathecoda: Could well be a nice way of doing this with some more investigation but I haven't found one that seems worth the churn.  Happy to look into this more in the future if it's a showstopper.

DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetLobby.cs
```diff
@@ -610,21 +628,59 @@ private void BtnJoinGame_LeftClick(object sender, EventArgs e)
 
         private void LbGameList_DoubleLeftClick(object sender, EventArgs e)
         {
+            JoinGameByIndex(lbGameList.SelectedIndex, string.Empty);
```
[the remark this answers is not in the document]
yodathecoda: Cool, latest commit has hopefully taken your approach here.

DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetLobby.cs
yodathecoda: The popup dialog that advises the host of invitation failure has been retained.  Reasoning is the host will want to know ASAP and obviously when an invitation has failed so they can do something about it.

Invitation failed

Metadorius: Good job, nice one with having them on the left, I thought about that but didn't write that idea. I would streamline the code for both anyways so that it would be kind of a generic notification framework for anything needed, I think it would be more logical for PM notifications to also stack, but you do you. Might be the feature for a separate pull request.

You could save space by doing something like this:

```
            GAME INVITE            
    From: username here        
    [Accept]            [Reject]    
```

yodathecoda: Restyled as specified, except kept the Ignore button instead of Reject as Reject may imply that the host will be notified of the rejection (which they won't).

yodathecoda: One more alternative if preferred for clarity:

`            GAME INVITE            
    Join fonger's game?       
    [Yes]                   [No]   `

yodathecoda: After discussion in discord, latest commit implements this notification: 
image

DXMainClient/DXGUI/Multiplayer/CnCNet/ChoiceNotificationBox.cs
```diff
@@ -0,0 +1,189 @@
...
+            AddChild(lblSender);
+
+            lblChoiceText = new XNALabel(WindowManager);
+            lblChoiceText.Name = "lblChoiceText";
+            lblChoiceText.FontIndex = 1;
+            lblChoiceText.ClientRectangle = new Rectangle(12, lblSender.Bottom + 6, 0, 0);
+            lblChoiceText.Text = "What do you want to do?";
+            AddChild(lblChoiceText);
+
+            affirmativeButton = new XNAButton(WindowManager);
```
Metadorius, line 81: Use `XNAClientButton` instead, this will save you from manually loading the texture.

DXMainClient/DXGUI/Multiplayer/CnCNet/ChoiceNotificationBox.cs
```diff
@@ -0,0 +1,189 @@
...
+            affirmativeButton.ClientRectangle = new Rectangle(ClientRectangle.Left + 8, lblChoiceText.Bottom + 6, 75, 23);
+            affirmativeButton.IdleTexture = AssetLoader.LoadTexture("75pxbtn.png");
+            affirmativeButton.HoverTexture = AssetLoader.LoadTexture("75pxbtn_c.png");
+            affirmativeButton.HoverSoundEffect = new EnhancedSoundEffect("button.wav");
+            affirmativeButton.Name = "affirmativeButton";
+            affirmativeButton.Text = "Yes";
+            affirmativeButton.LeftClick += AffirmativeButton_LeftClick;
+            AddChild(affirmativeButton);
+
+            negativeButton = new XNAButton(WindowManager);
```
Metadorius, line 92: Ditto.

DXMainClient/DXGUI/Multiplayer/CnCNet/ChoiceNotificationBox.cs
```diff
@@ -0,0 +1,189 @@
...
+            negativeButton.LeftClick += NegativeButton_LeftClick;
+            AddChild(negativeButton);
+
+            base.Initialize();
+        }
+
+        public void Show(string headerText, Texture2D gameIcon, string sender, string choiceText, string affirmativeText, string negativeText, int timeout)
+        {
+            Visible = true;
+            Enabled = true;
```
Metadorius, line 109: Those two lines can and should be replaced with call of `Enable()` method for this object.

DXMainClient/DXGUI/Multiplayer/CnCNet/ChoiceNotificationBox.cs
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        public void Hide()
+        {
+            isDown = false;
+            locationY = -Height;
+            ClientRectangle = new Rectangle(X, (int)locationY,
+                Width, Height);
+            Visible = false;
+            Enabled = false;
```
Metadorius, line 140: Ditto with `Disable()` method.

DXMainClient/DXGUI/Multiplayer/CnCNet/ChoiceNotificationBox.cs
```diff
@@ -0,0 +1,189 @@
...
+                while (lblChoiceText.Width > Width)
+                {
+                    lblChoiceText.Text = lblChoiceText.Text.Remove(lblChoiceText.Text.Length - 1);
+                }
+            }
+
+            downTime = TimeSpan.Zero;
+            isDown = true;
+
+            downTimeWaitTime = TimeSpan.FromSeconds(timeout);
```
Metadorius, line 130: Do those choice notif-s really need to time out in all cases? My suggestion is to make that optional (maybe set `timeout = 0` as a default argument value and don't check for wait time, so the notif. stays indefinitely).

DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetLobby.cs
```diff
@@ -1111,14 +1108,25 @@ private void HandleGameInviteCommand(string sender, string argumentsString)
...
+
+            WindowManager.AddAndInitializeControl(gameInviteChoiceBox);
+
+            // show the invitation at top left for 60 seconds
+            gameInviteChoiceBox.Show(
+                "GAME INVITATION",
+                GetUserTexture(sender),
+                sender,
+                "Join " + gameName + "?",
+                "Yes", "No", 60);
```
Metadorius, line 1121: In my opinion game invitations don't have to expire at all. F. ex. if the user is AFK, he wouldn't even know he was invited in the first place.

yodathecoda:
Reworked in line with @Metadorius comments above.  Invitations now remain indefinitely until acted upon or the related game is closed.  Only one invitation from a channel/user combination is now displayed at once.  Invitations can now be sent to and accepted by users in other game lobbies.  A "failed to join" message is now displayed in the event the game cannot be joined after pressing Yes to the invitation:

image

The "failed to invite" dialog has also been reworded following user feedback during testing:

image

yodathecoda:
Added option to only receive game invitations from friends:

image

This option defaults to off.  Tweaked popup dialog wording per below:

image

image

Rampastring:
One suggestion, without checking the code yet: if you invite a lot of people, the "invitation failed" pop-up might get annoying. Wouldn't it be better to display the error message in the game lobby's chat box instead? As a yellow warning message like all other error messages.

yodathecoda:
Agreed, can change in a few days once back at my PC.

yodathecoda:
> One suggestion, without checking the code yet: if you invite a lot of people, the "invitation failed" pop-up might get annoying. Wouldn't it be better to display the error message in the game lobby's chat box instead? As a yellow warning message like all other error messages.

Done:
image

DXMainClient/DXGUI/Multiplayer/CnCNet/ChoiceNotificationBox.cs
```diff
@@ -0,0 +1,193 @@
...
+{
+    /// <summary>
+    /// A box that allows users to make a choice,
+    /// top-left of the game window.
+    /// </summary>
+    public class ChoiceNotificationBox : XNAPanel
+    {
+        const double DOWN_TIME_WAIT_SECONDS = 4.0;
+        const double DOWN_MOVEMENT_RATE = 2.0;
+        const double UP_MOVEMENT_RATE = 2.0;
```
Rampastring, line 19: Mark these constants and private fields as explicitly private. While not all code in the client does it yet because there's a lot of outdated code around, it's good to apply higher standards to new code.

DXMainClient/DXGUI/Multiplayer/CnCNet/ChoiceNotificationBox.cs
```diff
@@ -0,0 +1,193 @@
...
+
+        bool isDown = false;
+
+        const int boxHeight = 101;
+
+        double locationY = -boxHeight;
+
+        public override void Initialize()
+        {
+            Name = "ChoiceNotificationBox";
```
Rampastring, line 48: Use `nameof` (in this case, `Name = nameof(ChoiceNotificationBox)`). The same applies to all child controls.

DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetLobby.cs
```diff
@@ -577,13 +590,23 @@ private void GameLoadingLobby_GameLeft(object sender, EventArgs e)
...
         private void GameLobby_GameLeft(object sender, EventArgs e)
         {
             topBar.SwitchToSecondary();
             isInGameRoom = false;
             SetLogOutButtonText();
+
+            // keep the friends window up to date so it can disable the Invite option
+            pmWindow.inviteChannelName = string.Empty;
+            pmWindow.inviteGameName = string.Empty;
+            pmWindow.inviteChannelPassword = string.Empty;
```
Rampastring, line 609: This is duplicate code with lines 594 to 597, transfer them into a method. Preferably the method should be in `PrivateMessageWindow` because its fields / properties are the ones being modified.

DXMainClient/DXGUI/Multiplayer/CnCNet/PrivateMessagingWindow.cs
```diff
@@ -18,6 +18,12 @@ namespace DTAClient.DXGUI.Multiplayer.CnCNet
 {
     internal class PrivateMessagingWindow : XNAWindow, ISwitchable
     {
+        // these are used by the "invite to game" feature in the
+        // context menu and are kept up-to-date by the lobby
+        public string inviteChannelName;
+        public string inviteGameName;
+        public string inviteChannelPassword;
```
Rampastring, line 25: Public fields and properties start with a capital letter. Also, turn them into properties with getters and setters (for example `public string InviteChannelName { get; set; }`. Also see the note related to these that I left for the CnCNet lobby.

Rampastring (changes requested):
The code is good and well documented for the most part, good job. Fix these and it should be mergeable.

DXMainClient/DXGUI/Multiplayer/CnCNet/PrivateMessagingWindow.cs
yodathecoda: Have made these private and permitted access via set/clear methods only as this seems to address both comments pretty cleanly.

DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetLobby.cs
yodathecoda: Added new SetInviteChannelInfo and ClearInviteChannelInfo methods to PrivateMessageWindow.

DXMainClient/DXGUI/Multiplayer/CnCNet/ChoiceNotificationBox.cs
yodathecoda: Done.

DXMainClient/DXGUI/Multiplayer/CnCNet/ChoiceNotificationBox.cs
yodathecoda: Done.

Rampastring (approved):

PR #152, opened 2020-05-20
Add IRC SSL server support

fredrikhr:
* Extends the Server structure to also store a `UseSSL` boolean which defaults to `false`.
* After TCP connection to the server is established, attempts to wrap the TCP connection stream into an SSL stream and authenticates the server.
* The Server certificate is validated against the OS-default certificate validation routine.


Metadorius:
@fredrikhr have you checked if XNA build compiles and works OK?

fredrikhr:
@Metadorius sorry, I am still working on this. I have encountered some issues with enforcing newer TLS versions. I am marking this as a draft while I try to figure it out

PR #161, opened 2020-06-05
Implement game filter text box

Metadorius:
`tbGameSearch` in CnCNet lobby filters by:
- game room name (partial match);
- map name (partial match);
- game mode name (full match);
- player present (full match).

Game search textbox

Searching by name


hesto2:
This was reported from AEFNET on discord, they don't have a github account so I'm putting it on here for them instead. 
image
The dropdown is positioned incorrectly it appears

Metadorius:
> This was reported from AEFNET on discord, they don't have a github account so I'm putting it on here for them instead. 
> image
> The dropdown is positioned incorrectly it appears

This is an issue of YR client config and not of this pull request. It isn't designed to accommodate for those changes yet.

DXMainClient/DXGUI/Multiplayer/GameListBox.cs
```diff
@@ -32,13 +32,12 @@ public class GameListBox : XNAListBox
...
 
-        public double GameLifetime
-        {
-            get { return _gameLifetime; }
-            set { _gameLifetime = value; }
-        }
+        /// <summary>
+        /// A predicate for setting a filter expression for displayed games.
+        /// </summary>
+        public Predicate<GenericHostedGame> GameMatchesFilter { get; set; }
```
Rampastring, line 40: Since this is never modified once assigned, nor is there any need to be able to access it from the outside, I think it'd be better to make it private and readonly and deliver it in the constructor.

Metadorius: `readonly` is not available in this case, made it private though.

Rampastring: @Metadorius You can make it readonly (assignable only from the constructor) by removing the `set` accessor entirely.

PR #162, opened 2020-06-05
Improve player nickname validation on startup & on saving settings

Starkku:
Unifies the validation of player nickname on client startup and when saving settings. Also fixes an error with saving settings where it previously wouldn't save the cleaned up nickname to settings INI, instead using raw text from the player name textbox.

This validation now uses same list of allowed characters as the name check upon attempt to connect to CnCNet, but it won't check for offensive words nor it will remove starting dashes / numbers. If such functionality is desirable, it could be added but I didn't deem it worthwhile.

### DIFF
--- a/ClientCore/CnCNet5/NameValidator.cs
+++ b/ClientCore/CnCNet5/NameValidator.cs
@@ -5,6 +5,8 @@ namespace ClientCore.CnCNet5
 {
     public static class NameValidator
     {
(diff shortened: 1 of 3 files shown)


Rampastring:
The only issue I see with this is that some singleplayer-only players might want to use special characters (like `öäå`) in their nickname. While not allowed online due to IRC limitations, they work fine in Skirmish.

Starkku:
Made the check for offline nicknames less strict, it now only removes characters that actually cause problems when used.

PR #166, opened 2020-07-20
Rework of map preview image extraction

Starkku:
- Added code to handle extraction of map preview images to the client, removed the `MapThumbnailExtractor.dll` dependency.
  - Has additional safeguards not present in old library code that should prevent it from crashing with corrupted or invalid preview data. Specifically, it checks beforehand if the LZO block that it is about to decompress exceeds the bounds of source data array or if the estimated uncompressed data would exceed the bounds of the destination data array.
- Moved the remaining PreviewPack checks from Map class to the new `MapPreviewExtractor.ExtractMapPreview` static method alongside all the other checks, and removed the blanket try-catch around the call to it as it is most likely unnecessary with all the checks present in `MapPreviewExtractor` itself. If it turns out not, reinstate or add additional checks.


DXMainClient/Domain/Multiplayer/Map.cs
```diff
@@ -466,33 +465,12 @@ public Texture2D LoadPreviewTexture()
...
-                    Logger.Log(mapIni.FileName + " - no [PreviewPack] exists");
-                    return AssetLoader.CreateTexture(Color.Black, 10, 10);
-                }
-                    
-                if (mapIni.GetStringValue("PreviewPack", "1", string.Empty) ==
-                    "yAsAIAXQ5PDQ5PDQ6JQATAEE6PDQ4PDI4JgBTAFEAkgAJyAATAG0AydEAEABpAJIA0wBVA")
-                {
-                    Logger.Log(mapIni.FileName + " - Hidden preview detected - not extracting.");
+                if (preview != null)
+                    return AssetLoader.TextureFromImage(preview);
```
Rampastring, line 471: `AssetLoader.TextureFromImage` will return null on failure, should handle that case and create the dummy 10x10 black texture. Although the existing checks seem to make sure that the extracted bitmap is correct, I don't want to take chances if some kind of a corner-case is found where the bitmap -> Texture2D conversion fails and someone could then crash the client through that.

PR #169, opened 2020-07-26
Implement blocking users from PM window

Metadorius:
Users can now block other users using PM window list. Since you can't block a user who is offline due to how IRC works the option is grayed out on offline users from friend list.

image

can't block offline users

### DIFF
--- a/DXMainClient/DXGUI/Multiplayer/CnCNet/PrivateMessagingWindow.cs
+++ b/DXMainClient/DXGUI/Multiplayer/CnCNet/PrivateMessagingWindow.cs
@@ -1,4 +1,4 @@
-using ClientCore;
+﻿using ClientCore;
 using ClientCore.CnCNet5;
 using ClientGUI;
 using DTAClient.Online;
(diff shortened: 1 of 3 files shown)


DXMainClient/DXGUI/Multiplayer/CnCNet/PrivateMessagingWindow.cs
```diff
@@ -339,6 +335,17 @@ private void LbUserList_RightClick(object sender, EventArgs e)
...
+            if ((bool)lbUserList.SelectedItem.Tag)
+            {
+                IRCUser iu = connectionManager.UserList.Find(u => u.Name == lbUserList.SelectedItem.Text);
+                playerContextMenu.Items[1].Text = cncnetUserData.IsIgnored(iu.Ident) ? "Unblock" : "Block";
+                playerContextMenu.Items[1].Selectable = true;
+            }
+            else
+            {
+                playerContextMenu.Items[1].Text = "Toggle Block";
+                playerContextMenu.Items[1].Selectable = false;
```
Rampastring, line 347: Would be clearer to just hide this instead. Set `Visible = false` or preferably supply a function reference to the visibility checker for `AddItem` .

Rampastring (changes requested):
Looks good aside from the note.

Rampastring (approved):

PR #173, opened 2020-08-25
Unix compatibility fixes

Starkku:
- Wine Mono implementation of .NET has `Application.StartupPath` return a value with trailing directory separator char that it normally does not have, which causes `ProgramConstants.GamePath` to be set to incorrect value and essentially prevented running the client properly on Wine as an alternative to dealing with Mono as-is and path problems. Wiping the trailing slash fixes this issue.
- OpenGL build didn't correctly update the window title. I believe the position in code where the title is set was changed prior to the last major version increment to Rampastring.XNAUI / MonoGame to make sure that it doesn't briefly show the default title before changing to the correct one. For whatever reason OpenGL MonoGame doesn't allow setting the title in `Game` class constructor anymore, but moving it back to the `Initialize` method seems to come with no drawbacks for any build of the client currently, so that is what I did.
- Changed the project configuration and build scripts to copy over the x64 and macOS versions of the required libraries for OpenGL MonoGame as well. I can personally attest to that at least the x64 libraries seem to be required on some systems. It is a few extra megabytes to copy over, but to me it seems like a smart idea to cover the possibility of them being required.

Left PR as a draft for now because there are further changes that could be done to complete the compatibility fixes. Might need some input before tackling these, albeit I wouldn't mind leaving it up for other people to resolve either.

- Replacing all hardcoded directory separator characters (backslashes) with `Path.DirectorySeparatorChar`. A rather simple thing to do really, just a matter of going through all of the code and picking the best method of applying this change, be it concatenation, formatting, string interpolation or `Path.Combine/Join`. A couple of things to note, however:
  - There is one instance of a hardcoded backslash used for paths in `Rampastring.Tools`, so it would require updating the libraries as well.
  - `ProgramConstants.GamePath` needs to be written into FinalSun/FinalAlert settings INI file with backslashes regardless of system-supported directory separator characters, map editor can't handle forward slashes at all.
- Addressing a similar problem with paths parsed from INI files. In practice I see three ways to go about solving this one:
  1. Replacing backslashes with directory separator char on path strings after they've been parsed from INI, separately everywhere where it is required. 
  2. Adding a new method to both `INIFile` & `INISection` (something like `GetPathStringValue`) that essentially does the same but with less repeated code, only requiring replacing the regular string value get call at quite a lot of locations.
  3. No action at all, leaving it up to modders to use forward slashes (which are also supported by Windows systems) in their configuration files.


Starkku:
Marked PR ready for review. The course of action taken for the file path separators is as follows:

- Hardcoded filepaths have had the separator changed from backslashes (\) to forward slashes (/) which should be supported on all relevant systems.
- INI-defined paths must be manually adjusted to use forward slashes instead (option number 3 listed in the initial description).

DXMainClient/Startup.cs
```diff
@@ -191,14 +191,14 @@ private static void GenerateOnlineId()
...
-                Registry.CurrentUser.CreateSubKey("SOFTWARE\\" + ClientConfiguration.Instance.InstallationPathRegKey).SetValue("Ident", cpuid + mbid + sid);
+                Registry.CurrentUser.CreateSubKey("SOFTWARE/" + ClientConfiguration.Instance.InstallationPathRegKey).SetValue("Ident", cpuid + mbid + sid);
             }
             catch (Exception)
             {
                 Random rn = new Random();
 
                 RegistryKey key;
-                key = Registry.CurrentUser.CreateSubKey("SOFTWARE\\" + ClientConfiguration.Instance.InstallationPathRegKey);
+                key = Registry.CurrentUser.CreateSubKey("SOFTWARE/" + ClientConfiguration.Instance.InstallationPathRegKey);
```
Rampastring, line 201: Using forward slashes supposedly doesn't work in Windows registry paths.

DXMainClient/Startup.cs
```diff
@@ -233,7 +233,7 @@ private void WriteInstallPathToRegistry()
             try
             {
                 RegistryKey key;
-                key = Registry.CurrentUser.CreateSubKey("SOFTWARE\\" + ClientConfiguration.Instance.InstallationPathRegKey);
+                key = Registry.CurrentUser.CreateSubKey("SOFTWARE/" + ClientConfiguration.Instance.InstallationPathRegKey);
```
Rampastring, line 236: See above.

Rampastring (changes requested):

DXMainClient/Startup.cs
Starkku: Resolved in a42d749.

DXMainClient/Startup.cs
Starkku: Resolved in a42d749.

Rampastring (approved):

PR #179, opened 2020-09-12
Tunnel autochooser

dkeetonx:
Adds the new tunnel selection algorithm that's designed to distribute players over tunnels more efficiently.

### DIFF
--- a/DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
+++ b/DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
@@ -581,6 +581,12 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
 
                 List<int> playerPorts = tunnelHandler.CurrentTunnel.GetPlayerPortInfo(Players.Count);
 
+                if (playerPorts.Count < Players.Count && tunnelHandler.Tunnels.Any())
+                {
+                    tunnel = tunnelHandler.Tunnels.Aggregate((i1, i2) => i1.Rating < i2.Rating ? i1 : i2);
+                    playerPorts = tunnel.GetPlayerPortInfo(Players.Count);
+                }
+
                 if (playerPorts.Count < Players.Count)
                 {
                     ShowTunnelSelectionWindow("An error occured while contacting " +
--- a/DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
+++ b/DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
@@ -583,9 +583,12 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
 
                 if (playerPorts.Count < Players.Count && tunnelHandler.Tunnels.Any())
                 {
+                    Logger.Log($"Picking a new tunnel server because '{tunnelHandler.CurrentTunnel.Name}' was full");
                     tunnelHandler.CurrentTunnel = tunnelHandler.Tunnels.Aggregate((i1, i2) => i1.Rating < i2.Rating ? i1 : i2);
                     playerPorts = tunnelHandler.CurrentTunnel.GetPlayerPortInfo(Players.Count);
                 }
+
+                Logger.Log($"Launching Game using tunnel server '{tunnelHandler.CurrentTunnel.Name}'");
 
                 if (playerPorts.Count < Players.Count)
                 {
--- a/DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
+++ b/DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
@@ -41,7 +41,7 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
 
         public CnCNetGameLobby(WindowManager windowManager, string iniName,
             TopBar topBar, List<GameMode> GameModes, CnCNetManager connectionManager,
-            TunnelHandler tunnelHandler, GameCollection gameCollection, CnCNetUserData cncnetUserData, MapLoader mapLoader, DiscordHandler discordHandler) : 
+            TunnelHandler tunnelHandler, GameCollection gameCollection, CnCNetUserData cncnetUserData, MapLoader mapLoader, DiscordHandler discordHandler) :
             base(windowManager, iniName, topBar, GameModes, mapLoader, discordHandler)
         {
             this.connectionManager = connectionManager;
@@ -175,7 +175,7 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
 
         private void GameBroadcastTimer_TimeElapsed(object sender, EventArgs e) => BroadcastGame();
 
-        public void SetUp(Channel channel, bool isHost, int playerLimit, 
+        public void SetUp(Channel channel, bool isHost, int playerLimit,
             CnCNetTunnel tunnel, string hostName, bool isCustomPassword)
         {
             this.channel = channel;
@@ -583,14 +583,14 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
 
                 if (playerPorts.Count < Players.Count && tunnelHandler.Tunnels.Any())
                 {
-                    tunnel = tunnelHandler.Tunnels.Aggregate((i1, i2) => i1.Rating < i2.Rating ? i1 : i2);
-                    playerPorts = tunnel.GetPlayerPortInfo(Players.Count);
+                    tunnelHandler.CurrentTunnel = tunnelHandler.Tunnels.Aggregate((i1, i2) => i1.Rating < i2.Rating ? i1 : i2);
+                    playerPorts = tunnelHandler.CurrentTunnel.GetPlayerPortInfo(Players.Count);
                 }
 
                 if (playerPorts.Count < Players.Count)
                 {
                     ShowTunnelSelectionWindow("An error occured while contacting " +
-                        "the CnCNet tunnel server." + Environment.NewLine + 
+                        "the CnCNet tunnel server." + Environment.NewLine +
                         "Try picking a different tunnel server:");
                     AddNotice("An error occured while contacting the specified CnCNet " +
                         "tunnel server. Please try using a different tunnel server ", ERROR_MESSAGE_COLOR);
@@ -641,7 +641,7 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
         {
             if (Map == null || GameMode == null)
             {
-                AddNotice("The game host needs to select a different map or " + 
+                AddNotice("The game host needs to select a different map or " +
                     "you will be unable to participate in the match.");
                 return;
             }
@@ -703,8 +703,8 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
             if (team < 0 || team > 4)
                 return;
 
-            if (side != pInfo.SideId 
-                || start != pInfo.StartingLocation 
+            if (side != pInfo.SideId
+                || start != pInfo.StartingLocation
                 || team != pInfo.TeamId)
             {
                 ClearReadyStatuses();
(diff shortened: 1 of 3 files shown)


DXMainClient/Domain/Multiplayer/CnCNet/CnCNetTunnel.cs
```diff
@@ -85,6 +85,29 @@ public static CnCNetTunnel Parse(string str)
...
+                    return int.MaxValue;
+
+                if (Official || Recommended)
+                {
+                    if (PingInMs <= -1)
+                        return int.MaxValue - 200000 + Clients;
+
+                    int ping = PingInMs < 1 ? 1000 : PingInMs;
+
+                    return (int)(Math.Ceiling((double)ping / 70) * 100000) + Clients;
```
Metadorius, line 102: Would be good to have a comment explaining this stuff (even better if it was a docstring for constants to be used instead of magic numbers and the property itself). The first comment about quadratic/exponential function also applies here. Also magic numbers again.

DXMainClient/Domain/Multiplayer/CnCNet/CnCNetTunnel.cs
```diff
@@ -85,6 +85,29 @@ public static CnCNetTunnel Parse(string str)
...
+                }
+                else
+                {
+                    return int.MaxValue - 100000 + Clients;
```
Metadorius, line 106: Ditto.

Metadorius (changes requested):
Tested it, this algorithm finally gave me adequate tunnel server with low ping. Overall the changes seem good but lack explanation in some parts of them, and the magic numbers need to be converted to constants and I suggest to us some kind of an exponential/quadratic function to set a rating for a server. I might take a stab at doing it myself though.

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
```diff
@@ -581,10 +581,16 @@ protected override void HostLaunchGame()
 
                 List<int> playerPorts = tunnelHandler.CurrentTunnel.GetPlayerPortInfo(Players.Count);
 
+                if (playerPorts.Count < Players.Count && tunnelHandler.Tunnels.Any())
+                {
+                    tunnelHandler.CurrentTunnel = tunnelHandler.Tunnels.Aggregate((i1, i2) => i1.Rating < i2.Rating ? i1 : i2);
+                    playerPorts = tunnelHandler.CurrentTunnel.GetPlayerPortInfo(Players.Count);
```
Metadorius, line 587: I'm against hidden changes of tunnel server, and this stuff also makes the whole following codeblock obsolete.

DXMainClient/Domain/Multiplayer/CnCNet/CnCNetTunnel.cs
```diff
@@ -85,6 +85,29 @@ public static CnCNetTunnel Parse(string str)
...
+            {
+                if (Clients + 24 >= MaxClients)
+                    return int.MaxValue;
+
+                if (Official || Recommended)
+                {
+                    if (PingInMs <= -1)
+                        return int.MaxValue - 200000 + Clients;
+
+                    int ping = PingInMs < 1 ? 1000 : PingInMs;
```
Metadorius, line 100: 0 is a valid ping value, and you have already checked for -1 previously.

Metadorius (changes requested):
On the second glance there are a couple more questionable changes.

DXMainClient/Domain/Multiplayer/CnCNet/CnCNetTunnel.cs
dkeetonx: Named constants with comments have been defined based on this recommendation.

DXMainClient/Domain/Multiplayer/CnCNet/CnCNetTunnel.cs
```diff
@@ -85,6 +85,29 @@ public static CnCNetTunnel Parse(string str)
         public double Distance { get; private set; }
         public int PingInMs { get; set; } = -1;
 
+        public int Rating
+        {
+            get
+            {
+                if (Clients + 24 >= MaxClients)
+                    return int.MaxValue;
```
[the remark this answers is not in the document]
dkeetonx: I think a more complicated algorithm might be better in narrow cases but IMO the complexity out weighs the benefit.

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
dkeetonx: I've added logging here so that the change is no longer hidden.

DXMainClient/Domain/Multiplayer/CnCNet/CnCNetTunnel.cs
```diff
@@ -85,6 +85,59 @@ public static CnCNetTunnel Parse(string str)
...
+                    if (PingInMs <= -1)
+                        return UNTRUSTED_RATING + Clients;
+
+                    int latency = 1 + PingInMs / LATENCY_PRECISION;
+
+                    return (latency * LATENCY_WEIGHT) + (Clients * CLIENTS_WEIGHT);
+                }
+                else
+                {
+                    return UNTRUSTED_RATING + Clients;
```
Metadorius, line 136: With this code it looks like the clients would almost never get an unofficial and non-recommended server selected. This is not the behavior I see as good, CnCNet is meant to be a community service, and IMO recommendation should boost the server, not basically "whitelist" non-recommended servers.

Metadorius (commented):
I am still concerned about the changes, up to further discussion in Discord.

DXMainClient/Domain/Multiplayer/CnCNet/CnCNetTunnel.cs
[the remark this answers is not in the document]
dkeetonx: The point of this code is not to select tunnels with fewer clients (that's handled elsewhere). The point of this code comes from the fact that the tunnel chooser first runs when the game lobby is created and after each ping, there is no way to know if the selected tunnel will have enough resources to host our game when we're ready to start (ie, resource is requested/granted on game start). So we're creating some headroom that will give us a reasonable chance to get our first tunnel choice on game start.

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
dkeetonx: This bit of code is designed to reduce frustration when a player wants to start a game and the tunnel server is out of resources. Manual tunnel selection is an advanced option  now but it will be exclusively a power user option in the future when v3 tunnels are implemented.

The manual selection code isn't really ready to handle corner cases like this. It's a bit hacky, it should have it's own PR to handle it correctly. I don't think it's a good idea to block this PR for that reason.

DXMainClient/Domain/Multiplayer/CnCNet/CnCNetTunnel.cs
dkeetonx: This is the best case, the recommended servers have all been tested and proven to be good. The goal is for the users to never want or need to select a tunnel server themselves. When they do manually select a tunnel server it should be their own server that they want to use for personal reasons.  This provides the added benefit that we can do some hacks on the master-list side to balance tunnel servers during the high load times.

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
Metadorius: The code redundancy must be fixed either way, preferably in a way I proposed - change tunnel server silently only in case the user didn't select the tunnel manually before.

PR shouldn't introduce more inoptimal solutions code-wise, in my opinion.

DXMainClient/Domain/Multiplayer/CnCNet/CnCNetTunnel.cs
```diff
@@ -85,6 +85,59 @@ public static CnCNetTunnel Parse(string str)
...
+        /// Rating Factor:
+        ///   Base Rating for untrusted servers and unpingable servers.
+        /// </summary>
+        private const int UNTRUSTED_RATING = int.MaxValue - 100000;
+
+        /// <summary>
+        /// Rating Factor:
+        ///   The amount of free client slots needed for the tunnel to be eligible for autoselection
+        /// </summary>
+        private const int CLIENTS_HEADROOM = 24;
```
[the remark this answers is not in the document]
Metadorius: Live example: Berryville tunnel has 50 slots, so the server will be running at 26 players max. in case of autochoosing it?
I think the best would be to use tangent function here, so when the tunnel clients amount is very close to max - the rating will skyrocket thus making it never chosen automatically.

DXMainClient/Domain/Multiplayer/CnCNet/CnCNetTunnel.cs
Metadorius: I still object against verified and official tunnels being effectively whitelisted against regular ones, this is not future-proof at all. I mean the example is already at hand. I have a feeling no one actually bothered to test Berryville, yet it has very low ping for Europe and I never had issues using it. Imagine CnCNet staff going very busy all at once, then some of the verified tunnel will also go off and the players will be automatically balanced to 2-3 verified servers in USA, for example, even if they are from EU and there is regular, non-verified EU servers. This is absolutely not a desirable solution.

Metadorius (changes requested):
Pushed a bit of fixes to tunnel autochooser branch. There are still concerns that need to be addressed.

DXMainClient/Domain/Multiplayer/CnCNet/CnCNetTunnel.cs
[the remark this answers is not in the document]
dkeetonx: The CLIENTS_HEADROOM is not for performance reasons. It's meant to make sure that when we go to use a tunnel server it has enough free slots to fill our request. The problem occurs because there is a time difference between when we select the tunnel server (game lobby creation) to when we request the user slots (game start).

DXMainClient/Domain/Multiplayer/CnCNet/CnCNetTunnel.cs
dkeetonx: The reason we should never use untrusted tunnel servers is because they fail more often. When a tunnel server fails and it breaks the game for a group of players, the players don't blame the tunnel server owner they blame CnCNet, we can't stake our reputation on untrusted servers.

> Imagine CnCNet staff going very busy all at once, then some of the verified tunnel will also go off and the players will be automatically balanced to 2-3 verified servers in USA, for example, even if they are from EU and there is regular, non-verified EU servers. This is absolutely not a desirable outcome IMO.

That's still more desirable than a broken tunnel server with low ping destroying everyone's game.

DXMainClient/Domain/Multiplayer/CnCNet/CnCNetTunnel.cs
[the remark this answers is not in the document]
Metadorius: I understood that already, yet the issue I mentioned is still here. Servers with low max. player amounts would never be chosen or chosen rarely.

IMO the issue of time difference should be dealt with some other sort of measure, f.ex. a task in game lobby checking for availability of the current tunnel on tunnel refresh. Ideally the players in lobby would be counted as players on servers even before starting but that would require server-side changes.

DXMainClient/Domain/Multiplayer/CnCNet/CnCNetTunnel.cs
Metadorius: My view of future CnCNet is as of a community service, not an organization, so I don't think this outcome is more desirable. I don't say that the system should be ditched. Per my idea verified servers would be "boosted" and chosen most of the time, but in case there's no local verified servers with low enough ping - regular one would be chosen. This would never happen in the current implementation.

Again, you say that untrusted servers are bad because all good servers are made verified. Is that so? Not really. Someone has to test them manually, and as I said I don't see anyone testing Berryville server or some other EU servers which are OK despite the lack of EU tunnels. I always get Canada East chosen and I live in Europe.

Metadorius:
Closing because of no activity. If someone wants to pick this up -- feel free to re-open or re-submit.

PR #142, opened 2020-04-17
Added DNS resolution and latency test for the Lobby servers.

pzhlkj6612:
Thanks for your great work. I've made some modification about the process of connection to the Lobby servers. Please review and test it. Thank you.

### Commit message

1. Get all IP addresses of servers by resolving the hostname.
2. Test the latency to the servers and store them in a list.
3. If the latency is greater than 400 ms, ignore the server.
4. Sort the list and connect to the servers one by one.

### It can...

1. Find the closest and reachable server in few seconds.
2. Connect to all servers bound to one domain name.

### It also can...

1. Resolve #141 .

### Test

At 10:16 PM 4/17/2020, there are 843 players on the server.

<details>
  <summary>client.log</summary>
  <pre>
...
17.04. 22:16:59.701    Writing settings INI.
17.04. 22:16:59.739    Attempting to DNS resolve GameSurge (irc.gamesurge.net).
17.04. 22:16:59.760    Attempting to DNS resolve GameSurge London, UK (Burstfire.UK.EU.GameSurge.net).
17.04. 22:16:59.769    Attempting to DNS resolve GameSurge Chicago, IL (ColoCrossing.IL.US.GameSurge.net).
17.04. 22:16:59.780    DNS resolved GameSurge (irc.gamesurge.net): 108.174.48.100, 91.217.189.76, 192.223.27.109, 162.248.94.123, 195.68.206.250, 195.8.250.180
17.04. 22:16:59.789    DNS resolved GameSurge London, UK (Burstfire.UK.EU.GameSurge.net): 195.68.206.250
17.04. 22:16:59.804    Attempting to ping GameSurge (195.8.250.180).
17.04. 22:16:59.812    DNS resolved GameSurge Chicago, IL (ColoCrossing.IL.US.GameSurge.net): 108.174.48.100
17.04. 22:16:59.829    Attempting to ping GameSurge London, UK (195.68.206.250).
17.04. 22:16:59.840    Attempting to ping GameSurge Chicago, IL (108.174.48.100).
17.04. 22:17:00.064    The latency in milliseconds to the server GameSurge (195.8.250.180): 249.
17.04. 22:17:00.076    Attempting to ping GameSurge (195.68.206.250).
17.04. 22:17:00.140    The latency in milliseconds to the server GameSurge Chicago, IL (108.174.48.100): 286.
17.04. 22:17:00.151    Attempting to DNS resolve GameSurge Newark, NJ (Gameservers.NJ.US.GameSurge.net).
[...]

PR #185, opened 2020-09-29
Enable IPv6 connection (dual stack) with Lobby servers on DirectX and OpenGL clients

pzhlkj6612:
Enhance #142 .

Learned from [Dual Mode Sockets - Never create an IPv4 Socket again | ASP.NET Blog](https://devblogs.microsoft.com/aspnet/dual-mode-sockets-never-create-an-ipv4-socket-again/) .

Why `#if XNA`: [`Socket.DualMode`](https://docs.microsoft.com/dotnet/api/system.net.sockets.socket.dualmode) is supported since .NET Framework 4.5, while XNA Framework depends on .NET Framework 4.0, so it is a bit complicated to implement dual-stack connection on XNA.

@Metadorius , thank you for your suggestion.

### DIFF
--- a/DXMainClient/Online/Connection.cs
+++ b/DXMainClient/Online/Connection.cs
@@ -173,7 +173,8 @@ namespace DTAClient.Online
                     {
                         connectionManager.OnAttemptedServerChanged(server.Name);
 
-                        TcpClient client = new TcpClient(AddressFamily.InterNetwork);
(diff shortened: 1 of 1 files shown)


pzhlkj6612:
There seems to be some problems on the server side, so we cannot fully test the IPv6 connection yet. This branch will keep alive and sometimes merge changes from the "master" branch.

PR #186, opened 2020-09-30
Features / improvements for Ares build of the client.

Starkku:
All functionality changes below only apply to `AresRelease` and `AresDebug` configurations. Sole exception is that the graphics mode reset hack in `GameInProgressWindow.HandleGameProcessExited` was moved so that the code executes before there's any chance for the method to exit. Previously if deleting the log files on game startup failed for any reason, this reset hack wouldn't be applied on game exit, which didn't seem like intended behaviour.

- Client will now prune Ares debug log files & directories on startup. Any files that were created at least 7 days ago will be deleted and any directories that are empty after deleting files are also removed. The time threshold could be made customizable if a good enough reason to allow it comes up, but I could not think of any.
- Moved `ProcessScreenshots` method from `GameLobbyBase` to `GameInProgressWindow` for consistency as that is where all existing file system operations on game start / exit are performed.
- Deprecated `ProcessScreenshots` setting, as there is practically never a genuine need to turn this feature off.
- Instead of copying any `except.txt` and `SYNC*.txt` files from game directory to `Client\ErrorLogs` on game exit like for other configurations, client will now do the following (ideally these would be handled by Ares itself, but until that happens this is the next best thing):
  - Copies `syringe.log` from game directory to a snapshot directory in `debug` directory on crash & desync.
  - Copies `SYNC*.txt` files from game directory to a snapshot directory in `debug` directory on a desync.
  - In case of a desync occurs and game does not create the snapshot directory (happens if player denies creating a full crash dump), client will create it.
  - If `debug.log` was written to during the game, a snapshot directory was created by either game or client (a crash / desync occured) and `debug.log` wasn't copied to the snapshot directory by game, client will copy it over.

### DIFF
--- a/DXMainClient/DXGUI/Generic/GameInProgressWindow.cs
+++ b/DXMainClient/DXGUI/Generic/GameInProgressWindow.cs
@@ -194,8 +194,8 @@ namespace DTAClient.DXGUI
             if (deletingLogFilesFailed)
                 return;
 
-            CopyErrorLog(ProgramConstants.GamePath + "Client/ErrorLogs", "EXCEPT.TXT", dtn);
-            CopySyncErrorLogs(ProgramConstants.GamePath + "Client/ErrorLogs", dtn);
+            CopyErrorLog(ProgramConstants.ClientUserFilesPath + "GameCrashLogs", "EXCEPT.TXT", dtn);
+            CopySyncErrorLogs(ProgramConstants.ClientUserFilesPath + "SyncErrorLogs", dtn);
 #endif
         }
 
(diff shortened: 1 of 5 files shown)


DXMainClient/DXGUI/Generic/GameInProgressWindow.cs
```diff
@@ -131,41 +152,223 @@ private void HandleGameProcessExited()
...
 
-                if (File.Exists(ProgramConstants.GamePath + "EXCEPT.TXT"))
+                CopyErrorLog(snapshotDirectory, "syringe.log", null);
+            }
+#else
+            if (deletingLogFilesFailed)
+                return;
+
+            CopyErrorLog(ProgramConstants.GamePath + "Client/ErrorLogs", "EXCEPT.TXT", dtn);
+            CopySyncErrorLogs(ProgramConstants.GamePath + "Client/ErrorLogs", dtn);
```
Metadorius, line 198: I wonder if we should copy the error logs in what seems to be the client-only stuff folder?

Metadorius (approved):

DXMainClient/DXGUI/Generic/GameInProgressWindow.cs
Starkku: As a compromise and to not clutter the main game directory with multiple more subdirectories, they are now saved to separate subdirectories within the Client directory.

Metadorius (approved):

PR #197, opened 2020-10-22
Bug fix: Parse the UTF-16 LE string in saved game files

SadPencil:
This PR fixed the bug that the name of saved games can't be displayed correctly, when the name contains non-ASCII characters. Now the program is able to handle MOST of the UTF-16 LE string correctly.

Also, see the [screenshot](https://ibb.co/3BdFr64).


SadPencil:
Please review this pull request as it solves the issue involved with multiple languages such as Russian, French, German, and Chinese, and thus it improves the user experience without cons.

PR #200, opened 2020-11-14
Reimplement InternalSideIndices and SpectatorInternalSideIndex GameOptions.ini keys

Metadorius:
`InternalSideIndices` (in `GameOptions.ini`, `[General]` section) is used to override default side IDs with the ones defined in lookup string. It's also now possible to override the side for spectator (and thus supply custom music, sidebar etc.) via explicitly specifying random side index via `SpectatorInternalSideIndex`, for example:

```ini
Sides=USA,United Kingdom,France...
InternalSideIds=0,4,2,...
SpectatorInternalSideId=16
```

### DIFF
--- a/ClientCore/ClientConfiguration.cs
+++ b/ClientCore/ClientConfiguration.cs
@@ -140,7 +140,11 @@ namespace ClientCore
 
         #region Game options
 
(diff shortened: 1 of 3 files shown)


Crimsonum:
Tested this feature, everything seems to work as intended.

PR #204, opened 2020-11-24
Improvements to TunnelSelectionWindow & GameCreationWindow

### LINKED ISSUE #191: Add a possiblity to cancel tunnel change prompt when invoked manually
Right now you can only select old tunnel from the list and it will be announced as a changed tunnel which is not the desired behavior.

Starkku:
**TunnelSelectionWindow:**
- Add Cancel button that can be used to exit the window without applying changes.
- Apply button is disabled when currently selected tunnel is the original one.
- Tunnel lists scrolls to the selected tunnel on opening the window.
- Less cramped control spacing.

tunnelselectionwindow

**GameCreationWindow:**
- Improve control spacing, especially regarding the buttons & bottom edge of the window.
- Set window width based on width of TunnelListBox and use global constants for control positions & sizes where possible.
- Set window height based on contents.
- Set names for controls to allow adjusting attributes from INI if needed.

gamecreationwindow
gamecreationwindow_advanced

Additionally `TunnelListBox` default height is now set to better match the contents and the way listbox scrolling currently works.

Main reason for changing `GameCreationWindow` here is that if in future one needs to adjust size of `TunnelListBox` for any reason, default size of `GameCreationWindow` and positioning of its child controls will now automatically match afterwards.

Resolves #191

### DIFF
--- a/DXMainClient/DXGUI/Multiplayer/CnCNet/TunnelListBox.cs
+++ b/DXMainClient/DXGUI/Multiplayer/CnCNet/TunnelListBox.cs
@@ -25,8 +25,10 @@ namespace DTAClient.DXGUI.Multiplayer.CnCNet
 
             SelectedIndexChanged += TunnelListBox_SelectedIndexChanged;
 
+            int headerHeight = (int)Renderer.GetTextDimensions("Name", HeaderFontIndex).Y;
+
             Width = 466;
-            Height = 200;
+            Height = LineHeight * 12 + headerHeight + 3;
             PanelBackgroundDrawMode = PanelBackgroundImageDrawMode.STRETCHED;
             BackgroundTexture = AssetLoader.CreateTexture(new Color(0, 0, 0, 128), 1, 1);
             AddColumn("Name", 230);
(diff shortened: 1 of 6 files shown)


DXMainClient/DXGUI/Multiplayer/CnCNet/TunnelListBox.cs
```diff
@@ -63,6 +65,17 @@ public void SelectTunnel(string address)
...
+        /// <summary>
+        /// Gets whether or not a tunnel from the list with the given address is selected.
+        /// </summary>
+        /// <param name="address">The address of the tunnel server</param>
+        /// <returns>True if tunnel with given address is selected, otherwise false.</returns>
+        public bool IsTunnelSelected(string address)
+        {
+            int index = tunnelHandler.Tunnels.FindIndex(t => t.Address == address);
+            return index == SelectedIndex;
+        }
```
Metadorius, line 77: Think the local variable is unnecessary here, could also use expression body here.

Metadorius (changes requested):
Looks good aside of a few places where expression bodies can be used. Also would be good to add screenshots of the new windows but that's just good-to-have when you are doing design changes.

Metadorius (approved):

PR #206, opened 2020-12-03
Remove assumption of MPColors being sorted by in-game color index when determining SpawnX indexes of players

Rampastring:


Metadorius:
Confirmed to fix the scrambing issue by @ayylmaorote

PR #207, opened 2020-12-12
Refine name checking code and Add feature AutoRemoveNonASCIIFromName

SadPencil:
This PR changes the behavior of DTA client:

- If the player name contains non-ASCII characters, replace them with '?'. This can be turned off by set `AutoRemoveNonASCIIFromName` as false in the ini file.
- Remove the unwanted invisible control characters.
- Will also check the new player name saved from the setting dialog.
- Refactor the code of checking offline player name, so that the code can be invoked in two places, i.e. when the client is starting as well as when user saved a new player name.

--------

Explanation of `AutoRemoveNonASCIIFromName`:

When the player name contains any non-ASCII characters, one of the following things happens:
- The name is not correctly displayed. Example: DTA, YR
- The game crashed. Example: Ares 3.0

Merging this PR to the repo solves this issue with minimal changes. And it's mostly benifiticial to users in East Asian.

### DIFF
--- a/ClientCore/CnCNet5/NameValidator.cs
+++ b/ClientCore/CnCNet5/NameValidator.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 
 namespace ClientCore.CnCNet5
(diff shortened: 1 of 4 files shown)


Rampastring:
Tiberian Sun (DTA) at least uses Windows-1252 encoding for the names rather than ASCII, so it might be preferable to support Windows-1252 characters instead of ASCII only for offline play.

Metadorius:
YR has full Unicode support internally, it's just that the font file is missing some codepages. I don't think that should be fixed here for YR, we just need to provide correct font files. I am not sure how did you get Ares to crash with Unicode characters.

SadPencil:
> 
> 
> YR has full Unicode support internally, it's just that the font file is missing some codepages. I don't think that should be fixed here for YR, we just need to provide correct font files. I am not sure how did you get Ares to crash with Unicode characters.

An example is Mental Omega 3.3.5 which includes Ares 3.0. Ares 3.0 will ignore any non-ASCII characters, so that the player name can be empty (all characters are ignored by Ares), and then Ares crashes.

SadPencil:
> 
> 
> Tiberian Sun (DTA) at least uses Windows-1252 encoding for the names rather than ASCII, so it might be preferable to support Windows-1252 characters instead of ASCII only for offline play.

Currently, the DTA client write `spawn.ini` file with UTF-8 encoding, which means that any characters that takes more than one byte in UTF-8 (i.e. any non-ASCII characters) will be corrupted.

SadPencil:
> 
> 
> Tiberian Sun (DTA) at least uses Windows-1252 encoding for the names rather than ASCII, so it might be preferable to support Windows-1252 characters instead of ASCII only for offline play.

Take an example. The following text is supported by Windows-1252 encoding.
```
an example painâ€” if youâ€™re
```
So, choose `painâ€` as the player name. Is the player name displayed correctly after entering the game? (not in the DTA client)

SadPencil:
Your latest code brings a new issue here:

```cs
            buffer = new byte[(BitConverter.ToInt32(buffer, 0) - 1) << 1];
```

You can't assume whether the computer is little-endian or big-endian, on which the behavior of `BitConverter.ToInt32` depends.

SadPencil:
I set the default value of `AutoRemoveNonASCIIFromName` to false, so that the rest changes of this PR can be merged:

- Remove the unwanted invisible control characters.
- Will also check the new player name saved from the setting dialog.
- Refactor the code of checking offline player name, so that the code can be invoked in two places, i.e. when the client is starting as well as when user saved a new player name.

And move the discussion of `AutoRemoveNonASCIIFromName` into an issue.

PR #211, opened 2020-12-18
Add feature of random loading screen support

SadPencil:
This PR add a new feature, where the client starts, it choose one file as the loading screen randomly, of the following picture files:
- `loadingscreen.png`
- ~~`loadingscreen0.png`~~
- `loadingscreen1.png`
- `loadingscreen2.png`
- ... and so on

~~The random code here obeys uniform distribution.~~

There is no breaking change in this PR.

### DIFF
--- a/DXMainClient/DXGUI/Generic/LoadingScreen.cs
+++ b/DXMainClient/DXGUI/Generic/LoadingScreen.cs
@@ -40,7 +40,25 @@ namespace DTAClient.DXGUI.Generic
             ClientRectangle = new Rectangle(0, 0, 800, 600);
             Name = "LoadingScreen";
 
-            BackgroundTexture = AssetLoader.LoadTexture("loadingscreen.png");
+            string backgroundName = "loadingscreen.png";
+
+            var rand = new System.Random();
+            for (int i = 0; ; i++)
+            {
+                string currentName = "loadingscreen" + i.ToString(System.Globalization.CultureInfo.InvariantCulture) + ".png";
+                if (!AssetLoader.AssetExists(currentName))
+                {
+                    break;
+                }
+
+                var roll = rand.Next();
+                if (roll % 2 == 0)
+                {
+                    backgroundName = currentName;
+                }
+            }
+
+            BackgroundTexture = AssetLoader.LoadTexture(backgroundName);
 
             base.Initialize();
 
--- a/DXMainClient/DXGUI/Generic/LoadingScreen.cs
+++ b/DXMainClient/DXGUI/Generic/LoadingScreen.cs
@@ -34,6 +34,7 @@ namespace DTAClient.DXGUI.Generic
 
         private Task updaterInitTask = null;
         private Task mapLoadTask = null;
+        
         private string GetBackgroundName(int id)
         {
             if (id == 0)
--- a/DXMainClient/DXGUI/Generic/LoadingScreen.cs
+++ b/DXMainClient/DXGUI/Generic/LoadingScreen.cs
@@ -40,22 +40,32 @@ namespace DTAClient.DXGUI.Generic
             ClientRectangle = new Rectangle(0, 0, 800, 600);
             Name = "LoadingScreen";
 
-            string backgroundName = "loadingscreen.png";
-
-            var rand = new System.Random();
-            for (int i = 0; ; i++)
+            string GetBackgroundName(int id)
             {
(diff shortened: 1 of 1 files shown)


DXMainClient/DXGUI/Generic/LoadingScreen.cs
```diff
@@ -40,7 +40,25 @@ public override void Initialize()
...
+            string backgroundName = "loadingscreen.png";
+
+            var rand = new System.Random();
+            for (int i = 0; ; i++)
+            {
+                string currentName = "loadingscreen" + i.ToString(System.Globalization.CultureInfo.InvariantCulture) + ".png";
+                if (!AssetLoader.AssetExists(currentName))
+                {
+                    break;
+                }
```
Metadorius, line 52: Unnecessary braces.

DXMainClient/DXGUI/Generic/LoadingScreen.cs
```diff
@@ -40,7 +40,25 @@ public override void Initialize()
...
+
+                var roll = rand.Next();
+                if (roll % 2 == 0)
+                {
+                    backgroundName = currentName;
+                }
```
Metadorius, line 58: Ditto.

Metadorius (changes requested):
Instead of rolling random every time in cycle better just determine a maximum number and roll random one time when breaking.

DXMainClient/DXGUI/Generic/LoadingScreen.cs
```diff
@@ -40,22 +40,32 @@ public override void Initialize()
             ClientRectangle = new Rectangle(0, 0, 800, 600);
             Name = "LoadingScreen";
 
-            string backgroundName = "loadingscreen.png";
+            string GetBackgroundName(int id)
```
Metadorius, line 43: AFAIK local functions are a feature of C# 7.0 which is unsupported by XNA build of the client which builds against .NET Framework 4.0. Should make it a regular function and add an access modifier, assumingly `private` here.

DXMainClient/DXGUI/Generic/LoadingScreen.cs
```diff
@@ -40,22 +40,32 @@ public override void Initialize()
...
-                var roll = rand.Next();
-                if (roll % 2 == 0)
-                {
-                    backgroundName = currentName;
-                }
+            }
+            if (backgroundMaximum > 0)
+            {
+                var rand = new System.Random();
+                var roll = rand.Next(backgroundMaximum + 1);
```
Metadorius, line 67: Per client code style the code shouldn't rely much on `var`s, especially in case of primitives.

Metadorius (changes requested):
The changes are good but there are a bit more issues to fix.

SadPencil:
Code updated. @Metadorius 
By the way, the coding style mentioned here can be defined at `.editconfig` file. https://github.com/CnCNet/xna-cncnet-client/issues/214

DXMainClient/DXGUI/Generic/LoadingScreen.cs
SadPencil: I'm not facing any problem when compiling all the three builds with Visual Studio 2019. The `clientxna.exe` file is generated correctly. Anyway, I have updated the code.

Metadorius (changes requested):
Small nitpick, I'll test meanwhile and merge.

Metadorius:
@SadPencil do you have a Discord? Since you are contributing actively lately would be good to use some form of messaging to stay in touch. If you do - hit me up (Kerbiter#3128).

Starkku:
The code's fine and the idea is solid, but I think ideally something like this should be implemented in `Rampastring.XNAUI`, by allowing defining a list of textures for the control properties that define texture names (`BackgroundTexture` et al) and it randomly picks one to use for the rest of the client session. As it stands this will only work for loading screen and only and only if you use the default texture name. If you set `BackgroundTexture` for `LoadingScreen` in INI, this feature will cease to function.

MahBoiDeveloper:
I disagree with the general approach to sampling file names in this PR. I suggest a more constructive approach to specifying specific file names for random sampling in the LoadScreen.ini file.

I also disagree with the proposal to reconsider the implementation of PR in XNAUI due to the fact that this solution is only needed by the client (WAE does not need this functionality at the moment).

Issue #164, opened 2020-07-01
Saved game names appear as garbage in Wine


cdavis5e:
This code assumes that the title of a saved game in [OLE Compound Document format](https://en.wikipedia.org/wiki/Compound_File_Binary_Format) is always stored at a particular offset:

https://github.com/CnCNet/xna-cncnet-client/blob/421f1b61f42b8a65b2eb847c5a6e30c57962829b/DXMainClient/Domain/SavedGame.cs#L34

This is a very fragile assumption, and it breaks on CDFs produced by Wine. Wine's CDFs, at least on my system, happen to have the title at offset 9c0h. When the client tries to read the Unicode string which it assumes is at 980h, it winds up reading garbage, and the save game names are unreadable in the client. (Saved games in the actual game engine work correctly and display the correct name.)

SadPencil:
Number 2256 is 0x8d0 in hex form, instead of 0x980 in your description above. Maybe you need to double check the offset? You say that the offset is dynamic on various system, is there any document available? And what are `CDFs`?

Besides, this part of codes has been updated with Unicode support. Is the problem still occurs on the latest build of the client on Wine?

SadPencil:
Okay, I got what you said. The problem can be solved. To validate the solution, could you upload a .sav file that is generated from your Wine environment?

```csharp
using OpenMcdf; // installed through Nuget

var cf = new CompoundFile(filename);
var archiveNameBytes = cf.RootStorage.GetStream("Scenario Description").GetData();
var archiveName = Encoding.Unicode.GetString(archiveNameBytes);
Console.WriteLine(archiveName);
```

cdavis5e:
> To validate the solution, could you upload a .sav file that is generated from your Wine environment?

Of course. Here's one from the Mental Omega mission, "Operation: Puppet Master" (Allied 14).
[APUPPET.sav.zip](https://github.com/CnCNet/xna-cncnet-client/files/6013783/APUPPET.sav.zip)

SadPencil:
> 
> 
> > To validate the solution, could you upload a .sav file that is generated from your Wine environment?
> 
> Of course. Here's one from the Mental Omega mission, "Operation: Puppet Master" (Allied 14).
> [APUPPET.sav.zip](https://github.com/CnCNet/xna-cncnet-client/files/6013783/APUPPET.sav.zip)

Yes. It is solved. I'll make a PR for this change.
See the picture below.

[...]

PR #213, opened 2021-01-03
change Saved Game feature 

frg2089:
some suggestions:
- Use [System.IO.Path](//docs.microsoft.com/dotnet/api/system.io.path) Member and **NOT** use character `'/'` or `'\\'` to cross platform.
- Use Properties instead of Methods.
- Make [.editorconfig](//docs.microsoft.com/visualstudio/ide/create-portable-custom-editor-options) file to constraint the code style.
- Use [nuget](//nuget.org) package instead of reference library.
- Use [Multilingual app toolkit](//docs.microsoft.com/windows/uwp/design/globalizing/use-mat) (well, it will be hard.)

### DIFF
--- a/ClientCore/SavedGameManager.cs
+++ b/ClientCore/SavedGameManager.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.IO;
+
 using Rampastring.Tools;
 
 namespace ClientCore
@@ -81,10 +82,8 @@ namespace ClientCore
             return false;
         }
 
-        private static string GetSaveGameDirectoryPath()
-        {
-            return ProgramConstants.GamePath + SAVED_GAMES_DIRECTORY;
-        }
+        private static string GetSaveGameDirectoryPath()// Use System.IO.Path Member and NOT use character '/' or '\\'
+            => Path.Combine(ProgramConstants.GamePath, SAVED_GAMES_DIRECTORY);
 
         /// <summary>
         /// Initializes saved MP games for a match.
--- a/ClientCore/SavedGameManager.cs
+++ b/ClientCore/SavedGameManager.cs
@@ -38,7 +38,7 @@ namespace ClientCore
 
         public static int GetSaveGameCount()
         {
-            string saveGameDirectory = GetSaveGameDirectoryPath() + "/";
+            string saveGameDirectory = SaveGameDirectoryPath + "/";
 
             if (!AreSavedGamesAvailable())
                 return 0;
@@ -60,7 +60,7 @@ namespace ClientCore
 
             List<string> timestamps = new List<string>();
 
-            string saveGameDirectory = GetSaveGameDirectoryPath() + "/";
+            string saveGameDirectory = SaveGameDirectoryPath + "/";
 
             for (int i = 0; i < saveGameCount; i++)
             {
@@ -76,13 +76,13 @@ namespace ClientCore
 
         public static bool AreSavedGamesAvailable()
         {
-            if (Directory.Exists(GetSaveGameDirectoryPath()))
+            if (Directory.Exists(SaveGameDirectoryPath))
                 return true;
 
             return false;
         }
 
-        private static string GetSaveGameDirectoryPath()// Use System.IO.Path Member and NOT use character '/' or '\\'
+        private static string SaveGameDirectoryPath // Use System.IO.Path Member and NOT use character '/' or '\\'
             => Path.Combine(ProgramConstants.GamePath, SAVED_GAMES_DIRECTORY);
 
         /// <summary>
@@ -120,7 +120,7 @@ namespace ClientCore
                 return;
             }
 
-            string saveGameDirectory = GetSaveGameDirectoryPath() + "/";
+            string saveGameDirectory = SaveGameDirectoryPath + "/";
 
             if (!File.Exists(saveGameDirectory + "SAVEGAME.NET"))
             {
@@ -187,7 +187,7 @@ namespace ClientCore
             {
                 for (int i = 0; i < 1000; i++)
                 {
-                    File.Delete(GetSaveGameDirectoryPath() +
+                    File.Delete(SaveGameDirectoryPath +
                         "/" + string.Format("SVGM_{0}.NET", i.ToString("D3")));
                 }
             }
--- a/ClientCore/SavedGameManager.cs
+++ b/ClientCore/SavedGameManager.cs
@@ -14,6 +14,27 @@ namespace ClientCore
 
         private static bool saveRenameInProgress = false;
 
+        /// <summary>
+        /// Get Archive Name
+        /// </summary>
+        /// <param name="stream">FileStream</param>
+        /// <returns>Archive Name</returns>
+        public static string GetSaveGameName(Stream stream)
+        {
+            var buffer = new byte[4];
+            stream.Seek(0x08CC, SeekOrigin.Begin);
+            stream.Read(buffer, 0, 4);// Get Name Length.
+
+            if (!BitConverter.IsLittleEndian)
+                Array.Reverse(buffer);
+            buffer = new byte[(BitConverter.ToInt32(buffer, 0) - 1) << 1];
+
+            stream.Read(buffer, 0, buffer.Length);// Get Name.
+            var result = System.Text.Encoding.Unicode.GetString(buffer);
+            Logger.Log($"[SavedGameManager]\tFound Archive: {result}");
+            return result;
+        }
+
         public static int GetSaveGameCount()
         {
             string saveGameDirectory = GetSaveGameDirectoryPath() + "/";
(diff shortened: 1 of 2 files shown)


SadPencil:
```cs
public static string GetSaveGameName(Stream stream)
        {
            var buffer = new byte[0x40];
```
Why would you set the limitation of the length of the saving game name while the game executable supports a much longer name?

SadPencil:
Apart from the limitation of the length, your new name loading logic brings other problems:

In the document of `Encoding.GetString Method`:
```
Exceptions
ArgumentException

The byte array contains invalid Unicode code points.
```

Suspose the name is shorter than `0x40`, it may causes the `ArgumentException`.

So, I strongly disagree with your new logic for loading the name of saved games. It added unnecessary limitations as well as unexpected exceptions. The current one in the repo is nicer.

frg2089:
> ```cs
> public static string GetSaveGameName(Stream stream)
>         {
>             var buffer = new byte[0x40];
> ```
> 
> Why would you set the limitation of the length of the saving game name while the game executable supports a much longer name?

sav file have a fixed block to save name.
at 0x09c0 and length is 0x40.
Actually, I don't know the other at 0x08d0 and there was no length limit.

AutoGavy:
你们两个老外可以说中文不

SadPencil:
> 
> 
> > ```cs
> > public static string GetSaveGameName(Stream stream)
> >         {
> >             var buffer = new byte[0x40];
> > ```
> > 
> > 
> > Why would you set the limitation of the length of the saving game name while the game executable supports a much longer name?
> 
> sav file have a fixed block to save name.
> at 0x09c0 and length is 0x40.
> Actually, I don't know the other at 0x08d0 and there was no length limit.

Which means, your code brings a stronger assumption on the `.sav` file: The name block is fixed to `0x40` and when a null character (0x00 0x00) is set, the rest bytes in the name block must also be `0x00` ( or any other bytes that doesn't raise `ArgumentException` ).

Well, in my opinion, a weaker assumption is better than a stronger one. Or, specify the assumption in the comment.

SadPencil:
Your latest code brings a new issue here:
```cs
            buffer = new byte[(BitConverter.ToInt32(buffer, 0) - 1) << 1];
```
You can't assume whether the computer is little-endian or big-endian, on which the behavior of BitConverter.ToInt32 depends.

SadPencil:
Actually, you should use `BitConverter.IsLittleEndian` to determine whether a reverse is needed, instead of picking a minimum value.
And only use English for comments please.

SadPencil:
Method `SavedGameManager.GetLength` make no sense.
From my view, the administrators in this repo only accept codes that is beautifully written. A code with such a method is definitely not pretty. 

Or you may wait for the suggestion about coding style from the administrators.

ClientCore/SavedGameManager.cs
```diff
@@ -76,10 +76,7 @@ public static List<string> GetSaveGameTimestamps()
 
         public static bool AreSavedGamesAvailable()
         {
-            if (Directory.Exists(SaveGameDirectoryPath))
-                return true;
-
-            return false;
+            return Directory.Exists(SaveGameDirectoryPath);
```
Metadorius, line 79: Should use expression body here.

Metadorius:
> Use System.IO.Path Member and NOT use character '/' or '\\' to cross platform.
What's the reason for this? `/` is way simpler to use and is supported by bot Windows and Unix. Also if you do it like this you have to change all occurences of slash also (again, for what?)

frg2089:
> > Use System.IO.Path Member and NOT use character '/' or '\' to cross platform.
> 
> What's the reason for this? `/` is way simpler to use and is supported by bot Windows and Unix. Also if you do it like this you have to change all occurences of slash also (again, for what?)

Used '/' is wrong. it should not be used in windows. Although it works, but it is not native to windows. New code should pay attention to this.
And it's only a suggestion, you don't have to take it.
[file-path-formats#canonicalize-separators](//docs.microsoft.com/dotnet/standard/io/file-path-formats#canonicalize-separators)
[System.IO.Path](//docs.microsoft.com/dotnet/api/system.io.path)

ClientCore/SavedGameManager.cs
frg2089: I'll fix it.
And you can make an .editorconfig file to constraint it.

Zero-Fanker:
> > > ```cs
> > > public static string GetSaveGameName(Stream stream)
> > >         {
> > >             var buffer = new byte[0x40];
> > > ```
> > > 
> > > 
> > > Why would you set the limitation of the length of the saving game name while the game executable supports a much longer name?
> > 
> > 
> > sav file have a fixed block to save name.
> > at 0x09c0 and length is 0x40.
> > Actually, I don't know the other at 0x08d0 and there was no length limit.
> 
> Which means, your code brings a stronger assumption on the `.sav` file: The name block is fixed to `0x40` and when a null character (0x00 0x00) is set, the rest bytes in the name block must also be `0x00` ( or any other bytes that doesn't raise `ArgumentException` ).
> 
> Well, in my opinion, a weaker assumption is better than a stronger one. Or at least specify the assumption in the comment.

See 'SavegameInformation' as Alexb discovered here :
https://github.com/Ares-Developers/YRpp/blob/master/LoadOptionsClass.h#L95

SadPencil:
> 
> 
> > > > ```cs
> > > > public static string GetSaveGameName(Stream stream)
> > > >         {
> > > >             var buffer = new byte[0x40];
> > > > ```
> > > > 
> > > > 
> > > > Why would you set the limitation of the length of the saving game name while the game executable supports a much longer name?
> > > 
> > > 
> > > sav file have a fixed block to save name.
> > > at 0x09c0 and length is 0x40.
> > > Actually, I don't know the other at 0x08d0 and there was no length limit.
> > 
> > 
> > Which means, your code brings a stronger assumption on the `.sav` file: The name block is fixed to `0x40` and when a null character (0x00 0x00) is set, the rest bytes in the name block must also be `0x00` ( or any other bytes that doesn't raise `ArgumentException` ).
> > Well, in my opinion, a weaker assumption is better than a stronger one. Or at least specify the assumption in the comment.
> 
> See 'SavegameInformation' as Alexb discovered here :
> https://github.com/Ares-Developers/YRpp/blob/master/LoadOptionsClass.h#L95

Good to know. Thanks.

SadPencil:
> 
> 
> > > > ```cs
> > > > public static string GetSaveGameName(Stream stream)
> > > >         {
> > > >             var buffer = new byte[0x40];
> > > > ```
> > > > 
> > > > 
> > > > Why would you set the limitation of the length of the saving game name while the game executable supports a much longer name?
> > > 
> > > 
> > > sav file have a fixed block to save name.
> > > at 0x09c0 and length is 0x40.
> > > Actually, I don't know the other at 0x08d0 and there was no length limit.
> > 
> > 
> > Which means, your code brings a stronger assumption on the `.sav` file: The name block is fixed to `0x40` and when a null character (0x00 0x00) is set, the rest bytes in the name block must also be `0x00` ( or any other bytes that doesn't raise `ArgumentException` ).
> > Well, in my opinion, a weaker assumption is better than a stronger one. Or at least specify the assumption in the comment.
> 
> See 'SavegameInformation' as Alexb discovered here :
> https://github.com/Ares-Developers/YRpp/blob/master/LoadOptionsClass.h#L95

After carefully reviewing the code you provided, the assumption given by @frg2089 at first is still stronger and thus not proper.

On his previous code (which has been deleted by a force push), he writes the code like this:
```
1. Get a Unicode string with the 40 bytes data. // Can raise exception on some bytes that is not UTF-16.
2. Remove any null characters at the end. // Assume that unused bytes are 0x00, which is a strong assumption.
```

However, the behavior of file `LoadOptionsClass.h` is different. The main difference is that, the type of archive name is `FixedString`, which behaves like a Pascal string. The content of unused bytes are **undefined**, which means it can be anything. One can't assume that these unused bytes are '0x00', which is the main problem of his previous code.

SadPencil:
> 
> 
> > > > ```cs
> > > > public static string GetSaveGameName(Stream stream)
> > > >         {
> > > >             var buffer = new byte[0x40];
> > > > ```
> > > > 
> > > > 
> > > > Why would you set the limitation of the length of the saving game name while the game executable supports a much longer name?
> > > 
> > > 
> > > sav file have a fixed block to save name.
> > > at 0x09c0 and length is 0x40.
> > > Actually, I don't know the other at 0x08d0 and there was no length limit.
> > 
> > 
> > Which means, your code brings a stronger assumption on the `.sav` file: The name block is fixed to `0x40` and when a null character (0x00 0x00) is set, the rest bytes in the name block must also be `0x00` ( or any other bytes that doesn't raise `ArgumentException` ).
> > Well, in my opinion, a weaker assumption is better than a stronger one. Or at least specify the assumption in the comment.
> 
> See 'SavegameInformation' as Alexb discovered here :
> https://github.com/Ares-Developers/YRpp/blob/master/LoadOptionsClass.h#L95

According to issue https://github.com/CnCNet/xna-cncnet-client/issues/164, the archive file is OLE Compound Document format.
The offset of the string is NOT ALWAYS at 0x09c0 or 0x08d0. It varies from platforms and programs. One should not assume a constant offset.

PR #217, opened 2021-02-20
Change the behavior of reading archive name with OpenMcdf

SadPencil:
Change the behavior of reading archive name with OpenMcdf to solve the cross-platform issue on Wine.

This PR fixed https://github.com/CnCNet/xna-cncnet-client/issues/164. It also closed https://github.com/CnCNet/xna-cncnet-client/pull/213 which is obsolete.

image


Starkku (approved):
Code is fine and this does fix the save game file parsing being unreliable. Adds another 3rd party library as a dependency, but probably justifiable in this case since writing a bespoke CDF parser would be quite a lot of work.

PR #219, opened 2021-02-23
Make client FPS user-configurable

Rampastring:
This PR does three things:

1) Makes client FPS user-configurable through `[Video] ClientFPS=`. Defaults to 60. This can be useful for users that have high refresh rate monitors and for people who want to lower the client's CPU usage at the expense of input smoothness.

2) Fixes MainMenuDarkeningPanel's alpha rate so it does not depend on the client's frame rate

3) Normalizes the alpha rate to be the same for MainMenuDarkeningPanel and the common DarkeningPanel


Starkku (approved):
Code looks good and the feature seems to work as expected.

hellpanderrr:
This feature seems to be working, but is it possible for it to be seen in the Client's options dialog?

PR #222, opened 2021-04-27
Multi Language Support To The Client And Increase Security.

AlperenFural:
- For Now, Multiple Language Support Has Only Added In The Headings In The Settings Panel
- Multi-Language Support Currently Includes Only Russian, English And Turkish.
- Game Broadcast Channel Channel Password Feature Added.
- The passwords in the rooms that were established were sent to the Game Broadcast Channel

### DIFF
--- a/ClientCore/Settings/UserINISettings.cs
+++ b/ClientCore/Settings/UserINISettings.cs
@@ -81,7 +81,6 @@ namespace ClientCore
             DragDistance = new IntSetting(iniFile, OPTIONS, "DragDistance", 4);
             DoubleTapInterval = new IntSetting(iniFile, OPTIONS, "DoubleTapInterval", 30);
             Win8CompatMode = new StringSetting(iniFile, OPTIONS, "Win8Compat", "No");
-            Language = new IntSetting(iniFile, OPTIONS, "Language", 0);
 
             PlayerName = new StringSetting(iniFile, MULTIPLAYER, "Handle", string.Empty);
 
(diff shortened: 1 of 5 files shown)


AlperenFural:
Screenshot_1
Screenshot_2

AlperenFural:
Şimdilik Sorunsuz Çalışıyor. Ama İstemcide Çevrilecek Çok Fazla Kelime Var.

pzhlkj6612:
Hi, I think these changes should be put into two PRs. One for "security" and the other for "multilingual UI".

AlperenFural:
I just thought of doing these for yr. unfortunately I don't have time to translate for all versions.

Metadorius:
This is not the desired way of doing translations. You're basically hardcoding another set of strings into the game code, which is the worst possible solution. Also please, don't translate into the language you don't know, that Russian is broken.

AlperenFural:
In The Initial Phase Only This Code Build Is Not Finished Yet. And I Added The Russian Language For Trial Purposes Only. Of course I will not translate all of everything.

Metadorius:
You're going the completely wrong way. There is already a branch with a base for translation feature that has to be done via reflection, but to be done properly it needs quite a a bit of effort to make. We are **not** going to do it via hardcode.

AlperenFural:
The Room Codes set up in the lobby are now written on the cncnet-yr-games channel, so the channel key must be put on the cncnet-yr-games channel. Increase the authority of administrators so that only the administrator can see them.

AlperenFural:
The channel key must be added to the cncnet-yr-games channel, so otherwise everyone will be able to see the set room passwords.

Metadorius:
Passwords should never be leaked into a public channel. You don't seem to really understand the concept of security. Also the "multi-language support" isn't really a support, you're just changing hardcoded strings.

PR #226, opened 2021-05-03
Fix auto-ready option keeping ready status on even while in-game

Starkku:
Auto-ready checkbox, when toggled on, used to keep players' ready status on even while they were in-game. With this fix, the status should be stripped when starting a game but re-instated after exiting the game.


alexp8:
I have downloaded and tested this fix with others in the YR client. Looks good to me to merge.

The only defect I have found is an edge case where the host picks a custom map that a player does not have downloaded. The player will remain ready and the host can start the game, the player without the map their client then crashes.

PR #227, opened 2021-05-21
the ability to join another user

devo1929:
This creates the ability to join another user from global player list or private message window player lists (friends or all players).

### DIFF
--- a/DXMainClient/DXGUI/Multiplayer/CnCNet/PasswordRequestWindow.cs
+++ b/DXMainClient/DXGUI/Multiplayer/CnCNet/PasswordRequestWindow.cs
@@ -9,10 +9,11 @@ using System;
 
 namespace DTAClient.DXGUI.Multiplayer.CnCNet
 {
-    public class PasswordRequestWindow : XNAWindow
+    internal class PasswordRequestWindow : XNAWindow
     {
-        public PasswordRequestWindow(WindowManager windowManager) : base(windowManager)
+        public PasswordRequestWindow(WindowManager windowManager, PrivateMessagingWindow privateMessagingWindow) : base(windowManager)
         {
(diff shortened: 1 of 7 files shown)


DXMainClient/DXGUI/Multiplayer/CnCNet/PasswordRequestWindow.cs
```diff
@@ -74,7 +78,16 @@ private void Keyboard_OnKeyPressed(object sender, KeyPressEventArgs e)
...
+            {
                 WindowManager.SelectedControl = tbPassword;
+                if (!privateMessagingWindow.Enabled) return;
+                pmWindowWasEnabled = true;
+                privateMessagingWindow.Disable();
+            } 
+            else if(pmWindowWasEnabled)
+            {
+                privateMessagingWindow.Enable();
+            }
```
devo1929, line 90: The updates to this function cover the scenario where a user is attempting to join another user from within the private messaging window and the game they're attempting to join requires a password. These updates will "hide" the primate messaging window when the password request window opens. It will then re-open the private messaging window when the password request window is canceled. In the scenario where the user specifies a password and clicks OK, the private messaging window is NOT reopened (see the OK btn handler function).

DXMainClient/DXGUI/Multiplayer/CnCNet/PasswordRequestWindow.cs
```diff
@@ -87,6 +100,7 @@ private void BtnOK_LeftClick(object sender, EventArgs e)
             if (string.IsNullOrEmpty(tbPassword.Text))
                 return;
 
+            pmWindowWasEnabled = false;
```
devo1929, line 103: This prevents the private messaging window from being reopened on password window close.

Issue #51, opened 2017-05-20
Clicking on someones name shows where they are


GrantBartlett:
It was quite a nice little feature back in the very original client, seeing where other players are.  If they're in a game or in the lobby etc.

It would be quite nice to see this in the core client again.

Rampastring:
Good idea. Although rather than clicking on someone's name, I'd prefer if it was hover-on info (like the game information) or something like that. But I'll implement it in some way once I can.

Martin-Veskilt:
Some players have been requesting this feature in the YR community.

Metadorius:
Thanks for heads up, we're aware of the requests.

devo1929:
FYI
https://github.com/CnCNet/xna-cncnet-client/pull/228

That PR provides the ability to join another user. It doesn't accomplish the specific request here, but it already has the necessary code to "find" the game that the selected user is in. Thus, allowing you to quickly add this feature as well.

PR #228, opened 2021-05-22
the ability to join another user

devo1929:
This creates the ability to join another user from global player list or private message window player lists (friends or all players).

When selecting a player from the various lists, the option to "Join" will appear in the right click menu. When this is clicked, the first thing that happens is it scans the list of currently available games to find which game that specific user might be in.

- If the selected user is not in a game, an error message is written.
- If the found game for the selected user is already started or locked, an error message is written.
- If the selected user happens to sign out when "Join" is selected, an error message is written that the selected user is no longer available.

All error messages are written to the "message view/list" based on their current context.

- If the Join action is clicked while in the CnC Lobby, errors are written to the CnC message board.
- If the Join action is clicked while in the private messaging window, errors are written to the private messaging window.


DXMainClient/DXGUI/Multiplayer/CnCNet/PasswordRequestWindow.cs
```diff
@@ -74,7 +78,16 @@ private void Keyboard_OnKeyPressed(object sender, KeyPressEventArgs e)
...
             if (Enabled)
+            {
                 WindowManager.SelectedControl = tbPassword;
+                if (!privateMessagingWindow.Enabled) return;
+                pmWindowWasEnabled = true;
+                privateMessagingWindow.Disable();
+            } 
+            else if(pmWindowWasEnabled)
+            {
+                privateMessagingWindow.Enable();
```
devo1929, line 89: The updates to this function cover the scenario where a user is attempting to join another user from within the private messaging window and the game they're attempting to join requires a password. These updates will "hide" the primate messaging window when the password request window opens. It will then re-open the private messaging window when the password request window is canceled. In the scenario where the user specifies a password and clicks OK, the private messaging window is NOT reopened (see the OK btn handler function).

This was all necessary due to the way in which the windows were constructed. The password request window is a child of the lobby, but the private messaging window is a child of the global window/sibling of the lobby. If the password request window is opened while the private messaging window is already open, the password request window will open behind the private messaging window.

The window structure could be altered to allow for the password request window to appear on top of the private messaging window, but would require much more risky changes.

DXMainClient/DXGUI/Multiplayer/CnCNet/PasswordRequestWindow.cs
```diff
@@ -87,6 +100,7 @@ private void BtnOK_LeftClick(object sender, EventArgs e)
             if (string.IsNullOrEmpty(tbPassword.Text))
                 return;
 
+            pmWindowWasEnabled = false;
```
devo1929, line 103: This prevents the private messaging window from being reopened on password window close.

devo1929:
This PR also completes some groundwork for this issue:
https://github.com/CnCNet/xna-cncnet-client/issues/51

DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetLobby.cs
```diff
@@ -645,8 +649,17 @@ private string CanJoinGameByIndex(int gameIndex)
...
+            return null;
+        }
+        
+        /// <summary>
+        /// Determines whether or not a game is join-able.
+        /// If not, an error message is returned. Otherwise, null.
+        /// </summary>
+        /// <param name="hg"></param>
+        /// <returns></returns>
+        private string CanJoinGame(HostedCnCNetGame hg)
```
Metadorius, line 661: I am not actually sure the method's code does what's written in method name. I mean, I would assume something named `CanJoinGame` should simply return a `bool`.

Metadorius: The code using it seems kinda counterintuitive because of that too.

Metadorius (changes requested):
Thanks for your contribution. The code style looks pretty good, the code is well documented. There's one small nitpick in how the `CanJoinGame` method is done though. @alexp8 is gonna test the feature a bit later and we could merge if there are no more issues.

DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetLobby.cs
devo1929: I can update this method name, if you want. I was just modeling it after the other index based method of `private string CanJoinGameByIndex(int gameIndex)` that already exists. As my first PR, I didn't want to introduce more risk than necessary by making more changes than I had to, while also balancing matching existing naming conventions.

devo1929: perhaps they become `GetJoinGameErrorByIndex(..)` and `GetJoinGameError(..)`?

devo1929: @Metadorius see latest commit

devo1929: @Metadorius I'm glad you made me look at this a little closer. I believer there may have been some issues with the previous code in that the index based version was doing some non-index based checks that need to be considered for this function as well. I've added another commit that creates a base function to check global states and is called by both the index based version and the `HostedCnCNetGame` based version.

Metadorius (approved):
Looks good now. Let's wait on those tests and merge if other maintainers won't mind.

alexp8:
Tested the feature and looks good:

Scenarios passed:

`try joining locked game
try joining unlocked game
try joining game with password
try joining game of person who blocked you?
try joining game of person who is not in a game
try joining game of person who is the host
try joining game of person who is not the host
try joining game of person via friends list
try joining game of person via player list
try joining game of person via lobby list
try joining someone's unlocked game then leave and try rejoining`

cc @Metadorius

PR #229, opened 2021-05-30
dependency injection

devo1929:
This implements dependency injection, using `GameClass` as the location where types are defined.


DXMainClient/DXMainClient.csproj
```diff
@@ -83,7 +83,7 @@
     <DefineConstants>TRACE;DEBUG;XNA</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
-    <TargetFrameworkVersion>v4.0</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
```
devo1929, line 86: I don't know if there is/was a specific reason for targeting 4.0, but it was necessary to update to at least 4.5 for Autofac compatibility.

Metadorius: Well, bad news but we actually have to use v4.0 because of WinXP support: https://docs.microsoft.com/en-us/dotnet/framework/install/on-windows-xp

devo1929: ah, that makes sense, I'll see what I can do...

devo1929:
Closing for now, until I can make it compatible.

PR #231, opened 2021-06-18
Azure pipelines

devo1929:
This contains two commits:

1. adds XNS dlls to sc so that they can be used in the remote pipeline
2. adds a yaml file for azure build pipelines

When the pipeline runs, it will use the 3 primary build scripts (YR, Ares, and TS) to create separate zip files. These zip files are then published to Azure for consumption, whether manually or another build pipeline.


Metadorius:
Assuming the artifacts are posted somewhere, how can modders and developers access them?

devo1929:
How familiar are you with Azure DevOps? I currently have a pipeline set up in my own personal free instance that looks at my fork of the repo. Ideally, there would be a cncnet owned DevOps instance where you could invite any developers you want to be able to access the pipelines, and the pipelines would look at your own master/stable/etc... branches. I believe you could also make the artifacts public for consumption.

I'm more than happy to assist in getting this set up if someone wants to create a base Azure Devops for cncnet and invite me. Otherwise, I can try to assist in other ways.

If you left it at this pipeline alone, it would be up to other devs to download these artifacts. However, you could always set up additional pipelines for the other packages (YR) to consume these artifacts for their own pipelines, and then auto create new versions of them (YR 8.4?).

Metadorius:
I am not really familiar with Azure DevOps, currently I've only used GitHub Actions to compile and publish nightly builds of Phobos (my other project aimed at expanding YR engine).

> I'm more than happy to assist in getting this set up if someone wants to create a base Azure Devops for cncnet and invite me. Otherwise, I can try to assist in other ways.
Think we need to ask what are @Rampastring's and @Starkku's opinions on this.

devo1929:
I'm not that familiar with Github actions myself, but I'm certainly not opposed to porting over what I have so far.

devo1929:
A github actions alternative:
https://github.com/CnCNet/xna-cncnet-client/pull/233

Metadorius:
The GH Actions one is indeed more preferable. Merged that one, closing this.

PR #232, opened 2021-06-18
Feature/disable multiple factions

alexp8:
Allow value **DisallowedSideIndex** to take a comma separated input to disable multiple sides.

Previously this flag would allow to disable just one faction, now can disable multiple factions via comma separated.

Example: https://github.com/CnCNet/cncnet-yr-client-package/blob/master/Resources/GameOptions.ini#L285


DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyCheckBox.cs
```diff
@@ -53,7 +53,7 @@ public class GameLobbyCheckBox : XNAClientCheckBox
         /// The side index that this check box disallows when checked.
         /// Defaults to -1, which means none.
         /// </summary>
-        public int DisallowedSideIndex { get; set; } = -1;
+        public string[] DisallowedSideIndex;
```
Rampastring, line 56: Change this to an `int` array and parse it as such when parsing the value of the key. Then the actual logic doesn't need to do any type conversions.

DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyCheckBox.cs
```diff
@@ -149,11 +152,16 @@ public void ApplyMapCode(IniFile mapIni, GameMode gameMode)
...
         public void ApplyDisallowedSideIndex(bool[] disallowedArray)
         {
-            if (DisallowedSideIndex < 0)
+            if (DisallowedSideIndex == null || DisallowedSideIndex.Length == 0)
                 return;
 
-            if (Checked != reversed)
-                disallowedArray[DisallowedSideIndex] = true;
+            if (Checked != reversed) {
+                for (int i = 0; i < DisallowedSideIndex.Length; i++) {
```
Rampastring, line 159: The client code typically has `{` on their own lines. Change these to match.

Rampastring (changes requested):

DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyCheckBox.cs
alexp8: thanks Rampa, I will update the PR accordingly

alexp8:
I have updated the PR with requested changes, let me know if any further change required. @Rampastring @Metadorius 

Use default `DisallowedSideIndex` to disable one faction
Or use new `DisallowedSideIndices` to disable one or more factions.

Thanks,
Burg

DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyCheckBox.cs
```diff
@@ -101,7 +101,11 @@ public override void ParseAttributeFromINI(IniFile iniFile, string key, string v
                     UserChecked = checkedValue;
                     return;
                 case "DisallowedSideIndex":
-                    DisallowedSideIndex = Conversions.IntFromString(value, DisallowedSideIndex);
+                     DisallowedSideIndices = new int[]{int.Parse(value)};
```
Metadorius, line 104:
```suggestion
                    DisallowedSideIndices = new int[]{ int.Parse(value) };
```

Metadorius (approved):
Looks good now.

DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyCheckBox.cs
```diff
@@ -101,7 +101,11 @@ public override void ParseAttributeFromINI(IniFile iniFile, string key, string v
                     UserChecked = checkedValue;
                     return;
                 case "DisallowedSideIndex":
-                    DisallowedSideIndex = Conversions.IntFromString(value, DisallowedSideIndex);
+                    DisallowedSideIndices = new int[]{ int.Parse(value) };
+                    return;
+                case "DisallowedSideIndices":
+                    string[] sides = value.Split(',');
+                    DisallowedSideIndices = Array.ConvertAll(value.Split(','), (s) => { return Conversions.IntFromString(s, -1); });
                     return;
```
Metadorius, line 109: Actually with the current way it's implemented the tags used will conflict when both are used. I think appending should be done instead of replacement.

Metadorius (changes requested):
Missed some stuff that could break in some cases, sorry for not spotting this earlier on.

DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyCheckBox.cs
Rampastring: Is there a reason to ever use both though?

Metadorius: That's for backwards compatibility and to make it foolproof.

alexp8:
@Metadorius updated, code is backwards compatible and both flags can be used at the same time

Metadorius (approved):

PR #235, opened 2021-06-22
Add dependency injection

devo1929:


DXMainClient/Domain/NullDiscordHandler.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        public void Connect()
+        {
+        }
+
+        public void Disconnect()
+        {
+        }
+    }
+}
```
Metadorius, line 37: I am not sure if this is a valid approach at all. Moreover it seems like it needs restart in order to recreate the handler.

Metadorius (changes requested):
Not sure of `NullDiscordHandler` approach at all. What's the purpose of a separate handler for doing nothing?

DXMainClient/DXGUI/GameClass.cs
```diff
@@ -143,10 +152,47 @@ protected override void Initialize()
...
+            container.RegisterSingleton<PrivateMessagingPanel>();
+            container.RegisterSingleton<LANLobby>();
+            container.RegisterSingleton<CnCNetGameLobby>();
+            container.RegisterSingleton<CnCNetGameLoadingLobby>();
+            container.RegisterSingleton<CnCNetLobby>();
+            container.RegisterSingleton<GameInProgressWindow>();
+            container.RegisterSingleton<SkirmishLobby>();
+            container.RegisterSingleton<MainMenu>();
+            container.RegisterSingleton<MapLoader>();
+            container.RegisterSingleton(() =>
```
devo1929, line 186: @Metadorius 
The reason for `NullDiscordHandler`, is because you cannot return a `null `value for this dependency using SimpleInjector. So, I created a dummy handler that would prevent usages of it from breaking the application, but nothing would actually happen. 

Regarding your comment about having to restart the client, you are correct. That would be a requirement in this case.

devo1929 (commented):
> Not sure of `NullDiscordHandler` approach at all. What's the purpose of a separate handler for doing nothing?

I left a comment on the exact line of code for why this class is necessary.

DXMainClient/DXGUI/GameClass.cs
Metadorius: Yes, but why empty handler? Can't that just be handled that from the `DiscordHandler` itself without introducing null handler and requiring restart?

devo1929: It could be. It was more for simplicity, I guess. If the dependency injection returns this empty handler, no other codes requires a change. If it returns a valid DiscordHandler, I was finding that nearly all of the public functions in the DiscordHandler required a null check on the internal client. It just felt very cluttered.

devo1929: Also, in the previous code, in `LoadingScreen.cs`, the DiscordHandler was only ever instantiated if Discord was enabled. Otherwise, it was null. Where it was attempted to be used could be easily null checked on the DiscordHandler instance itself. Like I said though, in my dependency injection, a dependency cannot be returned as null. So, this was just another solution.

PR #236, opened 2021-06-27
DiscordHandler cleanup.

Starkku:
- Remove unnecessary using statements.
- Remove unnecessary update method override.
- Do not require restart when changing Discord integration setting.

### DIFF
--- a/DXMainClient/Domain/DiscordHandler.cs
+++ b/DXMainClient/Domain/DiscordHandler.cs
@@ -1,40 +1,38 @@
 ﻿using ClientCore;
 using DiscordRPC;
 using DiscordRPC.Message;
-using DTAClient.Online;
 using Microsoft.Xna.Framework;
 using Rampastring.Tools;
 using Rampastring.XNAUI;
-using System;
-using System.Collections.Generic;
-using System.Globalization;
-using System.IO;
-using System.Net;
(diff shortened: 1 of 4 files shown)


DXMainClient/Domain/DiscordHandler.cs
```diff
@@ -68,23 +98,44 @@ public override void Initialize()
...
-            base.Dispose(disposing);
-        }
+            if (client == null || !client.IsInitialized)
+                return;
+
+            // HACK warning
+            // Currently DiscordRpcClient does not appear to have any way to reliably disconnect and reconnect using same client object.
+            // Deinitialize does not appear to completely reset connection state & resources and any attempts to call Initialize afterwards will fail.
+            // A hacky solution is to dispose current client object and create and initialize a new one.
+            InitializeClient(); //client.Deinitialize();
```
Metadorius, line 135: Have you checked whether this is fixed in the new versions of the library?

Metadorius (commented):
Looks good, I assume it works as before in other parts except the no restart part.

DXMainClient/Domain/DiscordHandler.cs
Starkku: Yes, the issue is still present.

Metadorius (approved):

PR #238, opened 2021-07-04
Game List Sort and Filters

devo1929:
This provides options for a few things:

- sort alphabetically
- show friend games only
- hide locked games
- hide passworded games
- hide incompatible games
- max player count

It creates a new panel that appears on top of the game list. This panel is where you'll find controls to modify the filters above.

There are two buttons added above the Game List as well. These buttons are a toggle button to quickly enable alphabetical sort and a button to quickly open the game filters panel on top of the game list. This filters toggle button will appear toggled "on" when there are filters enabled that are not the defaults.

All filter settings are stored to the settings ini (ra2md.ini).

The icons required for these new toggle buttons are located in the cncnet-yr-client-package:
https://github.com/CnCNet/cncnet-yr-client-package/pull/46

### DIFF
--- a/DTAConfig/OptionsWindow.cs
+++ b/DTAConfig/OptionsWindow.cs
@@ -68,7 +68,6 @@ namespace DTAConfig
             var updaterOptionsPanel = new UpdaterOptionsPanel(WindowManager, UserINISettings.Instance);
             updaterOptionsPanel.OnForceUpdate += (s, e) => { Disable(); OnForceUpdate?.Invoke(this, EventArgs.Empty); };
 
-
             optionsPanels = new XNAOptionsPanel[]
             {
                 displayOptionsPanel,
--- a/DTAConfig/OptionsWindow.cs
+++ b/DTAConfig/OptionsWindow.cs
@@ -27,7 +27,6 @@ namespace DTAConfig
         private ComponentsPanel componentsPanel;
 
         private DisplayOptionsPanel displayOptionsPanel;
-        private GameFilterOptionsPanel gameFilterOptionsPanel;
         private XNAControl topBar;
 
         private GameCollection gameCollection;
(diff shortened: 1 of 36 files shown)


Metadorius:
Some screenshots would definitely help.

devo1929:
Screenshot 2021-07-04 143857

Screenshot 2021-07-04 143931

Rampastring:
I think this new tab should be next to the CnCNet tab as it's still strictly CnCNet-related.

Metadorius:
I think this is the case where a new control wouldn't hurt, perhaps something like a dropdown with other controls?

Rampastring:
That's actually a great idea, then there wouldn't be a need for it to reside as an extra tab in the options menu. Basically a button that shows any XNAControl instance below itself when clicked on should be easy to implement and a good addition to the `XNAUI` library.

devo1929:
I'd be willing to build something like that. I don't know if I have access to the XNAUI repo, currently, but I can make it in cncnet client for now.

Metadorius:
> I'd be willing to build something like that. I don't know if I have access to the XNAUI repo, currently, but I can make it in cncnet client for now.

It's in Rampastring's repos.

devo1929:
I'm typically not a fan of using drop downs or popups for more complex UIs. If it were simply a multi select, then I think that type of UI would work for this (the max count filter ruins that). 

What if it were a child panel of the GameListBox itself? IE, clicking on the new filter button would "swap" the list view for a filter options view. This filter options view would then have the Save and Cancel buttons as it currently does, which would then toggle back to the game list.

devo1929:
I made some changes. Instead of the options window route, clicking the Filters button will show a panel that sits on top of the game list. The contents look the same as the screen shot above when it was in the options window, except that the sort alpha checkbox was removed, because it's quicker to just use the sort alpha button that's right above this panel.
Screenshot:

Screenshot 2021-07-06 125121

alexp8:
looks great, please approve and merge @Metadorius

DTAConfig/OptionsWindow.cs
```diff
@@ -68,6 +68,7 @@ public override void Initialize()
             var updaterOptionsPanel = new UpdaterOptionsPanel(WindowManager, UserINISettings.Instance);
             updaterOptionsPanel.OnForceUpdate += (s, e) => { Disable(); OnForceUpdate?.Invoke(this, EventArgs.Empty); };
 
+
```
Metadorius, line 71:
```suggestion
```

Metadorius (approved):
The code quality is fine and the implementation is pretty good.

PR #239, opened 2021-07-06
Update Rampastring.Tools and Rampastring.XNAUI

Rampastring:
Updates Tools and XNAUI to enable their latest functionality for the client. Also modifies the code to be compatible with latest XNAUI and improves the code of `XNAChatTextBox.HandleKeyPress` to have clearer function exit points.


ClientGUI/XNAChatTextBox.cs
```diff
@@ -35,10 +32,10 @@ protected override bool HandleKeyPress(Keys key)
...
                 if (currentNode == null)
                 {
-                    if (enteredMessages.First == null)
-                        return true;
-
-                    currentNode = enteredMessages.First;
+                    if (enteredMessages.First != null)
+                    {
+                        currentNode = enteredMessages.First;
+                    }
```
Metadorius, line 38: Not sure why those are braced.
```suggestion
                        currentNode = enteredMessages.First;
```

PR #243, opened 2021-07-12
Feature/add map name to custom maps

alexp8:
Custom maps downloaded from CnCNet will contain the name of the map in the filename to allow easier navigation of custom maps.

Approach:
Leave map upload as is, uploading the file with sha1 sum as the filename.
After map is downloaded, rename the map with the map name in the filename.


devo1929:
two things:

1. I believe someone had raised the concern with the filenames _possibly_ being too long. If this is still a concern, we could always limit the sha to the first 6-10 characters in the filename.
2. There are a lot of locations in the code where the filename is being constructed and we're concatenating "mapName_sha" for each map. It'd be nice if there was a shared "utility" function to construct this name for a passed in mapName and sha, so that it's one location in the code constructing it every time, should it ever need to change again.

alexp8:
> two things:
> 
> 1. I believe someone had raised the concern with the filenames _possibly_ being too long. If this is still a concern, we could always limit the sha to the first 6-10 characters in the filename.
> 2. There are a lot of locations in the code where the filename is being constructed and we're concatenating "mapName_sha" for each map. It'd be nice if there was a shared "utility" function to construct this name for a passed in mapName and sha, so that it's one location in the code constructing it every time, should it ever need to change again.

1. We would want to display the full sha1 sum in the file name to guarantee uniqueness, I'm not aware of any issue where the filenaame could be too long.
2. I can create utility function.

alexp8:
I pushed the utility function

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
```diff
@@ -1534,8 +1544,8 @@ private void MapSharer_HandleMapDownloadFailed(SHA1EventArgs e)
 
         private void MapSharer_HandleMapDownloadComplete(SHA1EventArgs e)
         {
-            Logger.Log("Map " + e.SHA1 + " downloaded, parsing.");
-            string mapPath = "Maps/Custom/" + e.SHA1;
+            Logger.Log("Map " + e.MapName + "_" + e.SHA1 + " downloaded, parsing.");
+            string mapPath = "Maps/Custom/" + e.MapName + "_" + e.SHA1;
```
devo1929, line 1548: two here for use of your new utility function

alexp8: done

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
```diff
@@ -973,11 +980,15 @@ private void ApplyGameOptions(string sender, string message)
                 AddNotice("The game host has changed ProtocolVersion to " + protocolVersion);
             }
 
+            string mapName = parts[partIndex + 8];
+
+
             GameMode currentGameMode = GameMode;
             Map currentMap = Map;
```
Metadorius, line 987:
```suggestion
            string mapName = parts[partIndex + 8];
            GameMode currentGameMode = GameMode;
            Map currentMap = Map;
```

DXMainClient/Domain/Multiplayer/CnCNet/MapSharer.cs
```diff
@@ -363,20 +365,28 @@ private static void Download(object details)
...
                     Download(array);
                 }
             }
         }
 
-        private static string DownloadMain(string sha1, string myGame, out bool success)
+        public static string GetMapFileName(string sha1, string mapName)
+        {
+            return mapName + "_" + sha1;
+        }
```
Metadorius, line 381:
```suggestion
        public static string GetMapFileName(string sha1, string mapName)
            => mapName + "_" + sha1;
```

Metadorius (approved):
Looks good. Should use expression bodies where it's applicable though. May also refactor the file to use them if you feel like it and there are other cases where it's not applied yet.

PR #244, opened 2021-07-12
Fix DisallowedSideIndex bug

devo1929:
@Metadorius @alexp8


DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyCheckBox.cs
```diff
@@ -101,13 +102,10 @@ public override void ParseAttributeFromINI(IniFile iniFile, string key, string v
                     UserChecked = checkedValue;
                     return;
                 case "DisallowedSideIndex":
```
devo1929, line 104: I removed the body from the DisallowedSideIndex case statement and just let it fall through. The logic in DisallowedSideIndices should be able to handle both cases.

PR #245, opened 2021-07-21
Favorite maps list

devo1929:
This allows any user to mark a map as a favorite when in any game lobby. 

Favorite maps are stored as a comma separated list of SHAs in the user settings ini file.

There is a new star icon button on the map preview box in the upper right hand corner. When the currently selected map is NOT a favorite, the star icon button appears "empty". When the current selected map IS a favorite, the star icon button appears "full".

The game mode drop down has an item inserted into the first position titled "Favorites". When selected, only maps that are marked as favorites will appear in the map list.

I will post screen shots.

### DIFF
--- a/ClientCore/ClientCore.csproj
+++ b/ClientCore/ClientCore.csproj
@@ -260,6 +260,7 @@
     <Compile Include="SavedGameManager.cs" />
     <Compile Include="Settings\IIniSetting.cs" />
     <Compile Include="Settings\IntRangeSetting.cs" />
+    <Compile Include="Settings\StringListSetting.cs" />
     <Compile Include="Settings\UserINISettings.cs" />
     <Compile Include="Settings\BoolSetting.cs" />
     <Compile Include="Settings\DoubleSetting.cs" />
--- a/ClientCore/Settings/StringListSetting.cs
+++ b/ClientCore/Settings/StringListSetting.cs
@@ -1,0 +1,32 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using Rampastring.Tools;
+
+namespace ClientCore.Settings
+{
+    /// <summary>
+    /// This is a setting that can be stored as a comma separated list of strings.
+    /// </summary>
+    public class StringListSetting : INISetting<List<string>>
+    {
+        public StringListSetting(IniFile iniFile, string iniSection, string iniKey, List<string> defaultValue) : base(iniFile, iniSection, iniKey, defaultValue)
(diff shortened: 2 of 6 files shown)


devo1929:
image
image
image
image

alexp8 (approved):
awesome stuff Devo, i'm excited to test this out!

Starkku:
This implementation currently has a number of issues, several stemming from the decision to put the Favourites list at top of the game mode dropdown.

1. The internal game modes list (`GameModes`) is frequently accessed via selected item index of the dropdown (`ddGameMode`). Because of the addition of Favourites item at top of the dropdown, it now treats the favourites list as the first real game mode (including allowing launching maps on it, even those that normally do not belong to it) minus listing its maps and the indices are no longer in sync and it throws off the map listings and outright crashes client if you select last game mode in the dropdown.
2. The favourites list is also now the default 'game mode' if no existing settings were loaded. This could *debatably* be preferrable but it is probably not desirable given that the favourites list can be empty, thus resulting in no real game mode or map being selected on start.
3. Likewise if a saved setting (such as for skirmish) has the first real game mode selected (which is not really available in dropdown anymore thanks to issue number 1), client will get stuck on startup.
4. Saving the favourited maps does not function because the value of the setting `FavoriteMapSHAs` is not changed when adding/removing favourites, merely the items in the collection held as its value are altered and thus the `Set` method is never called.

Depending on how one might want to handle these issues, something to keep in mind is that maps can appear in more than one game mode.

Metadorius:
If I recall correctly, same map could be located in different game modes, each of which would apply different INIs when selected. What if the map is indeed located in two different game modes and is favorite? Which game mode would it have in favorites?

Metadorius:
I suggest a better implementation - a "Show only favorites" checkbox which would work on any game mode.

devo1929:
I believe you are correct in the out of sync index of the Game Mode drop down. I will definitely look into this further and see what I can come up with.

I will give some more thought on how the "default" scenarios should work, like NOT selecting Favorites first, if there are no favorite maps.

I do like the idea of having a checkbox for "favorites only" so that it can be used on any game mode, but that also then requires that a person remembers which game mode a particular map is assigned to. If you're a more simple player and only know that you like a few specific maps, finding them might be a chore. Especially, because one of the primary ways for a person to mark a map as a favorite might be as a non-host player. At that point, you're simply marking the current map selected by the host as a favorite in the preview box and might not think to note the game mode it's assigned to.

PR #246, opened 2021-07-21
Revert "Update Rampastring.Tools and Rampastring.XNAUI"

devo1929:
This reverts commit 36e111f48ec4a0e2127876b28b22cb44e94dd352.


devo1929:
I'm not sure what all went into the XNAUI updates, but reverting this commit fixes the game list hover bug (the game info doesn't appear). 

I'm not necessarily assuming this will get merged, but I figure I can create it, in case it actually is the right thing to do, for now.

Rampastring:
I'll investigate the bug when I get the time. Might take up to 1½ weeks in the worst case though.

devo1929:
> I'll investigate the bug when I get the time. Might take up to 1½ weeks in the worst case though.

@Rampastring is there anything in these changes that absolutely needs to go out in this next release? Are we ok to revert it for now and re-merge it back in when you have time to investigate?

Rampastring:
There shouldn't be anything absolutely critical, the updated references mostly enable some new methods.

I'm not a huge fan of having revert commits in the git log though, maybe you could ship the next bunch of YR binaries from your own repo if you're planning on doing an update very soon?

PR #249, opened 2021-08-11
Improved map loading - caching

devo1929:


devo1929:
The workflow:

- Load previously cached custom maps
- Iterate over all local map files, processing each in their own Task
- Calculate SHA of currently processed Map and attempt to locate it in the cache with SHA
- If not contained in the cache, load from .map file and add to cache
- Clean up cache, removing any old cache entries that were not found during local map iteration
- Save new state of cache
- Use current cache for UI

devo1929:
Some benchmarks loading 1878 custom maps before and after on 4-core i7 cpu, 32gb of memory, m2 SSD:

Before: 22-24 seconds
After: 1-2 seconds

DXMainClient/Domain/Multiplayer/Map.cs
```diff
@@ -107,72 +120,96 @@ public Map(string path, bool official)
...
         int Bases = -1;
 
+        [JsonProperty]
         string[] localSize;
 
+        [JsonProperty]
         string[] actualSize;
 
-        IniFile mapIni;
+        private IniFile CustomMapIni;
```
Metadorius, line 190: @Rampastring is this OK with our code style?

DXMainClient/Domain/Multiplayer/Map.cs
```diff
@@ -346,39 +387,41 @@ public Point MapPointToMapPreviewPoint(Point mapPoint, Point previewSize, int le
...
+        /// This function provides the ability to load when needed.
+        /// </summary>
+        /// <returns></returns>
+        private IniFile GetCustomMapIniFile()
         {
-            // TODO necessary for custom map caching
-            throw new NotImplementedException();
+            if (CustomMapIni != null) return CustomMapIni;
+            
+            CustomMapIni = new IniFile {FileName = CustomMapFilePath};
```
Metadorius, line 399:
```suggestion
            CustomMapIni = new IniFile { FileName = CustomMapFilePath };
```

DXMainClient/Domain/Multiplayer/Map.cs
```diff
@@ -107,72 +120,96 @@ public Map(string path, bool official)
...
         string[] localSize;
 
+        [JsonProperty]
         string[] actualSize;
 
-        IniFile mapIni;
+        private IniFile CustomMapIni;
 
+        [JsonProperty]
+        private string CustomMapFilePath;
```
Metadorius, line 193: Same here.

Metadorius (changes requested):
The code looks good, I don't see any glaring issues. I've pointed out small bit of code style issues though.

DXMainClient/Domain/Multiplayer/Map.cs
Rampastring: Private fields should be in lowercase.

devo1929:
> Some benchmarks loading 1878 custom maps before and after on 4-core i7 cpu, 32gb of memory, m2 SSD:
> 
> Before: 22-24 seconds
> After: 1-2 seconds

I should also add that the _first_ time starting the client with this new code resulted in about 6-8 seconds load time, compared to the previous 22-24 seconds. That was due to the cache not having been built yet, but there was still a significant gain due to the parallel loading.

DXMainClient/Domain/Multiplayer/Map.cs
devo1929: Resolved

DXMainClient/Domain/Multiplayer/Map.cs
devo1929: Resolved

DXMainClient/Domain/Multiplayer/Map.cs
devo1929: Resolved

devo1929:
@Metadorius all requested changes were made.

Metadorius (approved):

PR #250, opened 2021-08-19
Fix game information panel not showing up with latest XNAUI

Rampastring:
As the title says, fixes the game information panel not showing up with latest XNAUI by making the game information panel a child of the CnCNet / LAN lobby instead of the game list box.

I've confirmed the fix to work properly in DTA's fork of the client, but it's best to test it before merging anyway in case I made some mistake in porting over the code.


devo1929:
I pulled everything down locally and poked around. Game Info Panel works and nothing else seems to have any issues. 👍

PR #256, opened 2021-09-14
fix issue with null SHAs when loading from cache

devo1929:
move custom_map_cache
version custom_map_cache
fix issue with null SHAs when loading from cache

### DIFF
--- a/DXMainClient/Domain/Multiplayer/MapLoader.cs
+++ b/DXMainClient/Domain/Multiplayer/MapLoader.cs
@@ -15,10 +15,11 @@ namespace DTAClient.Domain.Multiplayer
     {
         public const string MAP_FILE_EXTENSION = ".map";
         private const string CUSTOM_MAPS_DIRECTORY = "Maps/Custom";
-        private const string CUSTOM_MAPS_CACHE = CUSTOM_MAPS_DIRECTORY + "/cache";
+        private static readonly string CUSTOM_MAPS_CACHE = ProgramConstants.ClientUserFilesPath + "custom_map_cache";
         private const string MultiMapsSection = "MultiMaps";
         private const string GameModesSection = "GameModes";
         private const string GameModeAliasesSection = "GameModeAliases";
+        private const int CurrentCustomMapCacheVersion = 1;
 
         /// <summary>
         /// List of game modes.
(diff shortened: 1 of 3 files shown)


DXMainClient/Domain/Multiplayer/MapLoader.cs
```diff
@@ -205,8 +211,16 @@ private void CacheCustomMaps(ConcurrentDictionary<string, Map> customMapCache)
             try
             {
                 var jsonData = File.ReadAllText(CUSTOM_MAPS_CACHE);
-            
-                return JsonConvert.DeserializeObject<ConcurrentDictionary<string, Map>>(jsonData) ?? new ConcurrentDictionary<string, Map>();
+
+                var customMapCache = JsonConvert.DeserializeObject<CustomMapCache>(jsonData);
+
+                var customMaps = customMapCache?.Version == CurrentCustomMapCacheVersion && customMapCache.Maps != null
+                    ? customMapCache.Maps : new ConcurrentDictionary<string, Map>();
```
devo1929, line 218: This should allow us to "force clear" someone's cache on an update, should there be an issue in the future or new data to be added to the cache.

DXMainClient/Domain/Multiplayer/MapLoader.cs
```diff
@@ -205,8 +211,16 @@ private void CacheCustomMaps(ConcurrentDictionary<string, Map> customMapCache)
...
+
+                foreach (var customMap in customMaps.Values)
+                    customMap.CalculateSHA();
```
Starkku, line 221: Calculating the SHA1 hash here for all cached maps and in `LoadCustomMaps` for all local maps effectively makes it so that client calculates the hash twice for all custom maps if they also exist in the cache. I am however unable to discern any performance impact from this, likely due to either operating system or hardware-level optimizations concerning file reading. Because of this and the fact that doing this in any other way would compromise the robustness of using SHA1 hash to identify maps in the first place, I don't believe this requires any action but I thought it was worth noting regardless.

Starkku (approved):
Looks good and addresses the issue. There's one thing I have noted in a [separate comment](https://github.com/CnCNet/xna-cncnet-client/pull/256#discussion_r708302178) though, that I don't believe requires action at this time so I'll approve and merge the changes.

PR #258, opened 2021-10-17
game list box improvements

devo1929:
### 1st commit: Maintain selection state on game list
The code originally attempted to maintain the selection state as games were being added and removed, but it was based on the variable HostedGames which is the entire list of games when it comes in from CnCNet. With game filtering now available, this is no longer the proper list of games to determine the selected index from. HostedGames is ALL games before they are filtered. I changed it to look directly at the Items list that IS the list that is rendered.

Furthermore, there was code to maintain this selection state in many locations (one at add, one at remove, etc..). I removed all of them and added it only the the Refresh function that is already called by all of them.

### 2nd commit: Show game info panel on click, instead of hover
As games are appearing in the list, it's nearly impossible to actually read the game information for a hovered game, as the game could easily move while sorting is being applied. Rather than hover, the panel is now shown when a game is selected. Also, the game info panel is always aligned at the top right corner of the game list box, rather than where you are selected. Same issue with games being added or removed, this panel previously moved up and down to "follow" the game that is selected (was hovered).
If you click on the chat box, the game list selection will be cleared and thus the panel will be hidden.

### DIFF
--- a/DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetLobby.cs
+++ b/DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetLobby.cs
@@ -220,6 +220,7 @@ namespace DTAClient.DXGUI.Multiplayer.CnCNet
             lbChatMessages.PanelBackgroundDrawMode = PanelBackgroundImageDrawMode.STRETCHED;
             lbChatMessages.BackgroundTexture = AssetLoader.CreateTexture(new Color(0, 0, 0, 128), 1, 1);
             lbChatMessages.LineHeight = 16;
+            lbChatMessages.LeftClick += (sender, args) => lbGameList.SelectedIndex = -1;
 
             tbChatInput = new XNAChatTextBox(WindowManager);
             tbChatInput.Name = nameof(tbChatInput);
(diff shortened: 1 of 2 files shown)


Rampastring:
> 2nd commit: Show game info panel on click, instead of hover

Might be better if this is made optional / customizable as the hover-on game info works well for the great majority of games that the client is used by. Only YR and possibly Mental Omega is too popular for it.

devo1929:
> > 2nd commit: Show game info panel on click, instead of hover
> 
> Might be better if this is made optional / customizable as the hover-on game info works well for the great majority of games that the client is used by. Only YR and possibly Mental Omega is too popular for it.

I think that any game that uses this client would benefit from it, not just YR. The client itself has the ability to sort games alphabetically. It's this sorting that causes a game to be inserted into the list at any position, rather than at the end of the list. Thus, the game you're hovering is likely to move a lot as games are being added and removed. Thus, the game info panel is constantly disappearing on you or changing as you're suddenly hovering a different game.

Metadorius:
> I think that any game that uses this client would benefit from it
Are you sure about that? Some active mods usually have max. of 3-5 games running concurrently, and this doesn't seem to be a better scenario in this case.

PR #265, opened 2021-11-06
Is ignored fix

devo1929:


DXMainClient/DXGUI/Multiplayer/CnCNet/PlayerContextMenu.cs
```diff
@@ -72,13 +72,19 @@ private void UpdateButtons(IRCUser ircUser, ChannelUser channelUser = null)
...
+            Items[2].Selectable = !isAdmin && !string.IsNullOrEmpty(ircUser.Ident);
             Items[3].Visible = isOnline && !string.IsNullOrEmpty(contextMenuData.inviteChannelName);
             Items[4].Visible = !contextMenuData.PreventJoinGame && isOnline;
 
             Items[1].Text = cncnetUserData.IsFriend(ircUser.Name) ? "Remove Friend" : "Add Friend";
+            Items[2].Text = string.IsNullOrEmpty(ircUser.Ident) ? "Checking For Block..." : 
+                cncnetUserData.IsIgnored(ircUser.Ident) ? "Unblock" : "Block";
             
-            Items[2].Text = cncnetUserData.IsIgnored(ircUser.Ident) ? "Unblock" : "Block";
+            GetIrcUserIdent(ident =>
```
devo1929, line 83: I made this change in a separate commit, because I don't know yet if this is a good idea or not.

What this will do is perform a "whois" command upon right-clicking any player, IF they do not currently have an Ident. I don't know if this is a bad idea or not, because of possible flooding. If this is acceptable, then if the user does not have an Ident, the experience will be:
- initially set the text of this option in the menu to "Checking for Block..." and will not be selectable
- upon "whois" reply, update the text to "Unblock" or "Block" appropriately and will be selectable

If this change is NOT acceptable, then the experience will be that this option will be initially labeled as "Block", because the "IsIgnored" call will return false due to a null Ident being passed in.

devo1929:
These changes bring back the "Block" item in the player context menu, whether or not the user has an Ident. If you select "Block" for a user that does not have an Ident, the client will perform a "whois" for their nickname to retrieve the Ident. Once acquired, the Block is successfully performed and the user in the player list is updated with their Ident for future requests.

alexp8 (approved):

PR #266, opened 2021-11-25
CnCNetLobby message right click context menu

devo1929:
Right click context menu for player-sent messages in the cncnet main lobby.

For now, this uses the existing functionality of the current player right-click context menu. However, it is a derived class leaving it open to other future menu options.

### DIFF
--- a/DXMainClient/DXGUI/Multiplayer/ChatListBox.cs
+++ b/DXMainClient/DXGUI/Multiplayer/ChatListBox.cs
@@ -52,17 +52,26 @@ namespace DTAClient.DXGUI.Multiplayer
 
         public void AddMessage(ChatMessage message)
         {
+            var listBoxItem = new XNAListBoxItem
+            {
+                TextColor = message.Color,
+                Selectable = true,
+                Tag = message
+            };
+            
             if (message.SenderName == null)
-                AddItem(Renderer.GetSafeString(string.Format("[{0}] {1}",
+            {
+                listBoxItem.Text = Renderer.GetSafeString(string.Format("[{0}] {1}",
                     message.DateTime.ToShortTimeString(),
-                    message.Message), FontIndex),
-                    message.Color, true);
+                    message.Message), FontIndex);
+            }
             else
             {
-                AddItem(Renderer.GetSafeString(string.Format("[{0}] {1}: {2}",
-                    message.DateTime.ToShortTimeString(), message.SenderName, message.Message), FontIndex),
-                    message.Color, true);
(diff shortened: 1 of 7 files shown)


devo1929:
image

Metadorius:
Would it be possible to add a "Copy text" menu entry and (when a link is detected) "Open link" and "Copy link"?

DXMainClient/DXGUI/Multiplayer/ChatListBox.cs
```diff
@@ -52,17 +52,22 @@ public void AddMessage(string sender, string message, Color color)
...
-                    message.Color, true);
+                    message.Message), FontIndex);
             else
-            {
-                AddItem(Renderer.GetSafeString(string.Format("[{0}] {1}: {2}",
-                    message.DateTime.ToShortTimeString(), message.SenderName, message.Message), FontIndex),
-                    message.Color, true);
-            }
+                listBoxItem.Text = Renderer.GetSafeString(string.Format("[{0}] {1}: {2}",
+                    message.DateTime.ToShortTimeString(), message.SenderName, message.Message), FontIndex);
```
Metadorius, line 68: Any multi-line code blocks should use braces for better readability and excluding the possibility of [goto fail error](https://www.synopsys.com/blogs/software-security/understanding-apple-goto-fail-vulnerability-2/).

Metadorius (changes requested):
Looks good aside from a braces issue.

devo1929:
> Would it be possible to add a "Copy text" menu entry and (when a link is detected) "Open link" and "Copy link"?

I like this idea, but maybe next PR? That'll include some parsing work.

DXMainClient/DXGUI/Multiplayer/ChatListBox.cs
devo1929: Fixed.

Metadorius:
Links are already parsed in code, try double-clicking any message with a link. It's just not highlighted due to how XNAUI works with text. Look for `http://` and `https://` in code, should be trivial to patch that into a menu entry.

devo1929:
> Links are already parsed in code, try double-clicking any message with a link. It's just not highlighted due to how XNAUI works with text. Look for `http://` and `https://` in code, should be trivial to patch that into a menu entry.

I found the code. I started to implement it and it works, but right now, the menu is in the context of a "player/user". It won't show if there's no user associated with the message you're right-clicking on. System messages (like the IP ban one) won't have a context menu. It's not a difficult add, but it's not insignificant either.

Metadorius:
Then I guess feel free to leave that for another PR 🤷‍♂️

Metadorius (approved):

PR #268, opened 2021-11-30
Context menu enhancements

devo1929:
- Add open/copy link options to message context menu
- Add context menu to private message window lists

I renamed the PlayerContextMenu to be GlobalContextMenu, because it can have different/hybrid contexts. It can be used for a player or a message (user sent or system sent).

image

image


devo1929:
I renamed the PlayerContextMenu to be GlobalContextMenu, because it can have different/hybrid contexts. It can be used for a player or a message (user sent or system sent).

alexp8 (approved):

PR #271, opened 2021-12-08
Disable all private messages

devo1929:
image

image

OLD UI for reference
image

### DIFF
--- a/ClientCore/Settings/UserINISettings.cs
+++ b/ClientCore/Settings/UserINISettings.cs
@@ -104,6 +104,7 @@ namespace ClientCore
             AllowGameInvitesFromFriendsOnly = new BoolSetting(iniFile, MULTIPLAYER, "AllowGameInvitesFromFriendsOnly", false);
             NotifyOnUserListChange = new BoolSetting(iniFile, MULTIPLAYER, "NotifyOnUserListChange", true);
             DisablePrivateMessagePopups = new BoolSetting(iniFile, MULTIPLAYER, "DisablePrivateMessagePopups", false);
+            DisablePrivateMessagesFromNonFriends = new BoolSetting(iniFile, MULTIPLAYER, "DisablePrivateMessagesFromNonFriends", false);
             DisablePrivateMessages = new BoolSetting(iniFile, MULTIPLAYER, "DisablePrivateMessages", false);
             EnableMapSharing = new BoolSetting(iniFile, MULTIPLAYER, "EnableMapSharing", true);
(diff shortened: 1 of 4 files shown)


Metadorius:
Maybe it would be good to also have a function to disable messages from non-friends?

devo1929:
> Maybe it would be good to also have a function to disable messages from non-friends?

Easily doable. I would add it as another "sub" option under the Disable Private Messages. Does it need to be in this PR?

Metadorius:
As you wish, I'd group them together but I don't mind either way.

devo1929:
@Metadorius 
image

image

Metadorius:
hmm not sure if it is intuitive this way, on the other thought... maybe it would be better to do a drop-down called `Allow PMs from:` with options `Everyone`, `Friends` and `No one` and keep `Disable PM Popups` like it was before?

devo1929:
> hmm not sure if it is intuitive this way, on the other thought... maybe it would be better to do a drop-down called `Allow PMs from:` with options `Everyone`, `Friends` and `No one` and keep `Disable PM Popups` like it was before? because I had to read into the names of the checkboxes for a good minute before realizing how do they interact.

Killin me...

devo1929:
> hmm not sure if it is intuitive this way, on the other thought... maybe it would be better to do a drop-down called `Allow PMs from:` with options `Everyone`, `Friends` and `No one` and keep `Disable PM Popups` like it was before? because I had to read into the names of the checkboxes for a good minute before realizing how do they interact.

I'm not against this, but it would require refactor, not only to what I've done recently.

Metadorius:
Then perhaps a different wording on the checkbox would help? Something like "Allow friends to PM", or invert checkbox and call it "Block PMs from friends as well".

devo1929:
> Then perhaps a different wording on the checkbox would help? Something like "Allow friends to PM", or invert checkbox and call it "Block PMs from friends as well".

I do like your drop down idea. The refactoring required would be to create a new INI property that would hold the "state" of this drop down and then being backwards compatible with the existing Disable PM Popups setting. IE, it would need converted. Alternatively, we leave it as 3 separate boolean properties as it currently is in the INI and have the drop down change event responsible for changing all 3 of those values.
Thoughts?

Metadorius:
I don't think backwards compatibility is needed though? Popups will stay the same old checkbox, it's just that the new PM disable settings would be done via a dropdown. That is if we do it soon enough, before the version with checkboxes goes into mods and YR client. But if we aren't gonna do that now I suggest just to apply rewording to remove confusion element.

devo1929:
Ah, you're correct. For some reason, I was thinking the "popup" one would need to be in the drop down.

PR #281, opened 2021-12-26
Team start mappings

devo1929:
This provides a number of options:

- Force random colors - disables the color drop down for all players and sets them to the random option
- Force random sides - disables the side drop down for all players and sets them to the random option
- Force random teams - disables the team drop down for all players and sets them to the random option
- Force random starts - disables the start drop down for all players and sets them to the random option
  - Also disables the ability to click on a spawn location on the map preview
- Use Team/Start Mappings
  - Enables the drop downs beneath the option to select teams and starting locations
  - Automatically checks and disables Force Random Teams and Force Random Starts. These two options are required for this to work as expected.
  - The drop downs beneath this check box are in two column of mapping pairs. Each pair has a drop down for Team letter and Starting position, in that order.
  - The team drop down also has a "-" option. This can be used to "block" a spawn location from being randomly assigned.
  - There is a help (question mark) button that will open a dialog with helpful information on how to use the mappings.
  - When used, players will be auto assigned to random starting locations first (using existing code logic). Then, using the mappings, their teams are assigned. That information is then written to the spawn.ini for the game to pick up. This is all mostly done using existing code.
  - MPMaps.ini can set these properties for a given map as a list of "presets", so that all of the drop downs are auto populated when a map is selected. There is a drop down listing the presets for a given map (see screen shots for example). The first preset listed in the INI for a given map will be auto selected when a map is selected.
  - Specific custom map files can also have these properties specified to do the same as above.
  - When a map is selected that doesn't contain these mappings, the drop downs are cleared.
- Compatible with cncnet, skirmish, and lan games
- Overall cleanup/consolidation of code

Screenshots.
This image shows two pairs of columns, only one pair column of which is enabled. This is because the map currently selected only has a max player count of 4. Thus, only 4 mappings are required.
image

image

image

image

image


Metadorius:
One issue right away. I don't think "Force random starts" should be always on for team/start mappings. Imagine the case when there are 2 high-tier players and a bunch of lower skill players. If they get in one team - it would be an unbalanced game, so we need to pre-set starts to those players so they don't end up in the same team.

Metadorius:
This might not be a popular case for YR but for mods it's very common when there are big skill differences and small active playerbases when everyone knows how each other plays.

DXMainClient/DXGUI/Multiplayer/PlayerExtraOptionsPanel.cs
```diff
@@ -0,0 +1,315 @@
...
+            chkBoxForceRandomColors.InputEnabled = enable;
+            chkBoxForceRandomStarts.InputEnabled = enable;
+            chkBoxForceRandomTeams.InputEnabled = enable;
+            chkBoxUseTeamStartMappings.InputEnabled = enable;
+
+            teamStartMappingsPanel.EnableControls(enable && chkBoxUseTeamStartMappings.Checked);
+        }
+
+        public PlayerExtraOptions GetPlayerExtraOptions() =>
+            new PlayerExtraOptions()
```
Metadorius, line 288: IMHO it's better to keep `=>` at the second line because
a) in case the function is gonna be expanded the line with signature won't need to be changed from git standpoint
b) it's easier to see such stuff at the beginning of a line
c) it's logically related to the body of the function more

DXMainClient/Domain/Multiplayer/Map.cs
```diff
@@ -367,6 +375,35 @@ public bool SetInfoFromMpMapsINI(IniFile iniFile)
             }
         }
 
+        private void GetTeamStartMappingPresets(IniSection section)
+        {
+            TeamStartMappingPresets = new List<TeamStartMappingPreset>();
+            for(var i = 0;;i++)
```
Metadorius, line 381:
```suggestion
            for (var i = 0; ; i++)
```

DXMainClient/Domain/Multiplayer/TeamStartMapping.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        /// <summary>
+        /// We write these out in the form A:1 where A is the team and 1 is the starting location on the map
+        /// as it would appear in the map preview.
+        /// </summary>
+        /// <returns></returns>
+        public override string ToString()
+        {
+            return Team;
+        }
```
Metadorius, line 44: Outdated comment? Also not sure if it should be done via `ToString` considering it now doesn't give exhaustive information about the team start mapping in a programmer sense.

DXMainClient/Domain/Multiplayer/TeamStartMapping.cs
```diff
@@ -0,0 +1,88 @@
...
+        /// <param name="index"></param>
+        /// <returns></returns>
+        private static TeamStartMapping FromString(string str, int index)
+        {
+            return new TeamStartMapping()
+            {
+                Team = str,
+                Start = index + 1
+            };
+        }
```
Metadorius, line 60: Not sure if this is worth leaving in as a separate method as well, don't see it being useful outside of parsing function.

DXMainClient/Online/Connection.cs
```diff
@@ -981,15 +981,39 @@ private void SendMessage(string message)
...
+            lock (messageQueueLocker)
+            {
+                var previousMessageIndex = MessageQueue.FindIndex(m => m.MessageType == qm.MessageType);
+                if (previousMessageIndex == -1)
+                    return false;
+                
+                MessageQueue[previousMessageIndex] = qm;
+                return true;
+
+            }
```
Metadorius, line 1001:
```suggestion
            }
```

DXMainClient/Online/Channel.cs
```diff
@@ -259,13 +259,23 @@ public void SendChatMessage(string message, IRCColor color)
                 "PRIVMSG " + ChannelName + " :" + colorString + message);
         }
 
-        public void SendCTCPMessage(string message, QueuedMessageType qmType, int priority)
+        /// <summary>
+        ///
```
Metadorius, line 263: Probably forgot to write? This may be helpful: https://en.wikipedia.org/wiki/Client-to-client_protocol

Metadorius (changes requested):
The code looks fine, one note regarding code style - I see a lot of one-line functions not being defined using expression bodies, commented on some of those. Please check other one-liners too.

Another note - when selecting a team in current client the `-` means "no team", and when assigning a team to a start position it should mean the same, but currently instead of this it means "don't assign a player to this spot" which is kind of counterintuitive. To fix this I propose one of the following (a is the easiest, c is IMO the best one):
a) change the blocked location symbol to `x`;
b) add a checkbox near each of the dropdowns which just disables the checkbox and position from being assigned;
c) same as a) or b) but also add `-` as "assign a player to this location but don't assign any team".

DXMainClient/Domain/Multiplayer/Map.cs
Rampastring: Additionally, as defined in the style guide:

> Never use `var` with primitive types.

DXMainClient/Domain/Multiplayer/TeamStartMapping.cs
devo1929: The purpose of the `FromString` and `ToString` methods was to containing the serializing in/out of this class IN this class. This serializing occurs when reading in Map data and for sending/receiving these mappings via a IRC message.

DXMainClient/Online/Channel.cs
devo1929: Not necessarily. There was no summary on this function to begin with. I just know how the replace functionality should work for it. So, I added the summary to explain the param. I didn't want to write a summary without a full knowledge of the underlying functionality of the messaging.

devo1929:
@Metadorius 
I updated this PR with all styling changes as well as removing the auto forcing of random starts upon checking the box for team/start mappings. This should allow for players to manually set the location for "pro" players in your example.

I looked into the thing I mentioned _might_ be an issue: the host manually blocks more locations than they should be allowed to. IE, they should only be able to block off as many locations as there are open spots in the player list.

While I think this is doable, it might be one of those things that we need to weigh: is it worth the effort or is it something just leave to the host to make sure they set it up correctly?

What will happen if they block more locations than there are empty spots is this: once the client has run out of "available" (non-blocked) spots to assign, the last X number of players in the list will be randomly assigned to any of those blocked spots.

DXMainClient/Domain/Multiplayer/TeamStartMapping.cs
Metadorius: I get that, but do we need to split the stuff so much? I personally don't see a point, considering this is for internal use and it's used only once.

DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
```diff
@@ -289,7 +287,7 @@ public override void Initialize()
             mapContextMenu.Width = 100;
             mapContextMenu.AddItem("Delete Map", DeleteMapConfirmation, () => Map != null && !Map.Official);
             AddChild(mapContextMenu);
-
+
```
Metadorius, line 290:
```suggestion

```

DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
```diff
@@ -508,6 +514,43 @@ protected void DdGameMode_SelectedIndexChanged(object sender, EventArgs e)
                 ChangeMap(GameMode, Map);
         }
 
+        protected void BtnPlayerExtraOptions_LeftClick(object sender, EventArgs e)
+        {
+            PlayerExtraOptionsPanel.Enable();
+        }
```
Metadorius, line 520: Expression body here?

DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
```diff
@@ -508,6 +514,43 @@ protected void DdGameMode_SelectedIndexChanged(object sender, EventArgs e)
...
+            }
+
+            SetPlayerExtraOptions(playerExtraOptions);
+            UpdateMapPreviewBoxEnabledStatus();
+        }
+
+        private void AddPlayerExtraOptionForcedNotice(bool disabled, string type)
+        {
+            AddNotice($"The game host has {(disabled ? "disabled" : "enabled")} {type} selection");
+        }
```
Metadorius, line 552: And here.

DXMainClient/DXGUI/Multiplayer/PlayerExtraOptionsPanel.cs
```diff
@@ -0,0 +1,314 @@
...
+
+            ignoreMappingChanges = true;
+            teamStartMappingsPanel.SetTeamStartMappings(teamStartMappings);
+            ignoreMappingChanges = false;
+        }
+
+        private void RefreshPresetDropdown()
+        {
+            ddTeamStartMappingPreset.AllowDropDown = _isHost && chkBoxUseTeamStartMappings.Checked;
+        }
```
Metadorius, line 158: And here.

DXMainClient/DXGUI/Multiplayer/PlayerExtraOptionsPanel.cs
```diff
@@ -0,0 +1,314 @@
...
+            const int mappingPanelDefaultY = 0;
+            if (index > 0 && index % maxColumnCount == 0) // need to start a new column
+                return new Rectangle(((index / maxColumnCount) * (teamMappingPanelWidth + mappingPanelDefaultX)) + 3, mappingPanelDefaultY, teamMappingPanelWidth, teamMappingPanelHeight);
+
+            var lastControl = index > 0 ? teamStartMappingsPanel.GetTeamStartMappingPanels()[index - 1] : null;
+            return new Rectangle(lastControl?.X ?? mappingPanelDefaultX, lastControl?.Bottom + 4 ?? mappingPanelDefaultY, teamMappingPanelWidth, teamMappingPanelHeight);
+        }
+
+        private void ClearTeamStartMappingSelections() =>
+            teamStartMappingsPanel.GetTeamStartMappingPanels().ForEach(panel => panel.ClearSelections());
```
Metadorius, line 104: IMHO the `=>` should be at second line, see comment below for explanation.

DXMainClient/DXGUI/Multiplayer/TeamStartMappingPanel.cs
```diff
@@ -0,0 +1,79 @@
...
+            TeamStartMapping.TEAMS.ForEach(ddTeams.AddItem);
+            AddChild(ddTeams);
+
+            ddTeams.SelectedIndexChanged += DD_SelectedItemChanged;
+        }
+
+        private void DD_SelectedItemChanged(object sender, EventArgs e)
+        {
+            OptionsChanged?.Invoke(sender, e);
+        }
```
Metadorius, line 50: And here.

DXMainClient/DXGUI/Multiplayer/TeamStartMappingPanel.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        public void EnableControls(bool enable)
+        {
+            ddTeams.AllowDropDown = enable;
+        }
+
+        public void ClearSelections()
+        {
+            ddTeams.SelectedIndex = _defaultTeamIndex;
+        }
```
Metadorius, line 68: And here.

DXMainClient/Domain/Multiplayer/PlayerHouseInfo.cs
```diff
@@ -107,52 +107,55 @@ public int InternalSideIndex
...
         {
+            overrideGameRandomLocations |= ClientConfiguration.Instance.UseClientRandomStartLocations;
             if (IsSpectator)
             {
                 StartingWaypoint = 90;
-                return false;
+                return;
             }
-
+
```
Metadorius, line 130:
```suggestion

```

DXMainClient/Domain/Multiplayer/TeamStartMapping.cs
```diff
@@ -0,0 +1,82 @@
...
+        [JsonIgnore]
+        public int TeamId => ProgramConstants.TEAMS.IndexOf(Team);
+
+        [JsonIgnore]
+        public int StartingWaypoint => Start - 1;
+
+        [JsonIgnore]
+        public bool IsBlock => Team == NO_TEAM;
+
+        public override string ToString() => Team;
```
Metadorius, line 38: The reason I don't think using ToString is good here is because it's `public` and is usually a representation of this object, which is pretty much conventional, so you would want an exhaustive (in a way; it may not be the full information but at least start and team should be enough) information about the object's contents. While team is surely a part of the object representation the full info would be team and start, so this override doesn't really describe the object exhaustively as it should. And because the method is so small I suggest just to remove it and just use a lambda.

DXMainClient/Domain/Multiplayer/TeamStartMapping.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        /// <summary>
+        /// Write these out in a delimited list.
+        /// </summary>
+        /// <param name="teamStartMappings"></param>
+        /// <returns></returns>
+        public static string ToListString(List<TeamStartMapping> teamStartMappings)
+        {
+            return string.Join(LIST_SEPARATOR.ToString(), teamStartMappings.Select(mapping => mapping.ToString()));
+        }
```
Metadorius, line 64: EB here as well.

Metadorius (changes requested):
Couple more nitpicks.

DXMainClient/DXGUI/Multiplayer/PlayerExtraOptionsPanel.cs
```diff
@@ -0,0 +1,315 @@
...
+
+            _map = map;
+
+            RefreshTeamStartMappingPanels();
+        }
+
+        public List<TeamStartMapping> GetTeamStartMappings()
+        {
+            return chkBoxUseTeamStartMappings.Checked ? teamStartMappingsPanel.GetTeamStartMappings() : new List<TeamStartMapping>();
+        }
```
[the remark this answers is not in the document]
Metadorius: No input on this one?

DXMainClient/DXGUI/Multiplayer/PlayerExtraOptionsPanel.cs
Metadorius: No input on this one?

Metadorius (approved):
I assume the two more comments got hidden and weren't noticed, please check them, otherwise it's good to merge.

DXMainClient/DXGUI/Multiplayer/PlayerExtraOptionsPanel.cs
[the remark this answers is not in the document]
devo1929: Not sure why this was hidden.

Issue #45, opened 2017-02-20
Question in Chinese localization.


Sewens:
Greetings, I'm a MO gamer in china.I'm trying to translation this DTAclient into Chinese version.
I try to compile a .SpriteFont file with Chinese characters.
But there still remains some problems.

![qq 20170220092612](https://cloud.githubusercontent.com/assets/9146919/23109243/3ae7b636-f752-11e6-978a-2da003822a9c.png)

![qq 20170220091508](https://cloud.githubusercontent.com/assets/9146919/23109167/4e523c06-f751-11e6-87e0-cf716a44fca0.png)

I replace **SpriteFont1.xnb** in MO 3.3's Resources folder, and it's works a little.
From the pic above u can see, some text textures' font in UI have been changed,and other Chinese characters in list box, check box, etc. still have some problem. Some characters were replaced by question mark.

![qq 20170220092556](https://cloud.githubusercontent.com/assets/9146919/23109240/33e288c0-f752-11e6-8060-d1f5848f05fe.png)

Also, I try to chat in Chinese in multiplayer mode, still there are some characters remain in question mark.

![qq 20170220095813](https://cloud.githubusercontent.com/assets/9146919/23109379/bbea8fbe-f753-11e6-94db-56612f9fc834.png)

And here's my speculation:

1.  IT IS the default font this project use that cause the question mark issue. For example suppose we use font "Times New Romans" as client default font, when we type in Chinese character client could find only small part of these Font and others become question mark.
2.  Only small part of text textures use SpriteFont files, and it helps a little in online chat and Chinese localization.
3.  WOULD YOU KIND create an INI interface from which players can change the default font that client use.So that it will be easier for players around the world to trans this client into their language version.

Besides, someone has done translation successful by decompiling some files.
Here's the pic:
![81588218367adab4f566e4e682d4b31c8601e4a7](https://cloud.githubusercontent.com/assets/9146919/23109697/51904016-f756-11e6-9146-5089f46493a0.png)

![3a5769224f4a20a48e76e46399529822730ed0e8](https://cloud.githubusercontent.com/assets/9146919/23109701/5ad77716-f756-11e6-822f-bfcd7d9bd66e.png)

And link: [百度贴吧](http://tieba.baidu.com/p/4975190949)

Rampastring:
Question marks appear in place of every character that doesn't exist in the SpriteFont.

All text textures use the SpriteFont files.

[...]

PR #285, opened 2022-01-13
Add translation feature for literal strings

SadPencil:
- Load the translation from an .ini file.
- Just add the translation label name after the string.
- Backward compatible. By default the program does not requires an translation file to be loaded. Merging this PR brings minimum migration cost.
- Generating translation stub feature

Example:

**Before**
```csharp
return "Please enter a name.";
```

**After**
```csharp
using Localization;

return "Please enter a name.".L10N("UI:ClientCore:EnterAName");
```


SadPencil:
https://github.com/CnCNet/xna-cncnet-client/actions/runs/1701442197
Corresponding `Resources\Translation.ini` file is as below. (99% completed. Strings that only appear in condition compilings are missing.)
[Translation.zip](https://github.com/CnCNet/xna-cncnet-client/files/7874841/Translation.zip)

devo1929 (approved):

DXMainClient/DXGUI/Multiplayer/CnCNet/LoadOrSaveGameOptionPresetWindow.cs
```diff
@@ -144,8 +145,8 @@ public override void Initialize()
         public void Show(bool isLoad)
         {
             _isLoad = isLoad;
-            lblHeader.Text = $"{(_isLoad ? "Load" : "Save")} Preset";
-            btnLoadSave.Text = _isLoad ? "Load" : "Save";
+            lblHeader.Text = _isLoad ? "Load Preset".L10N("UI:Main:LoadPreset") : "SavePreset".L10N("UI:Main:SavePreset");
```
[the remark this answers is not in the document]
devo1929: eagle eye

Localization/TranslationTable.cs
```diff
@@ -0,0 +1,153 @@
...
+            IniSection general = ini.GetSection("General");
+            LanguageTag = general?.GetStringValue("LanguageTag", null);
+            LanguageName = general?.GetStringValue("LanguageName", null);
+            string CultureInfoName = general?.GetStringValue("CultureInfo", null);
+            IniSection translation = ini.GetSection("Translation");
+
+            if (general == null || translation == null || LanguageTag == null || LanguageName == null || CultureInfoName == null)
+            {
+                throw new Exception("Invalid translation table file.");
+            }
```
Metadorius, line 63:
```suggestion
            if (general == null || translation == null || LanguageTag == null
                || LanguageName == null || CultureInfoName == null)
            {
                throw new Exception("Invalid translation table file.");
            }
```

Localization/TranslationTable.cs
```diff
@@ -0,0 +1,153 @@
...
+                return defaultValue;
+            }
+        }
+
+        public static string EscapeIniValue(string raw)
+        {
+            if (raw.Contains(IniNewLinePattern))
+            {
+                throw new Exception($"Pattern {IniNewLinePattern} is forbidden as this pattern is used to represent the new line.");
+            }
```
Metadorius, line 141:
```suggestion
                throw new Exception($"Pattern {IniNewLinePattern} is forbidden as this pattern is used to represent the new line.");
```

Localization/TranslationTable.cs
```diff
@@ -0,0 +1,153 @@
...
+
+            string value = raw.Replace(Environment.NewLine, "\n");
+            value = value.Replace("\n", IniNewLinePattern);
+            return value;
+        }
+
+        public static string UnescapeIniValue(string escaped)
+        {
+            return escaped.Replace(IniNewLinePattern, "\n");
+        }
```
Metadorius, line 151:
```suggestion
            => escaped.Replace(IniNewLinePattern, "\n");
```

Localization/MissingTranslationEventArgs.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Threading.Tasks;
+
+namespace Localization
+{
+    public class MissingTranslationEventArgs : EventArgs
+    {
+        public string LanguageTag { get; set; }
+        public string Label { get; set; }
+        public string DefaultValue { get; set; }
+        public MissingTranslationEventArgs(string languageTag, string label, string defaultValue)
```
Metadorius, line 14:
```suggestion
        public string DefaultValue { get; set; }

        public MissingTranslationEventArgs(string languageTag, string label, string defaultValue)
```

Localization/ExtensionMethods.cs
```diff
@@ -0,0 +1,21 @@
...
+        /// <summary>
+        /// Mark this string to be translated, with the given label name.
+        /// </summary>
+        /// <param name="defaultValue">The default string value as a fallback.</param>
+        /// <param name="label">The unique label name.</param>
+        /// <returns>The translated string value.</returns>
+        public static string L10N(this String defaultValue, string label)
+        {
+            return LanguageInstance.TranslationTable.GetTableValue(label, defaultValue);
+        }
```
Metadorius, line 19:
```suggestion
            => LanguageInstance.TranslationTable.GetTableValue(label, defaultValue);
```

Localization/LanguageInstance.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Localization
+{
+    public static class LanguageInstance
```
Metadorius, line 10: I don't see any reason to separate this instance into another class. Instead I suggest to make `TranslationTable` class a singleton, you can do this easy with an `Instance` property and null-coalescing operators.

Metadorius (changes requested):
Some changes need to be made, otherwise the code is good.

DXMainClient/DXGUI/Multiplayer/CnCNet/LoadOrSaveGameOptionPresetWindow.cs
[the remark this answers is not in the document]
Metadorius: literally just randomly stumbled upon this xD

Localization/MissingTranslationEventArgs.cs
SadPencil: Format on coding style

Localization/LanguageInstance.cs
SadPencil: I don't agree on this as `TranslationTable` is designed to be corresponding to an .ini file, instead of the current language table of the running program. For instance, the program may load all the .ini file in something like `/Resources/Translation/*.ini` and make all the `TranslationTable.LanguageName` into a list to let user select the language, in the future.

Metadorius: There's no reason for this to be a separate class *at this stage of development*. It's not hard to change from singleton class to non-singleton when that actually gets implemented, which may never happen because of complexity of reloading everything.

If you don't want to make it a singleton the way I described, then it's okay, but we still have to get rid of this redundant class because it just creates unnecessary separation which makes the code harder to navigate.

DXMainClient/PreStartup.cs
```diff
@@ -73,6 +74,53 @@ public static void Initialize(StartupParams parameters)
...
+            try
+            {
+                if (ClientConfiguration.Instance.GenerateTranslationStub)
+                {
+                    // Not use using statement as variable "fs" will be destroyed after the lambda function be destroyed
+                    var fs = File.Open("Client/Translation.stub.ini", FileMode.Create);
+                    var sw = new StreamWriter(fs, new System.Text.UTF8Encoding(false)) { AutoFlush = true };
+                    Logger.Log("Generating translation stub feature is now enabled.");
+
+                    // Note, the TranslationTable.SaveIni() method is not used to avoid re-writing the file
```
Metadorius, line 98: Also I am not sure why would you meed to duplicate code for that, couldn't you adjust the class methods for that?

DXMainClient/PreStartup.cs
```diff
@@ -73,6 +74,53 @@ public static void Initialize(StartupParams parameters)
...
+                Logger.Log("Failed to load the translation file. " + ex.Message);
+            }
+
+            try
+            {
+                if (ClientConfiguration.Instance.GenerateTranslationStub)
+                {
+                    // Not use using statement as variable "fs" will be destroyed after the lambda function be destroyed
+                    var fs = File.Open("Client/Translation.stub.ini", FileMode.Create);
+                    var sw = new StreamWriter(fs, new System.Text.UTF8Encoding(false)) { AutoFlush = true };
```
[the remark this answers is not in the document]
SadPencil: This is because when a label is missing, the ini file needs to be re-written. Use an INI class or TranslationTable.SaveIni() causes the file to be totally re-written every time a new label is missing, instead of just append the value to the end at the text stream.

DXMainClient/PreStartup.cs
SadPencil: Calling `TranslationTable.SaveIni()` and append values to the end of the text stream requires the `[Translation]` section must be the last section in the ini file. Currently, the code does not guarantee that.

To solve the problem in this file the only way seems to be either re-writing the ini file everytime and sacrifice the performance, or generating the stub file only when program exits

SadPencil: Is creating a timer for re-generating the ini file periodically acceptable?

Localization/LanguageInstance.cs
SadPencil: Alright. Now I removed this class and moved the field into `TranslationTable.Instance`. But I did not choose to use `??=` operator to prevent possible (although I've not found) `L10N()` invokes before the Instance get assigned.

DXMainClient/PreStartup.cs
Metadorius: Is there any benefit to incrementally writing translation stubs instead of dumping them on unload? If yes then I guess we can keep this as is, if no - let's just dump the INI at exit.

SadPencil: The benefit is to preserve the stub file if the program crashes... Well it seems indeed unnecessary.
Now I re-write this part to use a timer.
To dump the INI at exit, where is the currect location to write something where the program exit? Maybe after calling `new Startup().Execute()`?

SadPencil: Well, I think maybe I can add a `ProgramExitEvent` to the program class and other components just subscribe to this event if they want to do something when exiting.

Metadorius: That seems like a nice addition, I like the event way.

SadPencil: `AppDomain.CurrentDomain.ProcessExit` is such an event. Now the stub file will be generated when program exits.

Metadorius (approved):

pzhlkj6612:
Does this solve part of #45 ?

SadPencil:
> Does this solve part of #45 ?

No. 
By the way, Fantasy ADVENTURE supports IME but their fork seems to be a private fork "thanks" to the MIT license.

PR #286, opened 2022-01-15
Explicitly set background texture of load/save config window

devo1929:


Metadorius:
Is this in-line with other dialog boxes?

devo1929:
> Is this in-line with other dialog boxes?

Not sure what you mean. This is a custom class that extends the XNAWindow. I just know that someone shared a screen shot of this window having no background at all. I'm not sure what would have caused that, but I know setting it explicitly resolves it.

Metadorius:
You can have any window have no backround on certain configurations. It's not a code problem, it's an INI config problem.

devo1929:
Well there's no explicit setting in the INI files for this window (or, at least I didn't create one...). If it were a misconfiguration for the generic xnawindow, wouldn't all of their windows have no background? Unless, they were explicitly set as well somewhere.

Rampastring:
All `XNAWindow`s that have their `Name` property set will parse either a INI file that matches their `Name` property, or use settings from `GenericWindow.ini`. So, you should assign `Name = nameof(LoadOrSaveGameOptionPresetWindow)` before calling `base.Initialize`.

devo1929:
I figured it came from Generic window.ini, but the person reported it being transparent. I would assume that all windows would have this problem then, unless someone manually updated their INI file to configure this specific window by name. I'm ok with the fact that this is not a code problem. I'm just trying to determine the cause of their issue.

Rampastring:
Most windows define a default background texture to use, like `BackgroundTexture = AssetLoader.LoadTexture("gamelobbybg.png")` or create a solid color texture, like you do. So having a default would be good. It should be assigned before calling `base.Initialize()` however, so the INI system can still override the default texture.

devo1929:
> Most windows define a default background texture to use, like `BackgroundTexture = AssetLoader.LoadTexture("gamelobbybg.png")` or create a solid color texture, like you do. So having a default would be good. It should be assigned before calling `base.Initialize()` however, so the INI system can still override the default texture.

Understood

devo1929:
Updated this PR per @Rampastring 's suggestion to swap the `base.Initialize()` call

PR #288, opened 2022-01-16
Fix issue with deleting custom map

devo1929:
Maps were being deleted properly, but the UI wasn't being updated to show it. If the user restarted their client, the map would no longer be there.

### DIFF
--- a/DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
+++ b/DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
@@ -43,19 +43,19 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
         /// </summary>
         /// <param name="windowManager"></param>
         /// <param name="iniName">The name of the lobby in GameOptions.ini.</param>
-        /// <param name="gameModeMaps"></param>
+        /// <param name="mapLoader"></param>
         /// <param name="isMultiplayer"></param>
         /// <param name="discordHandler"></param>
         public GameLobbyBase(
             WindowManager windowManager, 
             string iniName,
-            GameModeMapCollection gameModeMaps, 
+            MapLoader mapLoader, 
             bool isMultiplayer, 
             DiscordHandler discordHandler
         ) : base(windowManager)
         {
             _iniSectionName = iniName;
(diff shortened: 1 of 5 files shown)


alexp8 (approved):

DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
```diff
@@ -687,16 +689,16 @@ private void DeleteSelectedMap(XNAMessageBox messageBox)
...
-                foreach (GameMode gameMode in GameModeMaps.GameModes)
-                {
-                    gameMode.Maps.Remove(Map);
-                }
+                MapLoader.DeleteCustomMap(GameModeMap);
 
                 tbMapSearch.Text = string.Empty;
                 if (GameMode.Maps.Count == 0)
+                    // this will trigger another GameMode to be selected
                     GameModeMap = GameModeMaps.Find(gm => gm.GameMode.Maps.Count > 0);
```
Rampastring, line 697: From the style guide:

> Braceless code block bodies should be made only when both code block head and body are single line, statements split into multiple lines and nested braceless blocks are not allowed within braceless blocks

While it's technically still a single line, the comment here makes it so it's not visually a single line. So add braces around the code block here.

DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
```diff
@@ -687,16 +689,16 @@ private void DeleteSelectedMap(XNAMessageBox messageBox)
...
+                else
+                    // this will trigger another Map to be selected
+                    lbGameModeMapList.SelectedIndex = lbGameModeMapList.SelectedIndex == 0 ?
+                        1 : lbGameModeMapList.SelectedIndex - 1;
```
Rampastring, line 701: Same note as above.

Rampastring (changes requested):

DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
devo1929: Will do.

PR #300, opened 2022-01-27
Show General Chat (#cncnet) banned message only in General Chat lobby

devo1929:
I know we talked about putting continued dev on hold until we get the translation and TS PRs merged, but I think enough people have been chiming in about this that it might warranted to get this included real quick.

### DIFF
--- a/DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetLobby.cs
+++ b/DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetLobby.cs
@@ -92,7 +92,7 @@ namespace DTAClient.DXGUI.Multiplayer.CnCNet
         private GameCollection gameCollection;
 
         private Color cAdminNameColor;
-        
+
         private Texture2D unknownGameIcon;
         private Texture2D adminGameIcon;
 
@@ -142,7 +142,7 @@ namespace DTAClient.DXGUI.Multiplayer.CnCNet
         {
             isJoiningGame = false;
         }
-        
+
         public override void Initialize()
         {
             invitationIndex = new InvitationIndex();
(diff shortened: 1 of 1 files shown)


DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetLobby.cs
```diff
@@ -574,6 +574,15 @@ private void PostUIInit()
         /// </summary>
         private void ConnectionManager_BannedFromChannel(object sender, ChannelEventArgs e)
         {
+            var chatChannel = connectionManager.FindChannel(e.ChannelName);
+            if (chatChannel != null)
+            {
+                if(chatChannel.ChannelName != ProgramConstants.GeneralCncnetChatChannelName)
```
devo1929, line 580: The purpose of the overall IF statement is to show the user that it's a specific chat channel that they're banned from, for one, not cncnet as a whole, which is what everyone is thinking when they see `#cncnet` in the error message.

This specific IF statement for the general chat can be removed in the future, if/when the ban is removed from the general chat in IRC.

DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetLobby.cs
```diff
@@ -578,15 +578,18 @@ private void ConnectionManager_BannedFromChannel(object sender, ChannelEventArgs
...
-            else
-            {
-                connectionManager.MainChannel.AddMessage(new ChatMessage(
-                    Color.White, "Cannot join game " + game.RoomName + ", you've been banned by the game host!"));
+                var chatChannel = connectionManager.FindChannel(e.ChannelName);
+                if (chatChannel == null)
+                    return;
+
+                if(chatChannel.ChannelName != ProgramConstants.GeneralCncnetChatChannelName)
+                    connectionManager.MainChannel.AddMessage(new ChatMessage(Color.White, $"Cannot join chat channel {chatChannel.UIName}, you're banned!"));
```
devo1929, line 586: This line might be temporary, if the ban is removed from the IRC general chat.

Otherwise, this IF block contains "chat channel" specific language to use the chat channel's UI name, rather than the #cncnet channel name, which is I believe what's confusing people. They think they're banned from cncnet as a whole, not just a chat channel.

PR #275, opened 2021-12-13
Tiberian Sun Client v6 Changes

Rampastring:
This draft PR ports over changes made to the client for the Tiberian Sun Client v6 release: https://www.moddb.com/mods/tiberian-sun-client/downloads/tiberian-sun-client-600-with-videos

Dawn of the Tiberium Age's fork of the client also contains these changes: https://github.com/Rampastring/xna-cncnet-client

The most important change is the `INItializableWindow` system. This is a new window class that allows complete layout customization through the INI file (including defining all controls), with syntax that allows defining control positions as relative to existing controls. UI no longer needs to be defined code-side, instead the code can simply fetch the controls defined on the INI files based on the controls' names and types.

Game lobbies have been remade as `INItializableWindow`s in this PR. All other windows still rely on the old, simpler window class as recreating them all would take a lot of (INI coding) work.

The obvious downside to this PR is that it breaks backwards-compatibility, requiring all projects to remake their game lobbies with the new `INItializableWindow` syntax.

For reference on how the `INItializableWindow` system works, I've attached the `SkirmishLobby.ini` (renamed to `.txt` as GitHub doesn't support INI files :') ) from v6 of the TS Client.

[SkirmishLobby.txt](https://github.com/CnCNet/xna-cncnet-client/files/7707434/SkirmishLobby.txt)

Oh, and **this PR also changes the client license to GNU GPL from MIT.**

PR #307, opened 2022-02-14
Make optimal and preferred resolutions configurable

SadPencil:
Modders may want a bigger screen size but magic values like `1280x800` are hard-coded.
This PR makes these hard-coded resolutions configurable.
It also fixes an incorrect implementation of `GetHashCode()`.


Rampastring:
#275 already does this.

PR #310, opened 2022-02-25
Fix a possible LAN game crash on an unexpected player name

devo1929:
There was a reported issue in the #support channel of Discord where someone's client was crashing while hosting a lan game as soon as another player joined. It happens when the "file hash" message is received from the joining user. When this happens, it attempts to find that player in their host's Player list. It attempts to set the `verified` property to `true` on that player. However, the client was not able to find the player in the list for some reason and thus a null pointer exception was being thrown.

I provided a temporary new exe to that person and they verified that they were able to play the game with fellow lan members without issue. I don't know if this change will cause any issues though...


SadPencil (approved):

SadPencil:
I think this PR should be merged.

- `HandleFileHashCommand` in CnCNetGameLoadingLobby has this null check.
- Even if we do not reproduce the issue, an adversary could indeed crash the client by faking this message with an unknown player name.

SadPencil:
btw, in CnCNetGameLoadingLobby, `pInfo.Verified = true;` is assigned only if the hash matches. I think maybe we need to also enforce this rule in LAN?

PR #311, opened 2022-03-04
Allow option to sort map list

devo1929:
This removes the default sorting of maps alphabetically and instead adds a sort button to the left of the game mode drop down list.

The list is not sorted by default. The setting for the sort selection is stored in user ini settings.

image

### DIFF
--- a/DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
+++ b/DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
@@ -12,6 +12,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using ClientCore.Enums;
 using DTAClient.DXGUI.Multiplayer.CnCNet;
 using DTAClient.Online;
 using DTAClient.Online.EventArguments;
@@ -127,6 +128,8 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
         protected XNALabel lblGameModeSelect;
         protected XNAContextMenu mapContextMenu;
         private XNAContextMenuItem toggleFavoriteItem;
+
+        protected XNAClientStateButton<SortDirection> btnMapSortAlphabetically;
 
         protected XNASuggestionTextBox tbMapSearch;
 
@@ -251,6 +254,8 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
                 ddGameModeMapFilter.AddItem(CreateGameFilterItem(gm.UIName, new GameModeMapFilter(GetGameModeMaps(gm))));
 
             lblGameModeSelect = FindChild<XNALabel>(nameof(lblGameModeSelect));
+            
+            InitBtnMapSort();
 
             tbMapSearch = FindChild<XNASuggestionTextBox>(nameof(tbMapSearch));
             tbMapSearch.InputReceived += TbMapSearch_InputReceived;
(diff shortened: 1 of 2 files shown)


DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
```diff
@@ -315,6 +318,21 @@ public override void Initialize()
...
+                { SortDirection.Asc, AssetLoader.LoadTexture("sortAlphaAsc.png") },
+                { SortDirection.Desc, AssetLoader.LoadTexture("sortAlphaDesc.png") },
+            });
+            btnGameModeMapSortAlpha.Name = nameof(btnGameModeMapSortAlpha);
+            btnGameModeMapSortAlpha.ClientRectangle = new Rectangle(
+                ddGameModeMapFilter.X + -ddGameModeMapFilter.Height - 4, ddGameModeMapFilter.Y,
+                ddGameModeMapFilter.Height, ddGameModeMapFilter.Height
+            );
+            btnGameModeMapSortAlpha.LeftClick += BtnGameModeMapSortAlpha_LeftClick;
+            btnGameModeMapSortAlpha.SetToolTipText("Sort Game Modes Alphabetically".L10N("UI:Main:GameModeSortAlphabet"));
```
Metadorius, line 333: I don't think you sort game modes here?

DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
```diff
@@ -393,6 +412,21 @@ public override void Initialize()
...
+        private void BtnGameModeMapSortAlpha_LeftClick(object sender, EventArgs e)
+        {
+            UserINISettings.Instance.GameModeMapSortState.Value = (int)btnGameModeMapSortAlpha.GetState();
+
+            RefreshGameSortAlphaBtn();
+            UserINISettings.Instance.SaveSettings();
+            ListMaps();
+        }
+
+        private void RefreshGameSortAlphaBtn()
```
Metadorius, line 424: Name is misleadingly shortened, maybe `RefreshMapSortButton` would be better? Also we don't usually use control name shortenings in method names I think.

Metadorius (changes requested):

DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
devo1929: ah, yea

DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
devo1929: Updated

devo1929:
@Metadorius  Updated code to be more map general in naming conventions. The action is actually sorting the specific class of GameModeMap. So, my naming convention followed suit. Although, at a higher level, you're actually sorting the map list. So, I renamed everything per your suggestion. Either way doesn't matter to me. Whatever is more clear.

Metadorius (approved):

Rampastring:
I suggest using the full word "alphabetically" instead of shortening it, maybe `btnSortMapsAlphabetically`. "Alpha" often refers to opacity in alpha blending so the current name seems confusing at first glance.

devo1929:
> I suggest using the full word "alphabetically" instead of shortening it, maybe `btnSortMapsAlphabetically`. "Alpha" often refers to opacity in alpha blending (which is heavily utilized on the graphical side of the client) so the current name seems confusing at first glance.

Updated

Metadorius:
Let's wait for TSCv6 to be merged then merge this.

Metadorius:
@Rampastring @devo1929 could you update this branch? We'll merge it afterwards.

PR #315, opened 2022-03-14
Post v6 updates

devo1929:
- Improve ini config file locating
  - This slightly improves the backwards compatibility of the new INItializableWindow implementation.
  - With the initial INItializableWindow updates, there seemed to be no inheritance of which ini config file to load for a given class implementing INItializableWindow. If it has an override ini name, it would attempt to use it, but not fall back to the name of the control itself, which could be the parent class' name.
- Make BasedOn a list for multi inheritance
- Fix missing toggle fav map event
- Add starting documentation for INI system
- Fix bug in XNAClientButton

### DIFF
--- a/ClientCore/CCIniFile.cs
+++ b/ClientCore/CCIniFile.cs
@@ -36,23 +36,33 @@ namespace ClientCore
 
         protected override void ApplyBaseIni()
         {
-            string basedOn = GetStringValue("INISystem", "BasedOn", string.Empty);
-            if (!string.IsNullOrEmpty(basedOn))
-            {
-                string path;
-                if (basedOn.Contains("$THEME_DIR$"))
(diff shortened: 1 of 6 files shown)


Rampastring (commented):
I haven't tested it, but the code changes look good.

I'm on the fence about multiple inheritance being a good idea as I'm expecting modders to be able to create confusing situations for themselves with it. But I guess we should rather provide them the power than prevent them from shooting themselves in the foot.

devo1929:
> I haven't tested it, but the code changes look good.
> 
> I'm on the fence about multiple inheritance being a good idea as I'm expecting modders to be able to create confusing situations for themselves with it. But I guess we should rather provide them the power than prevent them from shooting themselves in the foot.

In the end, I ended up not using it. So, if you want me to revert it, I'm ok with that. It won't hurt to leave it in though as it's entirely optional to use.

Metadorius (approved):
Code style and the doc draft look OK. Don't think I am qualified to review it from the architectural standpoint.

PR #316, opened 2022-03-15
Fix broken PlayerExtraOptionsPanel, adjust its layout to fit TSC better

Rampastring:
The massive merge of Tiberian Sun Client v6 changes left the PlayerExtraOptionsPanel broken. My bad.

This should fix it and also adjusts the layout of the panel to fit the layout of the Tiberian Sun Client. As far as I can see, this adjustment shouldn't cause issues for YR.


Metadorius (changes requested):
Either the fix is missing something or you didn't explain what else needs to be done to fix it, because nothing changed for me, still gives NullReferenceException at extra options panel.

Metadorius (approved):

PR #318, opened 2022-03-26
Update MultiplayerGameLobby controls visibility logic

frg2089:
This will provide more flexible control visibility support.
Applying different styles depending on the suffix will no longer be limited to "lbChatMessages" and "tbChatInput" controls.

Now we can write INI like this.
```ini
[MultiplayerGameLobby]
$CC-ChatMessages=ChatMessages:XNAPanel

[ChatMessages]
$X=1
$Y=1
$CC-lbChatMessages=lbChatMessages:ChatListBox

[ChatMessages_Host]
$BaseSection=ChatMessages
Size=100,200
BackgroundTexture=ChatMessagesBorder_Host.png

[ChatMessages_Player]
$BaseSection=ChatMessages
Size=100,100
BackgroundTexture=ChatMessagesBorder_Player.png
```
On `INItializableWindow` Initialized.
`ParseExtraControls()` method will create an `XNAPanel` named ChatMessages.

Then the `Refresh()` method will be called when creating a game or joining a game.
The ChatMessages control will be renamed to ChatMessages_Host or ChatMessages_Player via the `IsHost` field. and reload its properties.

### DIFF
--- a/DXMainClient/DXGUI/Multiplayer/GameLobby/MultiplayerGameLobby.cs
+++ b/DXMainClient/DXGUI/Multiplayer/GameLobby/MultiplayerGameLobby.cs
@@ -35,9 +35,9 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
             chatBoxCommands = new List<ChatBoxCommand>
             {
                 new ChatBoxCommand("HIDEMAPS", "Hide map list (game host only)".L10N("Client:Main:ChatboxCommandHideMapsHelp"), true,
-                    s => HideMapList()),
+                    s => ChangeControls(false)),
                 new ChatBoxCommand("SHOWMAPS", "Show map list (game host only)".L10N("Client:Main:ChatboxCommandShowMapsHelp"), true,
-                    s => ShowMapList()),
+                    s => ChangeControls(true)),
                 new ChatBoxCommand("FRAMESENDRATE", "Change order lag / FrameSendRate (default 7) (game host only)".L10N("Client:Main:ChatboxCommandFrameSendRateHelp"), true,
                     s => SetFrameSendRate(s)),
                 new ChatBoxCommand("MAXAHEAD", "Change MaxAhead (default 0) (game host only)".L10N("Client:Main:ChatboxCommandMaxAheadHelp"), true,
@@ -613,9 +613,9 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
 
             btnLaunchGame.Text = IsHost ? BTN_LAUNCH_GAME : BTN_LAUNCH_READY;
 
+            ChangeControls(IsHost);
             if (IsHost)
             {
-                ShowMapList();
                 btnSaveLoadGameOptions?.Enable();
 
                 btnLockGame.Text = "Lock Game".L10N("Client:Main:ButtonLockGame");
@@ -639,7 +639,6 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
             }
             else
             {
-                HideMapList();
                 btnSaveLoadGameOptions?.Disable();
 
                 btnLockGame.Enabled = false;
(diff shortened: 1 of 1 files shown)


devo1929 (approved):

DXMainClient/DXGUI/Multiplayer/GameLobby/MultiplayerGameLobby.cs
```diff
@@ -626,62 +626,58 @@ protected void Refresh(bool isHost)
...
+            if (isHost)
+            {
+                sourceSuffix = "_Player";
+                targetSuffix = "_Host";
+
+                ddGameModeMapFilter.Enable();
+                lblGameModeSelect.Enable();
+                lbGameModeMapList.Enable();
+                tbMapSearch.Enable();
+                btnPickRandomMap.Enable();
```
Metadorius, line 647: Pretty sure you forgot `btnMapSortAlphabetically`. Also wouldn't tunnel selection button need to be handled too at CnCNet game lobby? Maybe would be good to have this function as protected virtual?

DXMainClient/DXGUI/Multiplayer/GameLobby/MultiplayerGameLobby.cs
```diff
@@ -626,62 +626,58 @@ protected void Refresh(bool isHost)
...
+            }
+            else
+            {
+                targetSuffix = "_Player";
+                sourceSuffix = "_Host";
```
Metadorius, line 652:
```suggestion
                sourceSuffix = "_Host";
                targetSuffix = "_Player";
```

DXMainClient/DXGUI/Multiplayer/GameLobby/MultiplayerGameLobby.cs
```diff
@@ -626,62 +626,58 @@ protected void Refresh(bool isHost)
...
-            btnPickRandomMap.Disable();
-            btnMapSortAlphabetically.Disable();
-        }
+            string sourceSuffix; // The name of the target control may have a suffix. e.g. lbChatMessages_Player
+            string targetSuffix; // Change the name of the control to match the new suffix.
+            List<XNAControl> needUpdateControls = new List<XNAControl>(); // Controls that will be updated.
+            if (isHost)
+            {
+                sourceSuffix = "_Player";
+                targetSuffix = "_Host";
```
Metadorius, line 641: You should make those consts at class level.

DXMainClient/DXGUI/Multiplayer/GameLobby/MultiplayerGameLobby.cs
```diff
@@ -626,62 +626,58 @@ protected void Refresh(bool isHost)
...
-            ddGameModeMapFilter.Disable();
-            lblGameModeSelect.Disable();
-            lbGameModeMapList.Disable();
-            tbMapSearch.Disable();
-            btnPickRandomMap.Disable();
-            btnMapSortAlphabetically.Disable();
-        }
+            string sourceSuffix; // The name of the target control may have a suffix. e.g. lbChatMessages_Player
+            string targetSuffix; // Change the name of the control to match the new suffix.
+            List<XNAControl> needUpdateControls = new List<XNAControl>(); // Controls that will be updated.
```
Metadorius, line 637:
```suggestion
            List<XNAControl> controlsToUpdate = new();
```
Change the occurences accordingly.

DXMainClient/DXGUI/Multiplayer/GameLobby/MultiplayerGameLobby.cs
```diff
@@ -626,62 +626,58 @@ protected void Refresh(bool isHost)
             }
         }
 
-        private void HideMapList()
+        /// <summary>
+        /// Change MapList and other controls visibility.
+        /// </summary>
+        /// <param name="isHost">Visibility</param>
+        private void ChangeMapListAndOtherControlsVisibility(bool isHost)
```
Metadorius, line 633: Not fond of the name, needs something more concise.

DXMainClient/DXGUI/Multiplayer/GameLobby/MultiplayerGameLobby.cs
```diff
@@ -35,9 +35,9 @@ public abstract class MultiplayerGameLobby : GameLobbyBase, ISwitchable
             chatBoxCommands = new List<ChatBoxCommand>
             {
                 new ChatBoxCommand("HIDEMAPS", "Hide map list (game host only)".L10N("UI:Main:ChatboxCommandHideMapsHelp"), true,
-                    s => HideMapList()),
+                    s => ChangeMapListAndOtherControlsVisibility(false)),
                 new ChatBoxCommand("SHOWMAPS", "Show map list (game host only)".L10N("UI:Main:ChatboxCommandShowMapsHelp"), true,
-                    s => ShowMapList()),
+                    s => ChangeMapListAndOtherControlsVisibility(true)),
```
Metadorius, line 40: No, you're mixing two things. The prefixes you have and the bool you use as a parameter suggest differentiation between host and player, yet you're mixing this with map list hidden or shown.

Metadorius (changes requested):
Couple of nitpicks. I am not quite sure this is the direction we want to take (with hardcoded suffixes and handling only the very special case of map list visibility), I would prefer a more generalized solution. But the execution of that idea is more or less fine.

@Rampastring what's your opinion? Basically the code is very specific to this case and introduces INI behavior based on section names, which is only tied to map box being shown or not.

Issue #301, opened 2022-01-27
Added/removed friends are not re-sorted in user list upon action


devo1929:
With the latest version, the user list is sorted by admins, then friends, then alphabetically. However, this sorting only occurs when the user list is refreshed. If a user is added/removed to/from friends list, they are not sorted based on this at that time.

User list needs to be refreshed after a friend is added/removed.

PR #322, opened 2022-04-10
#301 make player list box reusable and fix list update issues

devo1929:
This makes the PlayerListBox that is used in the CncnetLobby reusable so that it can then be reused in the Private Messaging Window.

It also resolves some issues in "updating" the lists for various actions including: user added, removed, friended, ignored, etc... When these actions occur, this list is updated accordingly where ever it is used and in the same way in all locations.

### DIFF
--- a/DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetLobby.cs
+++ b/DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetLobby.cs
@@ -196,14 +196,9 @@ namespace DTAClient.DXGUI.Multiplayer.CnCNet
             lbGameList.AllowMultiLineItems = false;
             lbGameList.ClientRectangleUpdated += GameList_ClientRectangleUpdated;
 
-            lbPlayerList = new PlayerListBox(WindowManager, gameCollection);
+            lbPlayerList = new PlayerListBox(WindowManager, cncnetUserData, connectionManager, gameCollection);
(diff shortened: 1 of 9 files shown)


Rampastring:
A word of warning related to performance. Refreshing of the player list box is a performance-critical operation for games with large player bases like YR and MO. Back when YR hit 5000 concurrent players in May 2020, the performance of the implementation turned out to be insufficient, which was causing the client to stall on low-performance machines and YR players were even getting kicked from the IRC network due to the client processing messages too slowly.

The current public implementation with `SortedUserCollection` and refreshing the list directly based on it was written back then and helped YR survive the surge in players. This suggested implementation _might_ (not necessarily saying it will, but it's necessary to keep an eye on it) be heavier with heavy use of `ToList` (creating new list objects), LINQ sorting and selection operations and, based on a quick look and assuming I haven't missed anything, re-sorting the users in the player list each time any user info is updated.

I support re-using code so sharing the player list box between the lobby and the PM window is a good idea to explore, and doing so obviously needs changes to the player list class as well. But I'm afraid this implementation will backfire if there's a new user spike in the future. For regular lobby channels, it could be performance-friendlier to instead edit the user comparison function given to `SortedUserCollection`.

devo1929:
@Rampastring 
Yea, I had the same concern, but I wasn't sure if it was completely warranted. You confirmed it, though. 
I want to do a little more of an overhaul then. I'll close this for now.

Issue #327, opened 2022-04-21
Make map search across all maps, not just for current game mode


devo1929:
When searching for maps, you have to know what game mode a map is in for the search to find it. The search should be across all available maps.

Because maps can exist in multiple game modes, search results should show the game mode of each map in the list, similar to how it is done when viewing Favorites.

PR #328, opened 2022-04-21
#327 Map search across all maps

devo1929:
image

### DIFF
--- a/DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
+++ b/DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
@@ -474,9 +474,9 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
                 string.Format("The game host has disabled {0}".L10N("UI:Main:HostDisableSection"), type) :
                 string.Format("The game host has enabled {0}".L10N("UI:Main:HostEnableSection"), type));
 
-        private List<GameModeMap> GetSortedGameModeMaps()
+        private List<GameModeMap> GetSortedGameModeMaps(bool getAll)
         {
-            var gameModeMaps = gameModeMapFilter.GetGameModeMaps();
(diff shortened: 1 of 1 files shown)


alexp8:
Why are you adding this functionality? Searching for maps within the current game mode is preferred imo

@devo1929

devo1929:
> Why are you adding this functionality? Searching for maps within the current game mode is preferred imo
> 
> @devo1929

It removes the limitation that you must know the game mode a map is in to find it. What's the reason for preferring it to stay within the game mode?

alexp8:
In the screenshot you showed there it is very confusing, a player searching for the official Mayflower map is now presented with 5 other options. Not great

alexp8:
If anything add a toggle to enable that. By default it should search within the current game mode

devo1929:
> In the screenshot you showed there it is very confusing, a player searching for the official Mayflower map is now presented with 5 other options. Not great

IMO, the names "Battle" and "Standard" are confusing. To me, they don't mean "Official" and "Custom" which is what they truly are and "Standard" seems more "Official" than "Battle" does. 

If I were in the "Stardard" game mode and searched for Mayflower. I would see a map with the exact same name as the official one. At least this way, it shows there's a difference.

alexp8:
If those names are confusing to you, your solution here actually exacerbates that confusion providing maps from both game modes.

alexp8:
My opinion is that this should not be merged without some sort of toggle on/off. I know for me this feature would annoy me lol.

devo1929:
> My opinion is that this should not be merged without some sort of toggle on/off. I know for me this feature would annoy me lol.

Fair enough. That's why I create the PRs, to create the conversation :)

Rampastring:
You could consider renaming `Standard` to `Custom Map` in YR by changing the game mode's UI name, or by using game mode aliases.

Metadorius:
I agree with @alexp8, this should be a toggleable functionality. Perhaps you could do it as a settings option.

11EJDE11:
Closing in favor of #862

PR #338, opened 2022-05-12
Settings overhaul

Starkku:
This PR overhauls the existing file setting checkboxes and dropdowns and introduces generic setting & dropdown controls. The existing checkboxes in `GameOptionsPanel` have been replaced by these `SettingCheckBox` controls, which handle their own user INI setting writing/reading instead of setting a hardcoded setting in `UserSettingsINI`. This system can be adapted to other `XNAOptionPanel` instances in future if needed.

Migration info and details:
- `CustomSettingFileCheckBox` and `CustomSettingFileDropDown` have been renamed to simply `FileSettingCheckBox` and `FileSettingDropDown`. This requires adjusting the control names in `OptionsWindow.ini`.
- `FileSettingCheckBox` has a fallback to legacy behaviour if the control has any files defined with `FileX`.
-  `SettingCheckBox` and `SettingDropDown` are new controls that work in similar way except that they only write settings instead of moving files.
  - `SettingCheckBox` has additional options: `EnabledSettingValue` & `DisabledSettingValue` that define custom string values to write to settings INI for each state if `WriteSettingValue` is set to true.
  - `SettingDropDown` has an additional option: `WriteItemValue` can be set to true to write dropdown item's value to INI instead of its index.
- All setting controls have following new options available:
  - `SettingSection` determines section in user settings INI where it will read & write its settings from / to. Defaults to `CustomSettings` if not set.
  - `SettingsKey` does the same thing but for the INI key. Defaults to `controlName_Checked` for check boxes and `controlName_SelectedIndex` for dropdowns if `WriteSettingValue` /  `WriteItemValue` is not set, otherwise to `controlName_Value`.
- Additionally the checkbox controls have `ParentCheckBoxName` and `ParentCheckBoxRequiredValue` (defaults to `true`) that can be used to tie the whether or not the checkbox can be checked to the checked state of an another checkbox.


Metadorius (approved):
Good, let's fix the minor nitpicks, add the description into a doc draft and merge it.

Metadorius:
Would be good to insert the migration info into now existing migration doc.

Starkku:
Fixed the nitpicks and added docs.

Metadorius (approved):

Issue #350, opened 2022-07-26
Allow downloading any map from CnCNet MapDB


Metadorius:
FF's client has a chat command in lobby to download an arbitrary map from CnCNet MapDB using it's hash, like so: `/map somehashgoeshere` (from Holland on Discord):

> yeah, players can use the map database to search for any map:
http://mapdb.cncnet.org/search.php?game=ts&age=0&search=training
in this website there is a hashtag, if u paste that in the ui when u host a game, like so: /map hashtag it will set that map for ya

PR #351, opened 2022-07-27
#350: Add a command to download a map via lobby chat

alexlambson:
- Add a command to download a map by hash: `/downloadmap HASH`
- Allow users to specify a filename after the hash, e.g. `/downloadmap HASH [2] My Battle Map`
- Fix a bug where renaming unzipped map files would cause a crash if the `.map` filename already exists.
- Relocate the file renaming to occur after we check that the file was successfully unzipped.

Users and devs can find map IDs using this URL template `http://mapdb.cncnet.org/search.php?game=GAME_ID&search=MAP_NAME_SEARCH_STRING`

Issue: https://github.com/CnCNet/xna-cncnet-client/issues/350

### DIFF
--- a/DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
+++ b/DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
@@ -95,6 +95,9 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
             AddChatBoxCommand(new ChatBoxCommand("CHANGETUNNEL",
                 "Change the used CnCNet tunnel server (game host only)".L10N("UI:Main:CommandChangeTunnel"),
                 true, (s) => ShowTunnelSelectionWindow("Select tunnel server:".L10N("UI:Main:SelectTunnelServer"))));
+            AddChatBoxCommand(new ChatBoxCommand("DOWNLOADMAP",
+                "Download a map from CNCNet's map server using a map ID and an optional filename.\nExample: \"/downloadmap MAPID [2] My Battle Map\"".L10N("UI:Command:DownloadMapCommandDescription"),
+                false, DownloadMapByIdCommand));
         }
 
         public event EventHandler GameLeft;
@@ -129,6 +132,7 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
         private string gameFilesHash;
 
         private List<string> hostUploadedMaps = new List<string>();
+        private List<string> chatCommandDownloadedMaps = new List<string>();
 
         private MapSharingConfirmationPanel mapSharingConfirmationPanel;
 
@@ -1569,6 +1573,14 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
                 channel.SendCTCPMessage(MAP_SHARING_FAIL_MESSAGE + " " + e.SHA1, QueuedMessageType.SYSTEM_MESSAGE, 9);
                 return;
             }
+            else if (chatCommandDownloadedMaps.Contains(e.SHA1))
+            {
+                // Notify the user that their chat command map download failed.
+                // Do not notify other users with a CTCP message as this is irrelevant to them.
+                AddNotice("Downloading map via chat command has failed. Check the map ID and try again.".L10N("UI:Command:DownloadMapCommandFailedGeneric"));
+                mapSharingConfirmationPanel.SetFailedStatus();
+                return;
+            }
 
             AddNotice("Requesting the game host to upload the map to the CnCNet map database.".L10N("UI:Main:RequestHostUploadMapToDB"));
 
@@ -1592,6 +1604,14 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
                     GameModeMap = GameModeMaps.Find(gmm => gmm.Map.SHA1 == lastMapSHA1);
                     ChangeMap(GameModeMap);
                 }
+            }
+            else if (chatCommandDownloadedMaps.Contains(e.SHA1))
+            {
+                // Somehow the user has managed to download an already existing sha1 hash.
+                // This special case prevents user confusion from the file successfully downloading but showing an error anyway.
+                AddNotice(returnMessage, Color.Yellow);
+                AddNotice("Map was downloaded, but a duplicate is already loaded from a different filename. This may cause strange behavior.".L10N("UI:Command:DownloadMapCommandDuplicateMapFileLoaded"),
+                    Color.Yellow);
             }
             else
             {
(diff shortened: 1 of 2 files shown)


DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
```diff
@@ -1593,6 +1605,16 @@ private void MapSharer_HandleMapDownloadComplete(SHA1EventArgs e)
                     ChangeMap(GameModeMap);
                 }
             }
+            else if (chatCommandDownloadedMaps.Contains(e.SHA1))
+            {
+                // Somehow the user has managed to download an already existing sha1 hash.
+                // This special case prevents user confusion by the file successfully downloading but showing an error.
+                AddNotice(returnMessage, Color.Yellow);
+                AddNotice(
+                    "Map was downloaded via chat, but a duplicate of the map with a separate filename is already loaded. This may cause strange behavior.",
```
Metadorius, line 1614: Needs a L10N call here, also not sure if we need to split the line like this.

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
```diff
@@ -1569,6 +1573,14 @@ private void MapSharer_HandleMapDownloadFailed(SHA1EventArgs e)
                 channel.SendCTCPMessage(MAP_SHARING_FAIL_MESSAGE + " " + e.SHA1, QueuedMessageType.SYSTEM_MESSAGE, 9);
                 return;
             }
+            else if (chatCommandDownloadedMaps.Contains(e.SHA1))
+            {
+                // Notify the user that their chat command map download failed.
+                // Do not notify other users with a CTCP message as this is irrelevant to them.
+                AddNotice("Downloading map via chat command has failed. Check your map ID and try again.");
```
Metadorius, line 1580: Needs L10N call here.

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
```diff
@@ -1738,6 +1760,77 @@ private void HandleMapSharingBlockedMessage(string sender)
...
+        private void DownloadMapByIdCommand(string parameters)
+        {
+            string sha1;
+            string mapName;
+            string message;
+
+            // Check if the parameter's contain spaces.
+            // The presence of spaces indicates a user-specified map name.
+            parameters = parameters.Trim();
+            int firstSpaceIndex = parameters.IndexOf(' ');
```
Metadorius, line 1786: Wouldn't this fail if a user supplied one word as an argument?

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
```diff
@@ -1738,6 +1760,77 @@ private void HandleMapSharingBlockedMessage(string sender)
...
+
+            chatCommandDownloadedMaps.Add(sha1);
+
+            // Replace any characters that are not safe for filenames.
+            char replaceUnsafeCharactersWith = '-';
+            // Use a hashset instead of an array for quick lookups in `invalidChars.Contains()`.
+            HashSet<char> invalidChars = new HashSet<char>(Path.GetInvalidFileNameChars());
+            string safeMapName = new String(mapName.Select(c => invalidChars.Contains(c) ? replaceUnsafeCharactersWith : c).ToArray());
+
+            message = String.Format("Attempting to download map via chat command: sha1={0}, mapName={1}", sha1, mapName);
```
Metadorius, line 1827: And here.

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
```diff
@@ -1738,6 +1760,77 @@ private void HandleMapSharingBlockedMessage(string sender)
...
+            // Remove erroneous "?" when someone double clicks and copies a map ID from the cncnet search endpoint.
+            // There is some weird whitespace that gets copied to chat as a "?" at the end. It's hard to spot, so just hold
+            // the user's hand.
+            sha1 = sha1.Replace("?", "");
+
+            GameModeMap loadedMap = GameModeMaps.Find(gmm => gmm.Map.SHA1 == sha1);
+            if (loadedMap != null)
+            {
+                message = String.Format(
+                    "The map for the ID \"{0}\" is already loaded from \"{1}.map\", delete the exiting file first before trying to download it again.",
```
Metadorius, line 1811: And here.

Metadorius (commented):
Looks good to me except the lack of localization (L10N) calls. Also is such multi-lining of arguments consistent with other places in client?

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
alexlambson: My bad, the line was long and I'm used to python where we aggressively split lines. I'll remove it and add the internationalization calls.

Metadorius: No worries, check the other changes, IIRC there was a line like that too. It's OK to split lines IIRC, just need to add a newline after them and leave the closing bracket at last argument line IIRC (consult the [styleguide](https://github.com/CnCNet/xna-cncnet-client/blob/develop/Contributing.md) if unsure).

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
alexlambson: It returns `-1` if there isn't a space. 

In my tests I was usually doing `/downloadmap HASH` and `/downloadmap HASH onewordname` and it seemed to work.

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
alexlambson: updated all strings that were relevant to use L10N

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
```diff
@@ -1738,6 +1758,79 @@ private void HandleMapSharingBlockedMessage(string sender)
...
+            parameters = parameters.Trim();
+            // Check if the parameter's contain spaces.
+            // The presence of spaces indicates a user-specified map name.
+            int firstSpaceIndex = parameters.IndexOf(' ');
+
+            if (firstSpaceIndex == -1)
+            {
+                // The user did not supply a map name.
+                sha1 = parameters;
+                mapName = "user_chat_command_download".L10N("UI:Main:DownloadCommandDefaultMapName");
```
Metadorius, line 1791: Not sure if this one needs to be localized, also starting from this you started writing `DownloadCommand...` instead of `DownloadMapCommand`.

Also maybe it would be better to use `UI:Command:...` instead? or `UI:Commands:...`. Though if other commands are not following the same style it may be better to do this as a separately if there is a lot to change.

Metadorius (changes requested):

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
alexlambson: Lol damn, I was worried I may have switched conventions. I should have double checked. 

My thought behind localizing that string is in case e.g. a Chinese user was looking at their files and wondering where they came from. I will remove it though.

alexlambson: I'll switch mine to `UI:Command:` since there aren't already translations for them, so renaming them now is easier. Perhaps there should be a ticket to switch the other commands to `UI:Command` or `UI:ChatCommand`.

Metadorius (approved):

Issue #352, opened 2022-07-27
Enhancement: Show "Custom" game mode if a map is added to the Maps/Custom/ folder while the client is running


alexlambson:
I noticed that the `Custom` gamemode doesn't appear if the `Maps/Custom` folder was empty at client start, even if I use `/downloadmap` and it downloads successfully or I add a map file to the folder using my OS. 

Perhaps we should add some kind of file-watcher to the `Maps/Custom` folder to constantly check if we need to update the `Custom` maps list and game mode dropdown.

Reproduction:

- Move or delete all files from `Maps/Custom`
- Start the client 
- Start a multiplayer lobby. `Custom` should not be an available game mode. 
- Use `/downloadmap HASH` to download a map, or add a `.map` file to `Maps/Custom/` using your Operating System.
- Open the game mode drop down again.

**Expected:** `Custom` is added to the game mode dropdown after a successful map download or adding a `.map` file to the `Maps/Custom` folder. The newly added map is listed as an option in the `Custom` map list.

**Actual:** `Custom` remains hidden even though there are maps in the folder now.

Metadorius:
Yeah this would be a good improvement, I actually suggested that too some time before. Could set up a file watcher that would update the custom map list.

alexlambson:
I can give it a shot after work if everyone else is okay with adding this. I think we'd only need it for `Custom`

Metadorius:
Yeah previously everyone else agreed it would be a nice addition. Non-custom maps are loaded from a pre-made list anyway.

alexlambson:
I'll definitely give it a shot then

PR #358, opened 2022-08-02
#352: Add custom map directory file watcher

alexlambson:
- Add a file watcher that will detect when a map is added or deleted, then update the game mode dropdown according to the new map list
- Add and use `CustomMapsDirectory` field to `MapLoader` to avoid hard coded strings.
- Add `GetLoadedMapBySha1` to remove some duplicate `GameModeMaps.Find()`.
- Add a function `RefreshGameModeDropdown` to modify the dropdown after client is loaded. The function is a bit complicated due to doing inline modifications to the items and keeping the selected mode.

Issue: https://github.com/CnCNet/xna-cncnet-client/issues/352


DXMainClient/Domain/Multiplayer/MapLoader.cs
```diff
@@ -248,37 +399,40 @@ public Map LoadCustomMap(string mapPath, out string resultMessage)
...
+                resultMessage = $"Loading map {mapPath} failed!";
 
-                        return null;
-                    }
-                }
+                return null;
+            }
 
-                Logger.Log("LoadCustomMap: Map " + mapPath + " added succesfully.");
+            // Make sure we don't accidentally load the same map twice.
```
alexlambson, line 414: This area is just a switch from 

```csharp

if (!error)
{
  doStuff();
}

handleError();
```

to 

```csharp
if (error)
{
  handleError();
}

doStuff();
```

DXMainClient/Domain/Multiplayer/MapLoader.cs
```diff
@@ -55,6 +88,121 @@ public void LoadMapsAsync()
...
         }
 
+        /// <summary>
+        /// Start the file watcher for the custom map directory.
+        ///
+        /// This will refresh the game modes and map lists when a change is detected. 
+        /// </summary>
+        public void StartCustomMapFileWatcher()
+        {
+            customMapFileWatcher = new FileSystemWatcher($"{ProgramConstants.GamePath}{CustomMapsDirectory}");
```
Rans4ckeR, line 98:
```suggestion
            customMapFileWatcher = new FileSystemWatcher(SafePath.CombineDirectoryPath(ProgramConstants.GamePath, CustomMapsDirectory));
```

Concatenating string paths won't work cross-platform. Always use SafePath when evaluating file or folder paths.

Rans4ckeR (changes requested):

DXMainClient/Domain/Multiplayer/MapLoader.cs
```diff
@@ -55,6 +88,121 @@ public void LoadMapsAsync()
...
+        ///
+        /// Adds the map to the GameModeMaps and updates the UI.
+        /// </summary>
+        /// <param name="sender">Sent by the file system watcher</param>
+        /// <param name="e">Sent by the file system watcher</param>
+        public void HandleCustomMapFolder_Created(object sender, FileSystemEventArgs e)
+        {
+            // Get the map filename without the extension.
+            // The extension gets added in LoadCustomMap so we need to excise it to avoid "file.map.map".
+            string name = e.Name.EndsWith(MAP_FILE_EXTENSION) ? e.Name.Remove(e.Name.Length - MAP_FILE_EXTENSION.Length) : e.Name;
```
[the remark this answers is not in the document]
alexlambson: Wow, this library made a lot of my if checks easier. Thank you.

DXMainClient/Domain/Multiplayer/MapLoader.cs
alexlambson: Thanks for the review. I think I found everywhere where I was concatenating paths and replaced them.

DXMainClient/Domain/Multiplayer/MapLoader.cs
```diff
@@ -14,13 +14,19 @@ namespace DTAClient.Domain.Multiplayer
     public class MapLoader
     {
         public const string MAP_FILE_EXTENSION = ".map";
-        private const string CUSTOM_MAPS_DIRECTORY = "Maps/Custom";
+        private const string CUSTOM_MAPS_DIRECTORY = "Maps\\Custom";
```
SadPencil, line 17: it breaks the client on non-Windows. just use `/` or SafePath

DXMainClient/Domain/Multiplayer/MapLoader.cs
```diff
@@ -46,6 +57,28 @@ public class MapLoader
         /// </summary>
         private string[] AllowedGameModes = ClientConfiguration.Instance.AllowedCustomGameModes.Split(',');
 
+        private FileSystemWatcher customMapFileWatcher;
+
+        /// <summary>
+        /// Check to see if a map matching the sha1 ID is already loaded.
```
SadPencil, line 63: SHA-1 instead of sha1 in text

DXMainClient/Domain/Multiplayer/MapLoader.cs
```diff
@@ -46,6 +57,28 @@ public class MapLoader
...
+        /// </summary>
+        /// <param name="sha1">The map ID to search the loaded maps for.</param>
+        /// <returns></returns>
+        public bool IsMapAlreadyLoaded(string sha1)
+        {
+            return GetLoadedMapBySha1(sha1) != null;
+        }
+
+        /// <summary>
+        /// Search the loaded maps for the sha1, return the map if a match is found.
```
SadPencil, line 73: SHA-1 instead of sha1 in text

DXMainClient/Domain/Multiplayer/MapLoader.cs
```diff
@@ -46,6 +57,28 @@ public class MapLoader
...
+        /// </summary>
+        /// <param name="sha1">The map ID to search the loaded maps for.</param>
+        /// <returns>The map matching the sha1 if one was found.</returns>
```
SadPencil, line 76: SHA-1 instead of sha1 in text

DXMainClient/Domain/Multiplayer/MapLoader.cs
```diff
@@ -55,6 +88,137 @@ public void LoadMapsAsync()
...
+        public void StartCustomMapFileWatcher()
+        {
+            customMapFileWatcher = new FileSystemWatcher(SafePath.CombineDirectoryPath(ProgramConstants.GamePath, CustomMapsDirectory));
+
+            customMapFileWatcher.Filter = $"*{MAP_FILE_EXTENSION}";
+            customMapFileWatcher.NotifyFilter = NotifyFilters.Attributes
+                                 | NotifyFilters.CreationTime
+                                 | NotifyFilters.DirectoryName
+                                 | NotifyFilters.FileName
+                                 | NotifyFilters.LastAccess
```
SadPencil, line 105: Why monitoring `LastAccess`?

SadPencil: Also why don't just use the default NotifyFilter

SadPencil (changes requested):

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
```diff
@@ -1595,14 +1595,14 @@ private void MapSharer_HandleMapDownloadComplete(SHA1EventArgs e)
         {
             string mapFileName = MapSharer.GetMapFileName(e.SHA1, e.MapName);
             Logger.Log("Map " + mapFileName + " downloaded, parsing.");
-            string mapPath = "Maps/Custom/" + mapFileName;
+            string mapPath = SafePath.CombineFilePath(MapLoader.CustomMapsDirectory, mapFileName);
```
Rans4ckeR, line 1598: I think this might break things since the string might be used for other stuff that relies on the previous exact value e.g. section headers in ini files. I recommend only using SafePath when evaluating the path (when actually accessing the filesystem).

DXMainClient/Domain/Multiplayer/MapLoader.cs
alexlambson: I was originally thinking we'd need to reload the map every time Final Alert saved it, but I realized that doesn't matter because the game just loads the current version of the file. Guess I forgot to pull this.

Metadorius (commented):
Apparently I wrote a(n incomplete I think) review on some stylistical stuff and forgot to publish it, so I'll leave it here, better late than never :P

alexlambson:
I *think* I addressed all the feedback

DXMainClient/Domain/Multiplayer/MapLoader.cs
```diff
@@ -55,6 +88,132 @@ public void LoadMapsAsync()
...
+                Logger.Log($"Renaming file. New extension is not '{MAP_FILE_EXTENSION}', moving on: file={e.Name}");
+                return;
+            }
+
+            Map map = LoadCustomMap(relativeMapPath, out string result);
+
+            if (map == null)
+            {
+                Logger.Log($"Failed to load renamed map file. Map is likely already loaded: original={e.OldName}, new={e.Name}, reason={result}");
+            }
```
Metadorius, line 199:
```suggestion
                Logger.Log($"Failed to load renamed map file. Map is likely already loaded: original={e.OldName}, new={e.Name}, reason={result}");
```

DXMainClient/Domain/Multiplayer/MapLoaderEventArgs.cs
```diff
@@ -0,0 +1,18 @@
...
+    /// </summary>
+    public class MapLoaderEventArgs : EventArgs
+    {
+        public MapLoaderEventArgs(Map map)
+        {
+            Map = map;
+        }
+
+        public Map Map { get; private set; }
+
```
Metadorius, line 16:
```suggestion
```

Metadorius (changes requested):
You missed some of my stylistical comments and there's something wrong in your code that CI fails, otherwise IMO it's fine.

DXMainClient/Domain/Multiplayer/MapLoader.cs
```diff
@@ -55,6 +88,132 @@ public void LoadMapsAsync()
...
+        }
+
+        /// <summary>
+        /// Handle errors in the filewatcher.
+        ///
+        /// Not much to do other than log a stack trace.
+        /// </summary>
+        /// <param name="sender"></param>
+        /// <param name="e"></param>
```
Rampastring, line 208: You can remove these param comments.

Also `Not much to do other than log a stack trace.` is an implementation detail, I'd leave it out from the documentation comments.

DXMainClient/Domain/Multiplayer/MapLoader.cs
```diff
@@ -55,6 +88,132 @@ public void LoadMapsAsync()
...
+        public void HandleCustomMapFolder_Error(object sender, ErrorEventArgs e)
```
Rampastring, line 209: As far as I can see, this is something that should never be called outside of this class. Make it `private`. The same also applies to other file watcher event handlers.

Rampastring (changes requested):
Does this currently handle a case where existing custom maps are modified? If that is done, the SHA1 for the map in question needs to be recalculated. See `FileSystemWatcher.Changed` event.

DXMainClient/Domain/Multiplayer/MapLoader.cs
```diff
@@ -55,6 +88,114 @@ public void LoadMapsAsync()
...
+
+            customMapFileWatcher.Filter = $"*{MAP_FILE_EXTENSION}";
+            customMapFileWatcher.NotifyFilter = NotifyFilters.Attributes
+                                 | NotifyFilters.CreationTime
+                                 | NotifyFilters.DirectoryName
+                                 | NotifyFilters.FileName
+                                 | NotifyFilters.LastAccess
+                                 | NotifyFilters.LastWrite
+                                 | NotifyFilters.Security
+                                 | NotifyFilters.Size;
```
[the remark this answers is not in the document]
alexlambson: I removed the filters per another review

DXMainClient/Domain/Multiplayer/MapLoader.cs
alexlambson: It does not, but the map will still load if you try to play it. The Client can already handle that case. I often just save the map in FA2 then just click "play" and it works.

alexlambson:
> Does this currently handle a case where existing custom maps are modified? If that is done, the SHA1 for the map in question needs to be recalculated. See `FileSystemWatcher.Changed` event.

It does not, but the map will still load if you try to play it. The Client can already handle that case. I often just save the map in FA2 then just click "play" and it works.

DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
```diff
@@ -2315,5 +2318,63 @@ public bool LoadGameOptionPreset(string name)
...
+        /// Handle the GameModeMapsUpdated event from the MapLoader.
+        ///
+        /// Updates the gamemode dropdown for new maps being added while the client is running
+        /// </summary>
+        /// <param name="sender"></param>
+        /// <param name="e"></param>
+        private void MapLoader_GameModeMapsUpdated(object sender, MapLoaderEventArgs e)
+        {
+            RefreshGameModeDropdown();
+        }
```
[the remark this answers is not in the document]
alexlambson: done

DXMainClient/Domain/Multiplayer/MapLoader.cs
alexlambson: added a return so it's no longer a one-liner

DXMainClient/Domain/Multiplayer/MapLoader.cs
alexlambson: Made all private

DXMainClient/Domain/Multiplayer/MapLoaderEventArgs.cs
alexlambson: fixed

DXMainClient/Domain/Multiplayer/MapLoader.cs
```diff
@@ -55,6 +88,114 @@ public void LoadMapsAsync()
...
+            // Get the map filename without the extension.
+            // The extension gets added in LoadCustomMap so we need to excise it to avoid "file.map.map".
+            string name = e.Name.EndsWith(MAP_FILE_EXTENSION) ? e.Name.Remove(e.Name.Length - MAP_FILE_EXTENSION.Length) : e.Name;
+            string relativeMapPath = $"{CustomMapsDirectory}{name}";
+            Map map = LoadCustomMap(relativeMapPath, out string result);
+
+            if (map == null)
+            {
+                Logger.Log($"Failed to load map file that was create / moved: mapPath={name}, reason={result}");
+            }
```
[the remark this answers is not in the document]
alexlambson: fixed

DXMainClient/Domain/Multiplayer/MapLoader.cs
alexlambson: fixed

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
```diff
@@ -1595,32 +1595,23 @@ private void MapSharer_HandleMapDownloadComplete(SHA1EventArgs e)
         {
             string mapFileName = MapSharer.GetMapFileName(e.SHA1, e.MapName);
             Logger.Log("Map " + mapFileName + " downloaded, parsing.");
-            string mapPath = "Maps/Custom/" + mapFileName;
-            Map map = MapLoader.LoadCustomMap(mapPath, out string returnMessage);
```
alexlambson, line 1599: This is now handled by the file watcher to prevent a race condition

DXMainClient/Domain/Multiplayer/CnCNet/MapSharer.cs
```diff
@@ -365,18 +365,27 @@ public static string GetMapFileName(string sha1, string mapName)
 
         private static string DownloadMain(string sha1, string myGame, string mapName, out bool success)
         {
-            string customMapsDirectory = SafePath.CombineDirectoryPath(ProgramConstants.GamePath, "Maps", "Custom");
+            string customMapsPath = SafePath.CombineDirectoryPath(ProgramConstants.GamePath, "Maps", "Custom");
+            string tempDownloadPath = SafePath.CombineDirectoryPath(ProgramConstants.GamePath, "Maps", "temp");
```
alexlambson, line 369: Use a temporary download directory so that the filewatcher doesn't prematurely load the `.map` file before it is fully extracted and renamed.

PR #364, opened 2022-09-03
Fix crashes caused by null PlayerExtraOptionsPanel

SadPencil:
In `GameLobbyBase.cs`:
```cs
protected PlayerExtraOptions GetPlayerExtraOptions() =>
    PlayerExtraOptionsPanel == null ? new PlayerExtraOptions() : PlayerExtraOptionsPanel.GetPlayerExtraOptions();
```

And in `PlayerExtraOptions.cs`:
```cs
public List<TeamStartMapping> TeamStartMappings { get; set; }  // this is by default null
```

Which causes a crash of a null pointer at `PlayerExtraOptions.ToString()`, where:
```cs
TeamStartMapping.ToListString(TeamStartMappings) // TeamStartMappings is null, while ToListString() crash on null
```

-----------------

Also, in `MultiplayerGameLobby.cs`:
```cs
PlayerExtraOptionsPanel.SetIsHost(isHost);
```

`PlayerExtraOptionsPanel` can be null.

### DIFF
--- a/DXMainClient/DXGUI/Multiplayer/GameLobby/MultiplayerGameLobby.cs
+++ b/DXMainClient/DXGUI/Multiplayer/GameLobby/MultiplayerGameLobby.cs
@@ -566,7 +566,7 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
             Locked = false;
 
             UpdateMapPreviewBoxEnabledStatus();
-            PlayerExtraOptionsPanel.SetIsHost(isHost);
+            PlayerExtraOptionsPanel?.SetIsHost(isHost);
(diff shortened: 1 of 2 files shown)


Rans4ckeR (approved):
I had to apply these same changes to my build in order to be able to run TS. Does this mean TS is currently broken? Is this specific to TS?

SadPencil:
> I had to apply these same changes to my build in order to be able to run TS. Does this mean TS is currently broken? Is this specific to TS?

I encounter this problem when testing on migrating MO 336 which is an Ares build. So I think it is broken in all builds.

devo1929 (approved):

Metadorius (approved):

PR #366, opened 2022-09-11
Use 24-hour time format in TopBar to prevent the string from going beyond the border

pzhlkj6612:
## Summary

If the time format settings of the OS contains an AM/PM designator, it will go beyond the right border of the window.

## Why

The placeholder of `lblTime` is "`99:99:99`":

https://github.com/CnCNet/xna-cncnet-client/blob/624dfccd23011970ed2ce668888fd71132b0a560/DXMainClient/DXGUI/Generic/TopBar.cs#L161-L167

So I think using 24-hour format (`HH:mm:ss`) is OK.

## Comparison

20220911-CnCNet-DXClient-TopBar-time-format

## And...

There is a wrong time format string:

https://github.com/CnCNet/xna-cncnet-client/blob/624dfccd23011970ed2ce668888fd71132b0a560/DXMainClient/Domain/Multiplayer/CnCNet/MapSharer.cs#L418-L422

We should use "`HH:mm:ss`" instead of "`hh:mm:ss`" because "`hh`" is a number from 01 to 12. Since the code is commented out and not relevant to this PR, I didn't modify it.

See also: [The "hh" custom format specifier § Custom date and time format strings | Microsoft Docs](https://docs.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings#hhSpecifier)

### DIFF
--- a/DXMainClient/DXGUI/Generic/TopBar.cs
+++ b/DXMainClient/DXGUI/Generic/TopBar.cs
@@ -441,7 +441,7 @@ namespace DTAClient.DXGUI.Generic
 
             DateTime dtn = DateTime.Now;
 
-            lblTime.Text = Renderer.GetSafeString(dtn.ToLongTimeString(), lblTime.FontIndex);
+            lblTime.Text = Renderer.GetSafeString(dtn.ToString("HH:mm:ss"), lblTime.FontIndex);
             string dateText = Renderer.GetSafeString(dtn.ToShortDateString(), lblDate.FontIndex);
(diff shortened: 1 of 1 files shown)


Rans4ckeR (changes requested):
The application should use correct localization, so en-US systems should display AM/PM.
A proper solution is to update the layout instead. For example by changing the placeholder like this:

```C#
lblTime.Text = new DateTime(1, 1, 1, 23, 59, 59).ToLongTimeString();
```

PR #367, opened 2022-09-12
Reimplement FindChild and add FindChildrenStartWith

SadPencil:
The original `FindChild<T>` is implemented in a less extendable way. It is nicely reimplemented now and a new function named `FindChildrenStartWith` is added.

I am working on a feature that allow mod developpers customize buttons with their own tags, therefore I need to match these buttons with prefix name but the current source code does not provide such a function.


Metadorius:
Closing because of existance of superseding PR #518.

Issue #368, opened 2022-09-12
Show colors in color drop down instead of text


devo1929:

pzhlkj6612:
How about showing both color and text in color drop down?

![20231126-CnCNet-issue-368-Show both color and text in color drop down](https://github.com/CnCNet/xna-cncnet-client/assets/29089388/8e002f3c-57a1-4379-bb20-2522e569b106)

pzhlkj6612:
> I am afraid it would be difficult to select a dark color if so?

Hmm.. How about using white text for the name of all colors and using rectangles instead of squares for color blocks?

pzhlkj6612:
I'm going to add a PR link here: #369.

PR #369, opened 2022-09-12
#368 show colors in color drop down

devo1929:
Decreased width of color drop down allows for increase of width of player name and side. Color drop down is now the same width as team and start drop downs.

image


DXMainClient/Domain/Multiplayer/MultiplayerColor.cs
```diff
@@ -17,6 +18,10 @@ public class MultiplayerColor
 
         private static List<MultiplayerColor> colorList;
 
+        private static string randomColorLabel;
+
+        private static readonly string RandomColorDefaultLabel = "Random".L10N("UI:Main:RandomSide");
```
SadPencil, line 23:
```suggestion
        private static readonly string RandomColorDefaultLabel = "Random".L10N("UI:Main:RandomColor");
```

And also considering making this a readonly property to prevent L10N being called before loading the translation table (although it should works without being a property as long as the class is initialized after L10N initialized but it is not straightforward enough).

SadPencil (changes requested):

DXMainClient/Domain/Multiplayer/MultiplayerColor.cs
devo1929: I'm not sure I understand. It is currently `readonly`. Are you saying it _shouldn't_ be readonly?

devo1929: Good catch on the incorrect string.

SadPencil: > I'm not sure I understand. It is currently `readonly`. Are you saying it _shouldn't_ be readonly?

No. I mean changing `=` to `=>`

devo1929: Ahhhhh, I got it. Will do.

devo1929: Done.

SadPencil (approved):
LGTM

Rampastring:
Seeing the gap between the dropdown border and the color, it might be best if either a custom dropdown control was added that got rid of the gap, or I made the gap customizable in XNAUI.

pzhlkj6612 (changes requested):
The design of this PR has side-effect on COOP maps. I'm now testing the YR client btw.

### Symptom

disallowed-colors-cannot-be-recognized-in-new-color-drop-down

We cannot see which colors are disallowed.

### Test

- Make a map. `Maps\Custom\yuri.map` for example.
- Modify `INI\MPMaps.ini` and add these lines:

```INI
[MultiMaps]
; ...
test=Maps\Custom\yuri

[Maps\Custom\yuri]
IsCoopMission=yes
GameModes=Battle
DisallowedPlayerSides=5,6,7,8 ; World Socialist Alliance
DisallowedPlayerColors=0,2,4,6 ; Yellow, Blue, Orange, Purple

```

.

pzhlkj6612:
If we don't have to decrease the width of color drop down, I think we can try the design like I said in https://github.com/CnCNet/xna-cncnet-client/issues/368#issuecomment-1826837154 to eliminate the side-effect on COOP maps with disallowed colors. Screenshot:

both-color-block-and-name-in-color-drop-down

Code:

```diff
                 foreach (MultiplayerColor mpColor in MPColors)
                 {
                     if(mpColor.Name.StartsWith("$"))
-                        ddPlayerColor.AddItem(string.Empty, AssetLoader.CreateTexture(mpColor.XnaColor, ddPlayerColor.Width - 2, ddPlayerColor.ItemHeight + 2));
+                        ddPlayerColor.AddItem(new XNADropDownItem()
+                        {
+                            Text = mpColor.Name.TrimStart('$'),
+                            Texture = AssetLoader.CreateTexture(mpColor.XnaColor, ddPlayerColor.ItemHeight - 2, ddPlayerColor.ItemHeight - 2),
+                        });
                     else
                         ddPlayerColor.AddItem(mpColor.Name, mpColor.XnaColor);
                 }

```

<br>

And of course, we can decrease the width of color drop down by reducing the width of color blocks. For example:

```csharp
Texture = AssetLoader.CreateTexture(..., (ddPlayerColor.ItemHeight - 2) / 3, ...),

```

.

DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
```diff
@@ -784,9 +784,14 @@ protected void InitPlayerOptionDropdowns()
                 ddPlayerColor.ClientRectangle = new Rectangle(
                     ddPlayerSide.Right + playerOptionHorizontalMargin,
                     ddPlayerName.Y, colorWidth, DROP_DOWN_HEIGHT);
-                ddPlayerColor.AddItem("Random".L10N("UI:Main:RandomColor"), AssetLoader.GetColorFromString(randomColor));
+                ddPlayerColor.AddItem(MultiplayerColor.GetRandomColorLabel(), AssetLoader.GetColorFromString(randomColor));
                 foreach (MultiplayerColor mpColor in MPColors)
-                    ddPlayerColor.AddItem(mpColor.Name, mpColor.XnaColor);
+                {
+                    if(mpColor.Name.StartsWith("$"))
+                        ddPlayerColor.AddItem(string.Empty, AssetLoader.CreateTexture(mpColor.XnaColor, ddPlayerColor.Width - 2, ddPlayerColor.ItemHeight + 2));
```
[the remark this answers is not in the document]
MahBoiDeveloper: Anyway this implementation isn't good enough, because doesn't provide client modders to choose what color column they want to use.

MahBoiDeveloper (changes requested):

PR #370, opened 2022-09-14
Add campaign group selector

SadPencil:
This PR is backward compatible as by default the option is off. Modders can enable the `CampaignTagSelectorEnabled` option to use this feature.

This feature enables modders to make a "choose your side" interface before the campaign selector. Choosing among "Act 1/Act 2/..." or any other tags is also possible.

The screenshot below shows the feature. (I am not an expert in beautifying user interfaces so it is only a working example. Fully configurable through ini files.)

img1
img2
img3

The modders define tags for missions in `battle.ini` and customize the selector window in `CampaignTagSelector.ini` using the new INItializableWindow ini format.

I use TSC v6 client as an example. The corresponding files are attached here.

[TSC.v6.example.zip](https://github.com/CnCNet/xna-cncnet-client/files/9687387/TSC.v6.example.zip)

-----------------------------------------------------
Prerequisite of this PR: #364 (merged now) and #367 (requesting merge).

### DIFF
--- a/ClientCore/ClientConfiguration.cs
+++ b/ClientCore/ClientConfiguration.cs
@@ -277,6 +277,8 @@ namespace ClientCore
 
         public string AllowedCustomGameModes => clientDefinitionsIni.GetStringValue(SETTINGS, "AllowedCustomGameModes", "Standard,Custom Map");
 
+        public bool CampaignTagSelectorEnabled => clientDefinitionsIni.GetBooleanValue(SETTINGS, "CampaignTagSelectorEnabled", false);
+
         public string GetGameExecutableName()
         {
             string[] exeNames = clientDefinitionsIni.GetStringValue(SETTINGS, "GameExecutableNames", "Game.exe").Split(',');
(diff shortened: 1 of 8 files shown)


SadPencil:
Ready for review now. Prerequisite #367.

Rampastring:
I'm not a fan of the design where the Battle.ini is constantly re-read and re-parsed here.

I'd rather read Battle.ini only once, have all the missions in the `Missions` list, and filter what is actually displayed in the list.

Metadorius:
Actually wouldn't it be better to have one INI per window? So separate screens would have different `Battle.ini` alternatives.

Rampastring:
I think the single file gives more control when it comes to the "Show All" view.

SadPencil:
> I'm not a fan of the design where the Battle.ini is constantly re-read and re-parsed here.
> 
> I'd rather read Battle.ini only once, have all the missions in the `Missions` list, and filter what is actually displayed in the list.

Thanks for the suggestion. Code updated.

Metadorius:
Closing because of the existence of superseding PR #518.

PR #341, opened 2022-06-12
.NET 6 Cross platform

Rans4ckeR:
# Updated frameworks
  - .NET 4.8
    - DirectX targets AnyCPU
    - OpenGL targets AnyCPU
    - XNA targets x86
  - .NET 6.0:
    - net6.0-windows: WinForms, Windows 7 and up
    - net6.0: cross platform
    - DirectX targets AnyCPU
    - OpenGL targets AnyCPU
    - XNA targets x86
# Updated dependencies
  - New Rampastring.Tools NuGet package
  - New Rampastring.XNAUI.* NuGet packages
    - XNA 4.0 Refresh
    - MonoGame DX & GL 3.8.0 (.NET 4.8)
    - MonoGame DX & GL 3.8.1 (.NET 6.0)
  - Replaced Ionic.Zip with System.IO.Compression
  - Use Izo.net NuGet package
  - Use OpenMcdf NuGet package
  - New ClientUpdater local NuGet package
# Build
  - Switch to new project file format
  - New solution configurations and platforms
    - Configurations: AresWindowsDXDebug;AresUniversalGLDebug;AresWindowsGLDebug;AresWindowsXNADebug;
        AresWindowsDXRelease;AresUniversalGLRelease;AresWindowsGLRelease;AresWindowsXNARelease;
        TSWindowsDXDebug;TSUniversalGLDebug;TSWindowsGLDebug;TSWindowsXNADebug;
        TSWindowsDXRelease;TSUniversalGLRelease;TSWindowsGLRelease;TSWindowsXNARelease;
        YRWindowsDXDebug;YRUniversalGLDebug;YRWindowsGLDebug;YRWindowsXNADebug;
        YRWindowsDXRelease;YRUniversalGLRelease;YRWindowsGLRelease;YRWindowsXNARelease
    - Platforms: AnyCPU, x86, x64, ARM64
  - Includes .editorconfig with my default warning rules. This triggers many new warnings. The rules can be reviewed and updated, or I can fix the current warnings in another PR.
  - Note: Switching between XNA or MonoGame configurations within Visual Studio may require you to unload and reload the ClientCore project (or restart VS) regarding MonoGame/XNA dependencies

# Cross platform
  - Adds an additional .NET6.0 'UniversalGL' build without WinForms. Allows client to be run on any device with .NET6.0 installed (new Rampastring.XNAUI package).
  - Improves .NET4.8 compatibility on Mono
  - Cross platform building (build on Linux, run on Windows and vice versa)

[...]

PR #372, opened 2022-09-14
Fixed .editorconfig and added validation to workflows

Belonit:
We use this workflow at Phobos and it works well


Rans4ckeR:
FYI we are planning to merge below PR before all others, which will generate quite some conflicts I'm afraid.
- https://github.com/CnCNet/xna-cncnet-client/pull/341

Metadorius:
I think it would be beneficial if you reset your branch to pre-fix and then reapply all warning fixes.

Belonit:
@Metadorius Yes, I'll do it a little later

Belonit:
I updated this PR and consider it is done

Metadorius (changes requested):
The council ™️ concluded that the style checks should absolutely NOT fail a workflow. Instead look into making a problem matcher and registering it so that it displays issues as inline annotations (same applies for Phobos one as it's quite annoying to get failed runs just because of a stray trailing space): https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md

Belonit:
> The council ™️ concluded that the style checks should absolutely NOT fail a workflow. Instead look into making a problem matcher and registering it so that it displays issues as inline annotations (same applies for Phobos one as it's quite annoying to get failed runs just because of a stray trailing space): https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md

Currently, GitHub Actions limit the annotation count in a workflow run.

- 10 warning annotations, 10 error annotations, and 10 notice annotations per step
- 50 annotations per job (sum of annotations from all the steps)
- 50 annotations per run (separate from the job annotations, these annotations aren’t created by users)

If your workflow may exceed these annotation counts, consider filtering of the log messages which the Problem Matcher is exposed to (e.g. by PR touched files, lines, or other).

Metadorius:
I think those counts are reasonable and we can work with that. Not sure why you closed the PR though?

Belonit:
> I think those counts are reasonable and we can work with that. Not sure why you closed the PR though?

I'm not sure why it closed either. Possibly missclick

Metadorius:
Closing because of a huge load of conflicts and no progress on problem matcher, feel free to reopen/submit a new one, whichever is easier, if you get addressing the conflicts AND problem matcher which doesn't fail the workflow.

PR #374, opened 2022-09-20
Build cleanup, Large address aware

Rans4ckeR:
WindowsXNA default build change from `x86` to `AnyCPU (32-bit preferred)`. Both are 32bit but the latter enables the flag `Large address aware`.

Build dependency optimizations.
Remove dependency on Microsoft.Windows.SDK.NET.dll (25MB) and WinRT.Runtime.dll by targeting net6.0-windows since we are unlikely to use WinRT functions.

Drop `net48` target for `UniversalGL`, it's always `net6.0`. This enables running and debugging `UniversalGL` using `WSL`.

Update packages.

Prerequisites:
- https://github.com/Rampastring/Rampastring.Tools/pull/6
- https://github.com/Rampastring/Rampastring.XNAUI/pull/18
- https://github.com/CnCNet/cncnet-client-updater/pull/6


SadPencil:
Why enabling `large address aware` for the client?

Rans4ckeR:
> Why enabling `large address aware` for the client?

[64-bit programming for Game Developers](https://learn.microsoft.com/en-us/windows/win32/dxtecharts/sixty-four-bit-programming-for-game-developers):

> It is a good practice to specify large-address-aware when building 32-bit applications, by using the linker flag /LARGEADDRESSAWARE, even if the application is not intended for a 64-bit platform, because of the advantages that are gained at no cost.

Metadorius (approved):

PR #376, opened 2022-09-23
Fix Map BaseFilePath parsing

Rans4ckeR:


Starkku:
Looks fine and fixes the problem with map loading. Merging.

PR #388, opened 2022-11-10
Make locale-aware time fully displayed in TopBar to prevent the string from going beyond the border

pzhlkj6612:
Hello!

If the time format of the current locale contains an AM/PM designator, it will go beyond the right border of the window.

I tried to remove the AM/PM designator; @Rans4ckeR reviewed my code and said:

> The application should use correct localization, so en-US systems should display AM/PM.
> A proper solution is to update the layout instead. For example by changing the placeholder like this:
> 
> ```C#
> lblTime.Text = new DateTime(1, 1, 1, 23, 59, 59).ToLongTimeString();
> ```
> 
> *source: https://github.com/CnCNet/xna-cncnet-client/pull/366#pullrequestreview-1103272340*

That's better, isn't it? So I make this PR. 

Here is the screenshot:

20221110-CnCNet-DXClient-TopBar-time-format

This PR will close #366 if gets merged.

### DIFF
--- a/DXMainClient/DXGUI/Generic/TopBar.cs
+++ b/DXMainClient/DXGUI/Generic/TopBar.cs
@@ -161,7 +161,7 @@ namespace DTAClient.DXGUI.Generic
             lblTime = new XNALabel(WindowManager);
             lblTime.Name = "lblTime";
             lblTime.FontIndex = 1;
-            lblTime.Text = Renderer.GetSafeString(new DateTime(1, 1, 1, 23, 59, 59).ToLongTimeString(), lblDate.FontIndex);
+            lblTime.Text = Renderer.GetSafeString(new DateTime(1, 1, 1, 23, 59, 59).ToLongTimeString(), lblTime.FontIndex);
             lblTime.ClientRectangle = new Rectangle(Width -
                 (int)Renderer.GetTextDimensions(lblTime.Text, lblTime.FontIndex).X - 12, 4,
                 lblTime.Width, lblTime.Height);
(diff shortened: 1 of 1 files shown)


DXMainClient/DXGUI/Generic/TopBar.cs
```diff
@@ -161,7 +161,7 @@ public override void Initialize()
             lblTime = new XNALabel(WindowManager);
             lblTime.Name = "lblTime";
             lblTime.FontIndex = 1;
-            lblTime.Text = "99:99:99";
+            lblTime.Text = Renderer.GetSafeString(new DateTime(1, 1, 1, 23, 59, 59).ToLongTimeString(), lblDate.FontIndex);
```
Rampastring, line 164: `lblDate.FontIndex`, shouldn't it rather be `lblTime.FontIndex`?

Rampastring (changes requested):

DXMainClient/DXGUI/Generic/TopBar.cs
pzhlkj6612: Oh, thanks. I just directly copied the ```Renderer.GetSafeString(DateTime.Now.ToShortDateString(), lblDate.FontIndex);``` line here.

pzhlkj6612: Fixed in 2dc65ac55dc69804628fc78f29b2f03dceadf1b5 .
Please squash the two commits when merging this PR, thank you.

Rampastring (approved):

PR #390, opened 2022-11-14
Dependency injection

devo1929:
This adds dependency injection for all services (or service-like classes) AND all XNAControl based controls. The class `ClientGUICreator` was re-purposed to facilitate this dependency injection for all XNAControl based controls. They can be instantiated as a part of dependency injection in a constructor or when instantiated via an INI file. Controls are simply registered in a different way, but the backing logic mostly remains intact.


DXMainClient/DXGUI/Generic/LoadingScreen.cs
```diff
@@ -100,71 +104,7 @@ private void Finish()
...
-            cncnetLobby.Disable();
-            cncnetGameLobby.Disable();
-            cncnetGameLoadingLobby.Disable();
-            lanLobby.Disable();
-            pmWindow.Disable();
-            optionsWindow.Disable();
-
-            WindowManager.AddAndInitializeControl(topBar);
-            topBar.AddPrimarySwitchable(mainMenu);
-
```
devo1929, line 167: All of the constructor calls above were deleted. They're now done via DI.

All initialization of controls were moved to `MainMenu.cs`. This was necessary due to timing of initialization.

DXMainClient/DXGUI/Generic/LoadingScreen.cs
```diff
@@ -88,9 +94,7 @@ private void Finish()
             ProgramConstants.GAME_VERSION = ClientConfiguration.Instance.ModMode ? 
                 "N/A" : Updater.GameVersion;
 
-            DiscordHandler discordHandler = null;
-            if (!string.IsNullOrEmpty(ClientConfiguration.Instance.DiscordAppId))
-                discordHandler = new DiscordHandler(WindowManager);
```
SadPencil, line 93: I am not familiar with dependency injection and wonder why removing these lines? What would happen if `ClientConfiguration.Instance.DiscordAppId` is empty?

DXMainClient/DXGUI/Generic/MainMenu.cs
```diff
@@ -385,9 +407,9 @@ private void SettingsSaved(object sender, EventArgs e)
                 ProgramConstants.PLAYERNAME = UserINISettings.Instance.PlayerName;
 
             if (UserINISettings.Instance.DiscordIntegration)
-                discordHandler?.Connect();
+                discordHandler.Connect();
             else
-                discordHandler?.Disconnect();
```
SadPencil, line 390: If `ClientConfiguration.Instance.DiscordAppId` is empty, the `discordHandler` would be null. So why replace `?.` with `.`?

DXMainClient/DXGUI/Generic/LoadingScreen.cs
devo1929: First, dependency injection is a way for the backing host application to determine how to instantiate a given class when specified in the constructor of any given class. So, using `public LoadingScreen(DiscordHandler discordHandler)...` as an example of the `LoadingScreen` constructor, dependency injection will automatically handle the instantiation of the `DiscordHandler` class. That is all configured in the `GameClass` class.

As a result, this line of code is no longer required. We're no longer required to manually instantiate this class.

The dependency injection is set up to ALWAYS create a properly instantiated `DiscordHandler` now. It'll never be null if referenced from a constructor. Therefore, there's no need for null checks. Additionally, the check for `ClientConfiguration.Instance.DiscordAppId` to see if we even can use Discord has been moved into the `DiscordHandler` itself. The `DiscordHandler` knows whether or not it can be used internally and will discard attempts to `UpdatePresence` if it cannot be.

DXMainClient/DXGUI/Generic/MainMenu.cs
devo1929: See comment above.

SadPencil:
Sounds good. I don't have questions now and let's wait for others' reviews

Rampastring (approved):

PR #392, opened 2022-11-16
Format Localization project to suppress warnings

SadPencil:
It seems that GitHub show some warnings of `Localization` in `Files changed` section in every PR, which feature is named `Unchanged files with check annotations Beta`.

This PR changes nothing to the code but formating to eliminate or suppress these warnings.


SadPencil:
It seems that GitHub shows more warnings even if some warnings are solved :(

PR #396, opened 2022-11-29
Implement support for parsing and displaying starting positions of Red Alert 1 and Tiberian Dawn maps, with non-isometric cells

Rampastring:
This PR adds optional, configurable support for displaying starting locations of players in Red Alert 1 and Tiberian Dawn maps correctly.

These games use non-isometric cells, which requires some changes to the cell math for figuring out the pixel coordinates of starting locations.

They also have a slightly different map format when it comes to specifying the size of the map, which required changes to the map size parsing code.

To support these features without making TS/YR support more complicated, two new keys were introduced to `ClientDefinitions.ini`:

```cs
[Settings]
UseIsometricCells=
WaypointCoefficient=
```

`WaypointCoefficient` is only used when calculating waypoint positions for RA and TD. `UseIsometricCells` defines whether the use TS/YR or RA1 logic for parsing the map size and for cell math. It defaults to `true` to keep compatibility with existing TS/YR implementations of the client.

As a bonus, this PR fixes some outdated syntax from `Map.cs`.

### DIFF
--- a/ClientCore/ClientConfiguration.cs
+++ b/ClientCore/ClientConfiguration.cs
@@ -213,6 +213,10 @@ namespace ClientCore
 
         public int MaxNameLength => clientDefinitionsIni.GetIntValue(SETTINGS, "MaxNameLength", 16);
 
+        public bool UseIsometricCells => clientDefinitionsIni.GetBooleanValue(SETTINGS, "UseIsometricCells", true);
+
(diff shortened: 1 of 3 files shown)


Rans4ckeR (approved):
.editorconfig doing its work, nice cleanup :)

Metadorius (approved):

PR #403, opened 2022-12-02
Add gitversion

devo1929:


Rans4ckeR (changes requested):
I strongly suggest having all assemblies versioned, not just DXMainClient.

devo1929:
> I strongly suggest having all assemblies versioned, not just DXMainClient.

Will do.

frg2089:
Try used this?
```targets

  <!-- Generate BuildStamp -->
  <Target Name="GenerateBuildStamp" BeforeTargets="GetAssemblyAttributes">
    <!-- This code fragment follows the MIT license agreement. -->
    <!-- Git Commit Hash -->
    <Exec Command="git rev-parse --short HEAD" ConsoleToMsBuild="true">
      <Output TaskParameter="ConsoleOutput" PropertyName="__GitRevision" />
    </Exec>

    <!-- Git Branch -->
    <Exec Command="git rev-parse --symbolic-full-name --abbrev-ref HEAD" ConsoleToMsBuild="true">
      <Output TaskParameter="ConsoleOutput" PropertyName="__GitBranch" />
    </Exec>

    <ItemGroup>
      <AssemblyMetadata Include="BuildTime" Value="$([System.DateTime]::Now.ToString('O'))" />
      <AssemblyMetadata Include="Revision" Value="$(__GitRevision)" />
      <AssemblyMetadata Include="Branch" Value="$(__GitBranch)" />
    </ItemGroup>
  </Target>
```

devo1929:
> I strongly suggest having all assemblies versioned, not just DXMainClient.

Done.

Metadorius:
Heads up, mod authors usually don't even know what NuGet is, and the word "Git" means a cryptic website where it's hard to download something usually 🤣

devo1929:
> Heads up, mod authors usually don't even know what NuGet is, and for them the word "Git" means a cryptic website where it's hard to download something usually 🤣

That's fine. We'll continue to provide the current method which is to download the artifacts from the build manually, but for those that _know_ how to use NuGet can :)

frg2089:
> The main need for this is the auto semantic versioning X.Y.Z. Hoping to get this into nuget so that mods can consume it as a dependency rather than having to manually copy/paste binaries around.

In fact, it does not contain a binary file. It only needs git in the path.
We can use the `FileVersion` and `AssemblyVersion` properties to generate X.Y.Z style versions.

Rans4ckeR (approved):

PR #404, opened 2022-12-04
VSCode Support

frg2089:


Metadorius (changes requested):
Should use input variables of `pickString` type to not duplicate a load of tasks for different configs, with a variable for game, platform and debug/release config.
https://code.visualstudio.com/docs/editor/variables-reference#_input-variables

.vscode/launch.json
```diff
@@ -0,0 +1,137 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Launch Ares UniversalGL Debug",
+      "type": "coreclr",
+      "request": "launch",
+      "preLaunchTask": "Build Ares UniversalGL Debug",
+      "program": "${workspaceFolder}/DXMainClient/bin/Debug/Ares/UniversalGL/net7.0/clientdx.dll",
```
Rans4ckeR, line 9: Fix filename for each build `dx`, `ogl`, `xna`.

.vscode/tasks.json
```diff
@@ -0,0 +1,389 @@
...
+      "args": [
+        "build",
+        "${workspaceFolder}/DXMainClient/DXMainClient.csproj",
+        "--configuration=Release",
+        "--framework=net7.0-windows",
+        "--property:GAME=YR",
+        "--property:ENGINE=WindowsXNA",
+        "--property:GenerateFullPaths=true",
+        "--consoleloggerparameters:NoSummary"
+      ],
```
Rans4ckeR, line 385: XNA builds crash on launch because platform `x86` and 32bit dotnet.exe are not set. Also see https://github.com/CnCNet/xna-cncnet-client/blob/develop/Directory.Build.targets.

Rans4ckeR (changes requested):

.vscode/launch.json
```diff
@@ -0,0 +1,84 @@
...
+      "id": "Suffix",
+      "description": "The Suffix to run for.",
+      "type": "pickString",
+      "default": "dx",
+      "options": [
+        "dx",
+        "ogl",
+        "xna"
+      ]
+    }
```
Metadorius, line 82: I am not sure what's the point of those three variables, especially the last one. What's the point of building one thing and running another? For others - they should be tied to selected `Engine` IMO, though those are not as critical as the last one.

Metadorius (changes requested):

.vscode/launch.json
frg2089: > I am not sure what's the point of those three variables, especially the last one. What's the point of building one thing and running another? For others - they should be tied to selected `Engine` IMO, though those are not as critical as the last one.

Is there any way to use different values depending on the variable?

Metadorius: > Is there any way to use different values depending on the variable?

I am not sure there's a built in way, but you could feed it into an inline console script that matches the correct target, I guess.

PR #409, opened 2022-12-11
Finalize localization system

Metadorius:
- Implement parsing of all dynamic elements (except map names, idk if needed, I personally think not)
- Implement custom parser that overrides base tag parsing with localizable parsing for all text and location/size (needed because different languages may have different word sizes)
- Consolidate tooltip code into `IHasToolTip` (@Rampastring maybe we should just provide tooltips for every control?)

TODO
- ~~Write documentation, doc comments and make sure the code looks nice~~
- Fixup / add tooltips everywhere to use the new interface / move it up the hierarchy; idk if in scope of that PR?
- ~~Sort out translation file postfixes/locale codes;~~ used folders instead
- ~~Add asset location parsing corresponding to locale code in order to load localized images (some buttons use text embedded into images as buttons)~~
- ~~Ensure `BasedOn` inheritance works with translations properly (currently doesn't work for selectively loading stuff from INI);~~ implemented overriding manually
- ~~Implement invalidation of map cache after switching languages;~~ not needed I think?
- ~~Ensure loading translation file from theme dir works;~~
- ~~Add file swapping based on locales (or come up with a mechanism to do so);~~
- ~~Add locale swapping;~~
- ~~Add locale detection.~~


ClientCore/ClientCore.csproj
```diff
@@ -46,8 +46,8 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
-    <PackageReference Include="Rampastring.XNAUI.$(Engine)" Version="2.2.9" Condition="'!$(Configuration.Contains(Debug))'" />
-    <PackageReference Include="Rampastring.XNAUI.$(Engine).Debug" Version="2.2.9" Condition="'$(Configuration.Contains(Debug))'" />
+    <PackageReference Include="Rampastring.XNAUI.$(Engine)" Version="1.0.0" Condition="'!$(Configuration.Contains(Debug))'" />
+    <PackageReference Include="Rampastring.XNAUI.$(Engine).Debug" Version="1.0.0" Condition="'$(Configuration.Contains(Debug))'" />
```
Metadorius, line 50: Temporary change for local development purposes.

ClientCore/I18N/Locale.cs
```diff
@@ -0,0 +1,269 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using System.Reflection.Metadata;
+using Rampastring.Tools;
+using Rampastring.XNAUI.XNAControls;
+
+namespace ClientCore.I18N;
```
SadPencil, line 9: I18N is different from L10N

ClientCore/I18N/Locale.cs
```diff
@@ -0,0 +1,269 @@
...
+
+public class Locale : ICloneable
```
SadPencil, line 11: Locale means a specific language, region, formatting settings.
It does not stand for a translation file for a locale.
e.g. en-US is a locale, while a text written in en-US is not a locale

ClientCore/ProgramConstants.cs
```diff
@@ -42,6 +43,19 @@ public static class ProgramConstants
...
+        /// </summary>
+        public const string HARDCODED_LOCALE_CODE = "en";
+
+        /// <summary>
+        /// Used to denote <see cref="Environment.NewLine"/> in the INI files.
+        /// </summary>
+        /// <remarks>
+        /// Historically Westwood used '@' for this purpose, so we keep it for compatibility.
+        /// </remarks>
+        public const string INI_NEWLINE_PATTERN = "@";
```
SadPencil, line 57: How to support displaying email addresses?

SadPencil (changes requested):

ClientCore/ProgramConstants.cs
Metadorius: I am sticking to client inherited westwood's norms, you can't display an email address in INI files either which is fine IMO.

ClientCore/I18N/Locale.cs
Metadorius: It's not only for text, I store size/position INI values there too (and assets too). Locale is short and concise name for that.

ClientCore/I18N/Locale.cs
Metadorius: Intentional.

ClientCore/ProgramConstants.cs
Metadorius: If very needed, can add support for it via something like `\@`.

ClientCore/I18N/Locale.cs
SadPencil: Localization: Providing multiple translations.
Internalization: Always use UTC time, english text for logs, etc.

Totally different

Metadorius: Wikipedia:
> Internationalization is the process of designing a software application so that it can be adapted to various languages and regions without engineering changes. Localization is the process of adapting internationalized software for a specific region or language by translating text and adding locale-specific components.

> Localization (which is potentially performed multiple times, for different locales) uses the infrastructure or flexibility provided by internationalization (which is ideally performed only once before localization, or as an integral part of ongoing development).[[3]](https://en.wikipedia.org/wiki/Internationalization_and_localization#cite_note-3)

ClientCore/I18N/Locale.cs
SadPencil: I mean, in the whole world, there is only one `locale` identified as en-US.
This word is only used for specifiying language, region, formatting settings. Not for any instances that written in a language.
Don't misuse it, especially for INI key names and file names.
Locale assets, or Localization should be fine.

Metadorius: > Locale assets

Redundancy...

>  Localization

That is a process.

I fail to see a problem in calling locale-specific stuff just locale, it identifies what the class is supposed to contain and do just fine.

SadPencil: > > Locale assets
> 
> Redundancy...
> 
> > Localization
> 
> That is a process.
> 
> I fail to see a problem in calling locale-specific stuff class just `Locale`, it identifies what the class is supposed to contain and do just fine.

Example: https://developer.playcanvas.com/api/pc.I18n.html
where 'localization data' and 'asset' are used.
`this.app.i18n.addData()` loads a specific translation and other stuffs

ClientCore/I18N/Locale.cs
SadPencil: You are right. I18N is fine.

ClientCore/I18N/Locale.cs
Metadorius: On the second thought what you suggested as LocaleAssets is already kinda is what I have, the `Locale` has `Values` (name pending, that is pretty much your assets, data or w/e). That is pretty much equivalent to `Assets`. So, `LocaleAssets.Assets`?..

SadPencil:
I have an idea to define `\\`, `\@`, `\;` and `\n` (both `@` and `\n` can be used to represent a newline) so that the translation table can write any texts without constraints.

----------------------

The follwing code is provided by chatgpt. Unfortunately chatgpt is in a high demand and I have not ask him for an encoding code.
```
Can you help me write a C# code to decode a backslash-escaped string

Certainly! Here's a simple function that takes a string as input and returns a decoded version of the string by replacing any backslash-escaped characters with their unescaped counterparts:

string DecodeString(string input)
{
    // Use a StringBuilder to build the output string
    StringBuilder output = new StringBuilder();

    // Keep track of the current position in the input string
    int pos = 0;

    // Loop through the input string
    while (pos < input.Length)
    {
        // If the current character is a backslash, process the escape sequence
        if (input[pos] == '\\')
        {
            // Get the next character in the string
            pos++;
            if (pos >= input.Length)
            {
                // If there are no more characters, the string is invalid
                throw new ArgumentException("Invalid escape sequence at end of string");
            }

            // Process the escaped character
            switch (input[pos])
            {
                case '\\':
                    output.Append('\\');
                    break;
                case '"':
                    output.Append('"');
                    break;
                case '\'':
                    output.Append('\'');
                    break;
                case 'n':
                    output.Append('\n');
                    break;
                case 'r':
                    output.Append('\r');
                    break;
                case 't':
                    output.Append('\t');
                    break;
                default:
                    // If the character is not a recognized escape sequence, just append it to the output
                    output.Append('\\');
                    output.Append(input[pos]);
                    break;
            }
        }
        else
        {
            // If the character is not a backslash, just append it to the output
            output.Append(input[pos]);
        }

        // Move on to the next character
        pos++;
    }

    // Return the decoded string
    return output.ToString();
}

To use this function, you can pass a string with backslash-escaped characters as the argument, and it will return the decoded version of the string. For example:

string input = "This is a\\tbackslash-escaped\\nstring";
string output = DecodeString(input);
Console.WriteLine(output);

This will output the following string:

This is a   backslash-escaped
string

I hope this helps! Let me know if you have any questions.
```

ClientCore/I18N/Locale.cs
SadPencil: The `Locale` class Contains `Values` or `Assets`, which means the class is not a locale but a collection of locale assets. It is strange that a locale class has a property named locale or locale code which refers to the real locale.
`LocaleAssets.Values` or `LocaleAssets.Items` for translation items and `LocaleAssets.Locale` for accessing the locale that it refers to. It is pretty fine.

ClientCore/ProgramConstants.cs
SadPencil: Replied at https://github.com/CnCNet/xna-cncnet-client/pull/409#issuecomment-1356167513

SadPencil:
> The `\;` won't work because INI syntax has no idea of that. Also why bother the ChatGPT if we can just use the built in string methods... (I also have CoPilot so eh 😛 )

Oh right... something like `\59` or `\semicolon` for `;`?

ClientCore/I18N/Locale.cs
Metadorius: > a locale class has a property named locale or locale code which refers to the real locale.

Fair point, I'll change that to just `Code`.

> The Locale class Contains Values or Assets, which means the class is not a locale but a collection of locale assets.

That I disagree with. The class is an entity, and each instance of that entity contains some data specific to it.

Metadorius:
> Oh right... something like `\59` or `\semicolon` for `;`?

Maybe. Will have to think about it.

ClientCore/I18N/Locale.cs
SadPencil: > That I disagree with. The class is an entity, and each instance of that entity contains some data specific/related to it.

Disagreed. Imagine a class named `Language` but it does not refers to `English` but all kinds of stuffs written in English. This is messy.

The class name should matches its content instead of an entity. For example, each `Map` has its `GameMode`, so we have two classes `Map` and `GameMode`. You can expect 'Map.GameMode' but not `GameMode.Maps`. When designing `GameMode`, there is no such thing as a map, while designing `Map`, one can assume `GameMode` is availble as `GameMode` is naturally something that should belongs to a `Map`. Then `Map` classes depends on `GameMode` but NOT vice versa, so one can store `GameMode` inside a `Map`. In some programming languages like Go, circular dependencies are prohibited. 

Now think the relationship between `Locale` and `Asset`. A `Locale` is defined directly depending on language, region, formats, etc., and it is like a tag (something like a `GameMode`). `Asset` on the otherhand, natually has the language and format, i.e. each `Asset` has its `Locale`.  Therefore, `Asset` (or `Assets`) classes depends on `Locale` but NOT vice versa. In this case, even if we can have an one-to-many relationship between `Locale` and `Asset`, it is still not a good way to store `Assets` in `Locale`. Such circular dependencies brings mess and ambiguity.

SadPencil: Besides, think how Qt handles the relationships of each concept:

a `QTranslator` object can load a `Translation`
a `QTranslator` object can automatically load a Translation according to the user's preferred UI display `Locale`

The relationship among `QTranslator`, `Translation`, and `Locale` is very clear. `Locale` is just a tag.

Even if the word `Translation` is mainly for textes instead of UI locations, it does have a clear relationship and the concept of `Locale` is not messed up. Therefore we should use `TranslationAssets` or `LocaleAssets` for the class name.

Metadorius:
Implemented everything remaining. Also made the client read `\@` as `@` and `\semicolon` as `;` from multiline INI strings.

Rans4ckeR:
Don't forget to also check the build scripts by running `Build-All.ps1 Debug` & `Build-All.ps1 Release`. AfterPublish.targets will still try to copy Localization.dll & Localization.pdb.

Metadorius:
> Don't forget to also check the build scripts by running `Build-All.ps1 Debug` & `Build-All.ps1 Release`. AfterPublish.targets will still try to copy Localization.dll & Localization.pdb.

Done, thanks for reminding.

devo1929 (approved):

ClientCore/I18N/Translation.cs
```diff
@@ -0,0 +1,302 @@
...
+    }
+
+    /// <summary>Shows the information about the author.</summary>
+    public string Author { get; private set; } = string.Empty;
+
+    /// <summary>
+    /// The key stands for a key name, and the value stands for a string that is used in System.string.Format().
+    /// The value can not contain IniNewLinePattern when loading or saving via ini format.
+    /// </summary>
+    private Dictionary<string, string> Values { get; } = new();
```
SadPencil, line 37: Comment here needs updates

ClientCore/I18N/Translation.cs
```diff
@@ -0,0 +1,302 @@
...
+    public const string INI_PREFIX = "INI";
+    /// <summary>Used for INI values.</summary>
+    public const string CONTROLS_PREFIX = "Controls";
+    /// <summary>Used for parent-agnostic INI values.</summary>
+    public const string GLOBAL_PREFIX = "Global";
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="Translation"/> class.
+    /// </summary>
+    public Translation(string translationCode)
```
SadPencil, line 59: `string localeCode` (to be consistent with other overloading functions)

ClientCore/ProgramConstants.cs
```diff
@@ -1,10 +1,11 @@
-﻿using Localization;
+﻿using ClientCore.Extensions;
 using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Text;
 using System.Reflection;
 using Rampastring.Tools;
+using ClientCore.Extensions;
```
SadPencil, line 8: Duplicate `using ClientCore.Extensions;`

DXMainClient/Domain/Mission.cs
```diff
@@ -1,4 +1,7 @@
-﻿using Rampastring.Tools;
+﻿using ClientCore;
+using ClientCore.Extensions;
+using ClientCore.Extensions;
```
SadPencil, line 3: Duplicates

DXMainClient/Domain/Multiplayer/Map.cs
```diff
@@ -13,6 +13,8 @@
 using Point = Microsoft.Xna.Framework.Point;
 using Utilities = Rampastring.Tools.Utilities;
 using static System.Collections.Specialized.BitVector32;
+using ClientCore.Extensions;
+using ClientCore.Extensions;
```
SadPencil, line 17: Duplicates

SadPencil (changes requested):

SadPencil:
@Rans4ckeR Could you please help me understand which version of `Rampastring.Tools` will be included after deleting `Localization` project? See #393

ClientCore/I18N/Translation.cs
```diff
@@ -0,0 +1,302 @@
...
+    /// that loads the translation from an ini file.
+    /// </summary>
+    /// <param name="ini">An ini file to be read.</param>
+    /// <param name="localeCode">A locale code for this translation.</param>
+    public Translation(IniFile ini, string localeCode)
+    {
+        LocaleCode = localeCode;
+
+        if (ini == null)
+            throw new ArgumentNullException(nameof(ini));
```
SadPencil, line 78: Dropping legacy .NET 4 support enabling us to use `ArgumentNullException.ThrowIfNull(ini)` to shorten the code

Rans4ckeR:
> @Rans4ckeR Could you please help me understand which version of `Rampastring.Tools` will be included after deleting `Localization` project? See #393

XNAUI pulls in Tools:
image

SadPencil:
@Rans4ckeR Take an example. Suspose XNAUI requires `>=2.0.4` and the client requires `>=2.1.0`. How to explicitly specify this constraint if non of the dependencies has this requirement.

Rans4ckeR:
> @Rans4ckeR Take an example. Suppose XNAUI requires `>=2.0.4` and the client requires `>=2.1.0`. How to explicitly specify this constraint if non of the dependencies has this requirement.

In that case you need to again add a Tools 2.1.0 PackageReference directly to the client (ClientCore for example).

Metadorius:
Updated.

SadPencil:
All `Environment.NewLine` should be replaced with `\n` as the former is a runtime thing and the compiler can not deal with it. Working on it.

DXMainClient/PreStartup.cs
```diff
@@ -146,7 +146,7 @@ public static void Initialize(StartupParams parameters)
 
                     Logger.Log("Translation stub generation feature is now enabled. The stub file will be written when the client exits.");
 
-                    // Load all literal strings
+                    // Lookup all comptime available strings
```
SadPencil, line 149: typo

SadPencil (changes requested):

DXMainClient/PreStartup.cs
Metadorius: Where?

SadPencil: compile-time

SadPencil:
Apart from the minor revision, there is one last thing to be discussed: should we load the CultureInfo as specified in the translation file? (Add a property in the translation file referring to the CultureInfo code)

DXMainClient/PreStartup.cs
Metadorius: comptime is a common shortening for that

Metadorius:
> Apart from the minor revision https://github.com/Metadorius/xna-cncnet-client/pull/3, there is one last thing to be discussed: should we load the CultureInfo in the client as specified in the translation file? (Add a property in the translation file referring to the CultureInfo code)

No, we should not, this is handled by the folder name. Multiple reasons for that:
1. Loading it from the INI means more file reading, while incremental it still adds to the pile, also easier to just enumerate folders.
2. CultureInfo codes and translations should be in a dictionary like relation, so that for each unique CultureInfo there should be exactly 0 or 1 translations. Making it an INI key breaks that behaviour and opens the door to issues of having multiple translations with the same CultureInfo. Why should one deal with such?

SadPencil:
> > Apart from the minor revision [Metadorius#3](https://github.com/Metadorius/xna-cncnet-client/pull/3), there is one last thing to be discussed: should we load the CultureInfo in the client as specified in the translation file? (Add a property in the translation file referring to the CultureInfo code)
> 
> No, we should not, this is handled by the folder name. Multiple reasons for that:
> 
>     1. Loading it from the INI means more file reading, while incremental it still adds to the pile, also easier to just enumerate folders.
> 
>     2. CultureInfo codes and translations should be in a dictionary like relation, so that for each unique CultureInfo there should be exactly 0 or 1 translations. Making it an INI key breaks that behaviour and opens the door to issues of having multiple translations with the same CultureInfo. Why should one deal with such?

We are talking about different things. I mean changing `CultureInfo.CurrentCulture` and `CultureInfo.CurrentUICulture`.
The locale code is used to identify a translation file while the cultureinfo code can be used to set up these two fields.

SadPencil:
> Making it an INI key breaks that behaviour and opens the door to issues of having multiple translations with the same CultureInfo. Why should one deal with such?

Although this is another topic, sometimes there is more than one translations for a lanuage and sometimes there can be a fanny and funny translation alongside with a normal one. 

In a special case, in China there is a need to replace every word of "China" so that one can broadcasting their game without being censored by gov, so there are actually two zh-CN translations, one of them is the normal one and the other meets the censorship requirement.

Metadorius:
Ah, I see. I don't think there's much point though? What benefits does that bring?

SadPencil:
> Ah, I see. I don't think there's much point though? What benefits does that bring?

The format of numbers, date & times, and (perhaps, not validated) exception message texts as one need to use English exception text to report the bug to modders.

SadPencil:
> > Making it an INI key breaks that behaviour and opens the door to issues of having multiple translations with the same CultureInfo. Why should one deal with such?
> 
> Although this is another topic, sometimes there is more than one translations for a lanuage and sometimes there can be a fanny and funny translation alongside with a normal one.
> 
> In a special case, in China there is a need to replace every word of "China" with a faked country name so that one can broadcasting their game without being censored by gov (ridiculous but it is), so there are actually two zh-CN translations, one of which is the normal one and the other meets the censorship requirement.

I realized this topic is also important and it is better to key a translation file using an ID text to support: 
- `zh-CN`
- `zh-CN/censorship` # otherwise users have to manually replace the files
- `zh-CN/fancy`

Metadorius:
> > Making it an INI key breaks that behaviour and opens the door to issues of having multiple translations with the same CultureInfo. Why should one deal with such?
> 
> Although this is another topic, sometimes there is more than one translations for a lanuage and sometimes there can be a fanny and funny translation alongside with a normal one. 
> 
> In a special case, in China there is a need to replace every word of "China" with a faked country name so that one can broadcasting their game without being censored by gov (ridiculous but it is), so there are actually two zh-CN translations, one of which is the normal one and the other meets the censorship requirement.

That all can be done without the need for specifying CultureInfo to match to. Maybe for overriding the CultureInfo of the thread, but not for matching, as neither should not be picked automatically.

SadPencil:
> > > Making it an INI key breaks that behaviour and opens the door to issues of having multiple translations with the same CultureInfo. Why should one deal with such?
> > 
> > 
> > Although this is another topic, sometimes there is more than one translations for a lanuage and sometimes there can be a fanny and funny translation alongside with a normal one.
> > In a special case, in China there is a need to replace every word of "China" with a faked country name so that one can broadcasting their game without being censored by gov (ridiculous but it is), so there are actually two zh-CN translations, one of which is the normal one and the other meets the censorship requirement.
> 
> That all can be done without the need for specifying CultureInfo to match to. Maybe for overriding the CultureInfo of the thread, but not for matching, as neither should not be picked automatically.

Yes. This is two topics.

- Whether to allow two translations for a single locale code.
- Whether to load CultureInfo after loading a translation file.

Metadorius:
> I realized this topic is also important and it is better to key a translation file using an ID text to support:
> 
> * `zh-CN`
> * `zh-CN/censorship` # otherwise users have to manually replace the files
> * `zh-CN/fancy`

You can just have different translations for that + `BasedOn` (assuming it takes relative paths). And the censored translation should be the default one IMO.

Metadorius:
I mean for safety reasons.

SadPencil:
> > Ah, I see. I don't think there's much point though? What benefits does that bring?
> 
> 
> 
> The format of numbers, date & times, and (perhaps, not validated) exception message texts as one need to use English exception text to report the bug to modders.

Now there is only one topic remaining. Since the locale code does not have to be a recognizable string, to support loading the CultureInfo an additional field named CultureInfo should be added (to specify the CultureInfo)

TranslationNotifierGenerator/TranslationNotifierGenerator.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
Rans4ckeR, line 4: Any reason for not targetting .NET7?

Rans4ckeR (changes requested):

TranslationNotifierGenerator/TranslationNotifierGenerator.csproj
SadPencil: netstandard2.0 is required by source generator

DTAConfig/OptionPanels/DisplayOptionsPanel.cs
```diff
@@ -388,12 +409,12 @@ public void PostInit()
             {
                 string defaultGame = ClientConfiguration.Instance.LocalGame;
 
-                var messageBox = XNAMessageBox.ShowYesNoDialog(WindowManager, "New Compatibility Fix".L10N("UI:DTAConfig:TSFixTitle"),
-                    string.Format("A performance-enhancing compatibility fix for modern Windows versions" + Environment.NewLine +
-                        "has been included in this version of {0}. Enabling it requires" + Environment.NewLine +
-                        "administrative priveleges. Would you like to install the compatibility fix?" + Environment.NewLine + Environment.NewLine +
+                var messageBox = XNAMessageBox.ShowYesNoDialog(WindowManager, "New Compatibility Fix".L10N("Client:DTAConfig:TSFixTitle"),
+                    string.Format("A performance-enhancing compatibility fix for modern Windows versions\n" +
```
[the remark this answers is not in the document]
SadPencil: `Environment.NewLine` can not be determined at compile time and the use of `\n` should not bring issues on windows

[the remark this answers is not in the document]
Metadorius: Not a compile time constant => can't be caught by the analyzer (unless you want to write a lot more code) as a comptime string literal => we change everything to `\n`. I don't think there would be any issues w/ that.

ClientCore/ClientCore.csproj
```diff
@@ -46,11 +46,12 @@
...
-    <PackageReference Include="Rampastring.XNAUI.$(Engine)" Version="2.2.10" Condition="'!$(Configuration.Contains(Debug))'" />
-    <PackageReference Include="Rampastring.XNAUI.$(Engine).Debug" Version="2.2.10" Condition="'$(Configuration.Contains(Debug))'" />
+    <PackageReference Include="Rampastring.Tools" Version="2.0.4" />
+    <PackageReference Include="Rampastring.XNAUI.$(Engine)" Version="2.3.0" Condition="'!$(Configuration.Contains(Debug))'" />
+    <PackageReference Include="Rampastring.XNAUI.$(Engine).Debug" Version="2.3.0" Condition="'$(Configuration.Contains(Debug))'" />
     <PackageReference Include="System.Text.Encoding.CodePages" Version="7.0.0" />
   </ItemGroup>
   <ItemGroup>
-    <ProjectReference Include="..\Localization\Localization.csproj" />
+    <ProjectReference Include="..\TranslationNotifierGenerator\TranslationNotifierGenerator.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
```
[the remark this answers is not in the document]
SadPencil: There is a tight coupling among the TranslationNotifierGenerator, the L10N things, and the cncnet client.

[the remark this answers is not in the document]
Metadorius: I don't see any big reason to. If anything I'd just override props per-project if possible and use a `Release` config everywhere.

Metadorius:
After giving it some more thought I don't see any benefit in touching `CultureInfo`. We pick the *language*, not the culture for dates etc. `CurrentCulture` is already set by the system, and `CurrentUICulture` affects only the UI language of the current thread, which we already manage by ourselves, so no point in changing it.

ClientCore/ClientCore.csproj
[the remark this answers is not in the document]
Metadorius: Resolving as I don't see having excess configs as a big issue, can fix later.

ClientCore/Extensions/StringExtensions.cs
```diff
@@ -1,25 +1,77 @@
...
+    /// <returns>The translated string value.</returns>
+    /// <remarks>
+    /// This method is referenced by <c>TranslationNotifierGenerator</c> in order to check if the const
+    /// values that are not initialized on client start automatically are missing (via notification
+    /// mechanism implemented down the call chain). Do not change the signature or move the method out
+    /// of the namespace it's currently defined in. If you do - you have to also edit the generator
+    /// source code to match.
+    /// </remarks>
+    public static string L10N(this string defaultValue, string key)
+        => Translation.Instance.LookUp(key, defaultValue);
```
Rans4ckeR, line 76:
```suggestion
        => Translation.Instance?.LookUp(key, defaultValue) ?? defaultValue;
```

Prevents an additional exception when an exception occurs on startup before init has run.

Metadorius: I am pretty sure we should prefer failsafe over failfast here. It shouldn't be called before initing a translation.

Rans4ckeR: It shouldn't but it can, PreStartup.HandleException() calls it. There are currently also L10N() calls in the code that runs before the init e.g.
`throw new NotImplementedException("-NOAUDIO is currently not implemented, please run the client without it.".L10N("Client:Main:NoAudio"));`

Metadorius: In my opinion would be better to remove those instead.

SadPencil: I prefer `Instance = new Translation()` instead of using `??`

SadPencil:
> After giving it some more thought I don't see any benefit in touching `CultureInfo`. We pick the _language_, not the culture for dates etc. `CurrentCulture` is already set by the system, and `CurrentUICulture` affects only the UI language of the current thread, which we already manage by ourselves, so no point in changing it.

Then I don't have further comments.

SadPencil (approved):

SadPencil:
It could be better to provide an option to dismiss empty value when generating translation stub.
e.g.:
```ini
; INI:Maps:MapsMO\Standard\actionreaction:Briefing=
; INI:Maps:MapsMO\Standard\advancedwarfare:Briefing=
; INI:Maps:MapsMO\Standard\advancement:Briefing=
; INI:Maps:MapsMO\Standard\alaskanoutback:Briefing=
```
These lines are annoying

Metadorius:
@SadPencil should we ignore them at all (don't translate) or just don't notify the translation about them? I think an empty value should always be empty, so we could just bail early in `L10N` function.

Metadorius:
- Fonts are now loaded as assets, thanks Rampa
- Localized and handled localizing for game loading lobbies, statistics window, game preview box
- Enabled searching by untranslated names where applicable
- Added a tooltip which shows the original map name in game lobby
- Added context menu actions to copy map name and original map name where applicable
- Added support for culture overriding via corresponding INI tag
- Languages are now displayed in currently selected language
- Added support for translation of map names, custom components, updater mirrors, window title, game options dropdowns, settings dropdowns localization, themes
- Handled writing untranslated names in stats, discord presence, game broadcast message
- Improved tooltips code, added `FollowCursor` option
- Updated docs

ClientCore/Extensions/StringExtensions.cs
SadPencil: https://github.com/Metadorius/xna-cncnet-client/pull/9

How to re-open this conversation?

TranslationNotifierGenerator/TranslationNotifierGenerator.cs
```diff
@@ -0,0 +1,153 @@
+﻿using System;
+//// uncomment to debug the generator
```
Rampastring, line 2: You **_really_** wanted to write a comment there, huh?

Rampastring (changes requested):
Amazing work! There's some things to fix, mostly stylistical nitpicks.

On the `CNCNET0001` warnings, they seem to be disabled in quite many places. If it needs to be disabled so often, is there any benefit to having that warning in the first place?

ClientCore/ClientConfiguration.cs
```diff
@@ -188,7 +192,10 @@ public void RefreshSettings()
 
         public string[] RecommendedResolutions => clientDefinitionsIni.GetStringValue(SETTINGS, "RecommendedResolutions", "1280x720,2560x1440,3840x2160").Split(',');
 
-        public string WindowTitle => clientDefinitionsIni.GetStringValue(SETTINGS, "WindowTitle", string.Empty);
+#pragma warning disable CNCNET0001 // L10N Failure
```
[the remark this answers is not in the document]
SadPencil: > Amazing work! There's some things to fix, mostly stylistical nitpicks.
> 
> On the `CNCNET0001` warnings, they seem to be disabled in quite many places. If it needs to be disabled so often, is there any benefit to having that warning in the first place?

Most `L10N` calls are for literal strings. When compiling the code, source generator will check and enforce the following condition:
- The string is a literal string. Its value can be determined at compile-time. (to prevent improper `L10N` calls by someone who is not familiar to the localization method)
- The string is not trimmble.
- If the key is defined more than once, the values must be the same.

If we don't explicitly suppress `CNCNET0001`, we have to define another call that does the same thing as `L10N` with a different method name.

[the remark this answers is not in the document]
Rampastring: So currently `CNCNET0001` has to be suppressed for all non-literal strings? That sounds rather impractical to me.

I think we should rather either disable the warning in such a case (which currently is a completely normal use case of L10N, as you can see from the large bunch of disabled `CNCNET0001` warnings) or introduce another method for non-literal strings.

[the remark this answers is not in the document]
SadPencil: > So currently `CNCNET0001` has to be suppressed for all non-literal strings? That sounds rather impractical to me.
> 
> I think we should rather either disable the warning in such a case (which currently is a completely normal use case of L10N, as is evident from the large bunch of disabled `CNCNET0001` warnings) or introduce another method for non-literal strings.

The former is not acceptable as there would be no way to prevent less-experienced developers from writing something like `$“The output is {text}”.L10N("key-name")` or `"string.Format("whatever {0}", val).L10N("key-name")"`.

The latter is the only choice in order to dismiss the suppression comment. `L10N_Dynamic()` or something.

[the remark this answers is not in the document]
Rampastring: I think code reviews could catch the clear majority of improper `L10N` use. But how you solve it is up to you and @Metadorius. As-is I don't think it's acceptable.

ClientCore/Extensions/StringExtensions.cs
```diff
@@ -1,25 +1,80 @@
...
+    /// <param name="notify">Whether to add this key and value to the list of missing key-values.</param>
+    /// <returns>The translated string value.</returns>
+    /// <remarks>
+    /// This method is referenced by <c>TranslationNotifierGenerator</c> in order to check if the const
+    /// values that are not initialized on client start automatically are missing (via notification
+    /// mechanism implemented down the call chain). Do not change the signature or move the method out
+    /// of the namespace it's currently defined in. If you do - you have to also edit the generator
+    /// source code to match.
+    /// </remarks>
+    public static string L10N(this string defaultValue, string key, bool notify = true)
```
SadPencil, line 76: I just realized that we need to avoid optional parameters in public method of a library.
Because the DLL file must be compiled together with whatever program includes it.
Including an already compiled DLL file with optional arguments will cause crashes.
I have not tested it but this "feature" should exist in even .NET 7 as it is by design.

https://learn.microsoft.com/en-us/visualstudio/code-quality/ca1026
https://github.com/dotnet/roslyn-analyzers/issues/626
https://github.com/SonarSource/sonar-dotnet/issues/3622

ClientCore/ClientConfiguration.cs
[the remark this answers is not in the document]
SadPencil: @Metadorius https://github.com/Metadorius/xna-cncnet-client/pull/10

ClientCore/Extensions/StringExtensions.cs
SadPencil: https://github.com/Metadorius/xna-cncnet-client/pull/10

ClientCore/Settings/UserINISettings.cs
```diff
@@ -126,13 +129,20 @@ protected UserINISettings(IniFile iniFile)
 
         public event EventHandler SettingsSaved;
 
-        /*********/
-        /* VIDEO */
-        /*********/
+        /***********/
+        /* DISPLAY */
```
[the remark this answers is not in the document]
Metadorius: IIRC I thought it corresponded to tab name and wondered why exactly it's named not like what's in GUI.

TranslationNotifierGenerator/TranslationNotifierGenerator.cs
Metadorius: Hah I just `Ctrl+/` another line :P

ClientCore/Extensions/StringExtensions.cs
Metadorius: I don't understand how is this an issue though?

ClientCore/ClientConfiguration.cs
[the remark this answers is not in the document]
Metadorius: @Rampastring what are your thoughts on just changing the warning level to a suggestion?

DTAConfig/OptionPanels/ComponentsPanel.cs
```diff
@@ -36,35 +36,39 @@ public override void Initialize()
...
+                    Text = buttonText,
+                    Tag = c
+                };
                 btn.LeftClick += Btn_LeftClick;
 
-                XNALabel lbl = new XNALabel(WindowManager);
-                lbl.Name = "lbl" + c.ININame;
-                lbl.ClientRectangle = new Rectangle(12, btn.Y + 2, 0, 0);
-                lbl.Text = c.GUIName;
+#pragma warning disable CNCNET0001 // L10N Failure
```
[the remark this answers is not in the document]
Metadorius: Because `$"INI:CustomComponents:{c.ININame}:UIName"` is not a literal hence why the warning will appear otherwsie.

DXMainClient/Domain/Multiplayer/CnCNet/HostedCnCNetGame.cs
```diff
@@ -33,9 +33,9 @@ public class HostedCnCNetGame : GenericHostedGame
...
         public override int Ping => TunnelServer.PingInMs;
 
-        public override bool Equals(GenericHostedGame other) => 
-            other is HostedCnCNetGame hostedCnCNetGame ? 
-                string.Equals(hostedCnCNetGame.ChannelName, ChannelName, StringComparison.InvariantCultureIgnoreCase) :
-                base.Equals(other);
+        public override bool Equals(GenericHostedGame other)
+            => other is HostedCnCNetGame hostedCnCNetGame
+                ? string.Equals(hostedCnCNetGame.ChannelName, ChannelName, StringComparison.InvariantCultureIgnoreCase)
+                : base.Equals(other);
```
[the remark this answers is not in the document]
Metadorius: I have a (hopefully pretty convincing) reasoning for this style/change:
1) having expression body arrow on the line after the function prototype makes the function prototype not changed in case of change to/from expression body to traditional method definition, reducing merge conflicts in the long run;
2) because we read English and source code from left to right, having operators on the beginning of the line allows to see what's written way more easy and fast than having them at the end of the line, especially when the line is long as we can see in this case (plus, the beginnings of the lines are aligned, so it's even easier to see what's going on compared to staggered endings of the lines).

TranslationNotifierGenerator/TranslationNotifierGenerator.cs
Metadorius: Though I am kinda unsure, my point with it was that someone may accidentally capture the comment when selecting multiple lines with a mouse and uncomment what shouldn't be uncommented, so that was kinda a safeguard against that.

ClientCore/Extensions/StringExtensions.cs
Metadorius: As in we're not going to consume the library elsewhere, everything here always gets compiled in one sweep.

ClientCore/ClientConfiguration.cs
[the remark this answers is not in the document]
SadPencil: I disagree. The point is to remind people. If the message is a `info` and is not suppressed, developers will ignore more messages as these messages are already shown before they make changes.

ClientCore/Extensions/StringExtensions.cs
SadPencil: > As in we're not going to consume the library elsewhere, everything here always gets compiled in one sweep.

If the two projects are coupled so tight, they should be merged into one whole project. `ClientCore` is a library.

Metadorius: I absolutely disagree with this reasoning. Projects being split indicates logical separation of concerns. `ClientCore` implements core logic regardless of the UI that is used to display it, hence why it's a separate project.

ClientCore/ClientConfiguration.cs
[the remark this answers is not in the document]
Metadorius: That's a developer problem that they ignore it.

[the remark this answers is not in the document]
SadPencil: > That's a developer problem if they ignore it.

How can it be the developer's problem when the code already has CNCNET0001 messages? Does it make sense to let the developer pay attention to how many time CNCNET0001 messages show, and only if the number is greater than the original count means they have a problem? Ridiculous to me.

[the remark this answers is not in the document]
Metadorius: Well firstly your point wasn't very understandable in your previous message.

Secondly, I think you overestimate the importance of this aspect. If a person implements some new translated string, the message will notify them of the behavior. If they ignore it carelessly - that's their problem.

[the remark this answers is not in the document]
Metadorius: More so, code analysis doesn't run on the whole solution by default, so it will mostly display only active document messages.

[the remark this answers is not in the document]
SadPencil: > Secondly, I think you overestimate the importance of this aspect. If a person implements some new translated string, the message will notify them of the behavior. If they ignore it carelessly - that's their problem.

Good codes express itself. A message is by default ignorable if messages of the same type are prompted in the original codes and no one suppresses them. Leaving the CNCNET0001 messages means we have to write docs to explain in which case a message should be ignored or not ignored. Developers have also to read this doc before making any code changes. Is it worth than making a `L10N_Dynamic` method?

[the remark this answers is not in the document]
Metadorius: Developers just have to read the *message* that appears, docs are not necessary in this case. Introducing another method that does an exact same thing just for the analyzer to shut up is an even more weird solution, and I am opposed to that solution.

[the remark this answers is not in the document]
Rampastring: After considering this for a while, I've come to the conclusion that I support @Metadorius' view on this one. It seems unlikely that developers would mistakenly use non-literal strings with L10N when unnecessary, and when it happens, it'll most likely be caught on code reviews.

So I'd say either change `CNCNET0001` to info level or remove it completely.

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
```diff
@@ -1135,7 +1134,7 @@ private void ApplyGameOptions(string sender, string message)
                 if (!success)
                 {
                     AddNotice(("Failed to parse drop down options sent by game host (2)! " +
-                        "The game host's game version might be different from yours.").L10N("UI:Main:HostGameOptionInvalidTheSecondTime"), Color.Red);
+                        "The game host's game version might be different from yours.").L10N("Client:Main:HostGameOptionInvalidTheSecondTime"), Color.Red);
```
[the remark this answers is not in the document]
Metadorius: IDK, maybe the `(2)` confused SadPencil there (TBH it confuses me too).

ClientGUI/XNAClientButton.cs
```diff
@@ -1,15 +1,29 @@
...
+        {
+            get => Initialized ? ToolTip?.Text : _initialToolTipText;
+            set
+            {
+                if (Initialized)
+                    ToolTip.Text = value;
+                else
+                    _initialToolTipText = value;
+            }
+        }
```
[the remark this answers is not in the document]
Metadorius: Trying this (at 4AM so take with a grain of salt) and I didn't quite like it, it doesn't solve the repetition problem fully but also exposes what shouldn't be exposed, so I am leaving this as is for now.

ClientCore/Settings/UserINISettings.cs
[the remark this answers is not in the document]
Metadorius: Also, I don't think this holds true anymore, everything in it is over the place and not necessarily in video, so my rename makes even more sense (also considering other categories):
image
image
image

Metadorius:
TODO for myself: make MultiColumnListBox adjustable, review the source generator/analyzer warnings.

ClientCore/Settings/UserINISettings.cs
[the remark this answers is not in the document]
Rampastring: It'd be better to organize them properly into INI sections then, or just keep them unchanged. User interface is not a thing at ClientCore level so using tab names there is wrong.

ClientCore/ClientConfiguration.cs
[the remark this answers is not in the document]
SadPencil: I am still against changing `CNCNET0001` to info. I made changes to the source generator so that it would try to recognize INI strings for the best and suppress warnings in such a situation. For other messages those are warnings. https://github.com/Metadorius/xna-cncnet-client/pull/10  @Metadorius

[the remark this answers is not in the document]
Metadorius: @SadPencil first things first, I checked how you use that `CNCNET0001` diagnostic and ugh, you should absolutely not use one diagnostic for *four* different issues. Make a new diagnostic per each kind of an issue with an appropriate diagnostic level depending on the issue. Second, I'll review it later, but I really don't understand the fixation on this. A more helpful thing to do would be implementing localization of XNAMultiColumnListBox column widths and parameters.

TranslationNotifierGenerator/TranslationNotifierGenerator.cs
```diff
@@ -65,51 +51,68 @@ public void Execute(GeneratorExecutionContext context)
...
                         continue;
                     }
 
+                    // Avoid trimmable strings
                     if (valueText.Trim() != valueText)
                     {
-                        Warn($"The value of key {keyName} should not have leading or trailing white spaces.", context, l10nSyntax);
+                        context.ReportDiagnostic(Diagnostic.Create(new DiagnosticDescriptor(
+                            "CNCNET0004", "Trimmable translation value",
+                            $"The value of key {keyName} should not have leading or trailing white spaces.",
```
Rampastring, line 114: `white spaces -> white-space`

Rampastring (changes requested):

ClientCore/Settings/UserINISettings.cs
```diff
@@ -126,13 +129,20 @@ protected UserINISettings(IniFile iniFile)
...
+        /***********/
```
Metadorius, line 134:
```suggestion
        /*********/
        /* VIDEO */
        /*********/
```

TranslationNotifierGenerator/TranslationNotifierGenerator.cs
```diff
@@ -0,0 +1,154 @@
...
+
+                        continue;
+                    }
+
+                    // Avoid trimmable strings
+                    if (valueText.Trim() != valueText)
+                    {
+                        context.ReportDiagnostic(Diagnostic.Create(new DiagnosticDescriptor(
+                            "CNCNET0004", "Trimmable translation value",
+                            $"The value of key {keyName} should not have leading or trailing white spaces.",
```
Metadorius, line 114:
```suggestion
                            $"The value of key {keyName} should not have leading or trailing whitespace.",
```

Rampastring (approved):
Seems like it needs a rebase. Finally, it looks good otherwise.

Metadorius:
> Seems like it needs a rebase. Finally, it looks good otherwise, assuming you've found no issues in testing.

Done. Few people have done tests and we didn't find any more outstanding issues.

PR #425, opened 2023-01-05
RA2/YR Intro Movie Crash Fix 

ayylmaoRotE:
Adds the [Intro] and Play=no section to RA2MD.ini to prevent an error that occurs only when a fresh copy of the client goes on a game install that has yet to play the intro movie, causing the game to never open

### DIFF
--- a/ClientCore/Settings/UserINISettings.cs
+++ b/ClientCore/Settings/UserINISettings.cs
@@ -11,6 +11,7 @@ namespace ClientCore
         private static UserINISettings _instance;
 
         public const string VIDEO = "Video";
+        public const string INTRO_SETTINGS = "Intro";
         public const string MULTIPLAYER = "MultiPlayer";
         public const string OPTIONS = "Options";
         public const string AUDIO = "Audio";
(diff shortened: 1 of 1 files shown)


Rampastring:
Optimally this should be fixed in the spawners.

Metadorius:
I don't think there's a whole ton of spawner bugs though?

But just out of interest, @Belonit do you have this issue fixed already in your spawner reimpl?

Belonit:
> I don't think there's a whole ton of spawner bugs though?
> 
> @Belonit do you have this issue fixed already in your spawner reimpl?

Yes.  `yrpp-spawner` does not play intro moves

SadPencil:
> > I don't think there's a whole ton of spawner bugs though?
> > @Belonit do you have this issue fixed already in your spawner reimpl?
> 
> Yes. `yrpp-spawner` does not play intro moves

i.e., this issue will not exist on the latest spawner, right? If so we could just close this issue with a "won't fix"

Belonit:
> > > I don't think there's a whole ton of spawner bugs though?
> > > @Belonit do you have this issue fixed already in your spawner reimpl?
> > 
> > Yes. `yrpp-spawner` does not play intro moves
> 
> i.e., this issue will not exist on the latest spawner, right? If so we could just close this issue with a "won't fix"

Yes, that's right.  But this only applies to `yrpp-spawner`, I don’t know what happens in other games with a other spawner

PR #429, opened 2023-01-22
Fix typo in registries

SadPencil:
`HKML` -> `HKLM`


Rampastring:
Why are these in HKLM instead of HKCU like the others?

SadPencil:
efdb3f323ca7d1a8cf74da4bc63b583fd6548f9c
8444a0e3cee0725e7dd29d7d098e96df79740ab4

@Starkku @ayylmaoRotE Can the registry root of these two mods be `HKCU` instead of an invalid `HKML` (should be `HKLM`) that does not refer to anything?

SadPencil:
Well, no matter `HKCU` or `HKML`, the current fix should be merged to get rid of the wrong registry entry name.

PR #436, opened 2023-01-28
Make gitversion conditionally included with .git folder

devo1929:


Rans4ckeR (changes requested):
This still has problems unfortunately. I would prefer removing the NuGet package and adding the necessary steps in the workflow.

Try this for example in a new folder:

`git clone -b conditional-gitversion --single-branch https://github.com/devo1929/xna-cncnet-client.git`

When building the solution:

> System.InvalidOperationException: Gitversion could not determine which branch to treat as the development branch (default is 'develop') nor release-able branch (default is 'main' or 'master'), either locally or remotely. Ensure the local clone and checkout match the requirements or considering using 'GitVersion Dynamic Repositories'

SadPencil:
I prefer tell developers always get codes via git clone

devo1929:
> I would prefer removing the NuGet package and adding the necessary steps in the workflow.

The NuGet package is what automatically embeds the version in the assembly.

devo1929:
> I prefer tell developers always get codes via git clone

I tend to agree. Maybe instead of making the package conditional on the existence of a .git folder, it throws a warning or even error message when the .git folder doesn't exist when attempting to build.

PR #440, opened 2023-02-12
399-the-main-menu-music-track-can-be-heard-after-the-client-closes

### LINKED ISSUE #399: The Main Menu music track can be heard after the client closes.
The Main Menu music track can be heard after the client closes.

- The music track plays for a variable amount of seconds after the client window closes.
- At roughly 1/3 of the time, the client binary process lingers in the background when closed, until the full menu track has been finished.

Both these issues can be observed if the client is run, and then ordered to close before the main menu shows, or a couple of seconds after the main menu shows.

Reproduced consistently on Windows 10.

Rans4ckeR:
Fixes #399
Also fixes the 2nd stage updater unable to update the music file on linux as it complains about the file still being in use even after the client has exited.

### DIFF
--- a/DXMainClient/DXGUI/Generic/MainMenu.cs
+++ b/DXMainClient/DXGUI/Generic/MainMenu.cs
@@ -997,6 +997,7 @@ namespace DTAClient.DXGUI.Generic
         {
             Logger.Log("Exiting.");
             WindowManager.CloseGame();
+            themeSong?.Dispose();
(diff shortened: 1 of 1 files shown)


Rampastring:
There should be a `MediaPlayer.Stop()` call to stop music playback before the track is disposed. I'd expect just disposing the song while it's playing might lead into an exception.

PR #445, opened 2023-02-17
Fix *nix map PreviewPath

Rans4ckeR:


devo1929 (approved):

SadPencil:
Just letting you know that `FileInfo.ToString()` and `DirectoryInfo.ToString()` do NOT guarantee returning full path and therefore cause crashes sometimes. Will be fixed in the revert branch.

PR #483, opened 2023-08-26
INI migration guide

MortonPL:
This PR extends the migration guide with a step-by-step example INI configuration migration. It's aimed at YR modders, but most steps are applicable to any game using this client.

### DIFF
--- a/Docs/Migration-INI.md
+++ b/Docs/Migration-INI.md
@@ -30,29 +30,11 @@ RequiredFiles=
 ForbiddenFiles=
 ```
 
-## Edit `SkirmishLobby.ini`
-
-Skirmish game lobby used to be based on multiplayer game lobby. Now, skirmish and multiplayer lobbies share a new base layout instead.
-Map list (`lbMapList`) has been moved to the base layout. Game speed dropdown (`cmbGameSpeedCapSkirmish`) has been moved here from `GameOptions.ini`.
-
-1. Change `[INISystem]->BasedOn` from `MultiplayerGameLobby.ini` to `GameLobbyBase.ini`
-2. Add `$BaseSection=GameLobbyBase` to `[SkirmishLobby]`
-3. Remove `[lbMapList]` section
-4. Add these two sections:
-
-```ini
-[GameOptionsPanel]
-$CC-GODD03=cmbGameSpeedCapSkirmish:GameLobbyDropDown
-
-[cmbGameSpeedCapSkirmish]
-$BaseSection=cmbGameSpeedCap
-Items=Fastest (MAX),Faster (60 FPS),Fast (30 FPS),Medium (20 FPS),Slow (15 FPS),Slower (12 FPS),Slowest (10 FPS)
-```
-
 ## Add `GameLobbyBase.ini`
 
+Skirmish game lobby used to be based on multiplayer game lobby. Now, skirmish and multiplayer lobbies share a new base layout instead.
 This file is the base layout of all game lobbies (skirmish, LAN, CnCNet). **Game options have been moved from `GameOptions.ini` to this file**.
-See example configuration below.
+See example configuration below or in [YR mod base][mod_base].
 
 <details>
 <summary>Click to show file content</summary>
@@ -147,7 +129,7 @@ $BaseSection=ddPlayerStartBase
 [lbMapList]
 $X=LOBBY_EMPTY_SPACE_SIDES
 $Y=EMPTY_SPACE_TOP + 33
-$Width=getWidth($ParentControl) - (getX($Self) + (getWidth(MapPreviewBox) + LOBBY_EMPTY_SPACE_SIDES + LOBBY_PANEL_SPACING)
+$Width=getWidth($ParentControl) - (getX($Self) + (getWidth(MapPreviewBox) + LOBBY_EMPTY_SPACE_SIDES + LOBBY_PANEL_SPACING))
 $Height=getBottom(MapPreviewBox) - getY($Self)
 SolidColorBackgroundTexture=0,0,0,128
 
@@ -555,7 +537,7 @@ DistanceFromBottomBorder=-8
 
 If your game/mod has custom game options, you have to port them yourself. To add controls in the game options panel, add `$CC-GO` prefixed list entries in `[GameOptionsPanel]`, then create their own sections. In example:
 
-Before in `GameOptions.ini`:
+In `GameOptions.ini` in previous versions:
 
 ```ini
 [MultiplayerGameLobby]
@@ -572,7 +554,7 @@ Checked=False
 Location=1126,79
 ```
 
-After in `GameLobbyBase.ini`:
+In `GameLobbyBase.ini` in the new version:
 
 ```ini
 [GameOptionsPanel]
@@ -586,13 +568,29 @@ Checked=False
 Location=1126,79 ; $X and $Y are recommended instead
 ```
 
-## Edit `GameOptions.ini`
+## Edit `SkirmishLobby.ini`
 
-After adding all game lobby options to `GameLobbyBase.ini`, remove them here, as well as `SkirmishGameLobby` and `MultiplayerGameLobby` sections.
+This file extends the game lobby base with skirmish-specific controls. Map list (`lbMapList`) has been moved to the base layout. Game speed dropdown (`cmbGameSpeedCapSkirmish`) has been moved here from `GameOptions.ini`.
+If you have a modified `[SkirmishLobby]` section in `GameOptions.ini`, move it here.
+
+1. Change `[INISystem]->BasedOn` from `MultiplayerGameLobby.ini` to `GameLobbyBase.ini`
+2. Add `$BaseSection=GameLobbyBase` to `[SkirmishLobby]`
+3. Remove `[lbMapList]` section
+4. Add these two sections:
+
+```ini
+[GameOptionsPanel]
+$CC-GODD03=cmbGameSpeedCapSkirmish:GameLobbyDropDown
+
+[cmbGameSpeedCapSkirmish]
+$BaseSection=cmbGameSpeedCap
+Items=Fastest (MAX),Faster (60 FPS),Fast (30 FPS),Medium (20 FPS),Slow (15 FPS),Slower (12 FPS),Slowest (10 FPS)
+```
 
 ## Rename/Remove old `MultiplayerGameLobby.ini` and create new `MultiplayerGameLobby.ini`
 
 This file extends the game lobby base with multiplayer-specific controls, such as the chat box and lock and ready buttons.
+If you have a modified `[MultiplayerGameLobby]` section in `GameOptions.ini`, move it here instead of using the example one below.
 
 <details>
 <summary>Click to show file content</summary>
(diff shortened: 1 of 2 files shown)


Docs/Migration-INI.md
```diff
@@ -0,0 +1,854 @@
...
+`Resources/clientogl.bat`:
+```
+cd ..
+YRLauncher.exe -OGL
+```
+`Resources/clientxna.bat`:
+```
+cd ..
+YRLauncher.exe -XNA
+```
```
SadPencil, line 845: We have migrated the client to .NET 4.8/8.0. This should brings backward compatibility of these client binaries. 

The modders can get the latest launcher here: https://github.com/CnCNet/dta-mg-client-launcher/releases/tag/v2.0.7

Docs/Migration-INI.md
```diff
@@ -0,0 +1,854 @@
...
+
+### Add new assets
+
+Every file here can be either in `Resources` or in theme directories:
```
SadPencil, line 849: Could you leave a note about where to find these files? For example, perhaps in the mod base?

Docs/Migration-INI.md
```diff
@@ -0,0 +1,854 @@
...
+
+## Add `GameLobbyBase.ini`
+
+This file is the base layout of all game lobbies (skirmish, LAN, CnCNet). **Game options have been moved from `GameOptions.ini` to this file**.
+To add controls in the game options panel, add `$CC-GO` prefixed list entries in `[GameOptionsPanel]`, then create their own sections.
+See example configuration below.
+
+```ini
+[INISystem]
+BasedOn=GenericWindow.ini
```
Metadorius, line 45: This big code block asks to be spoilered.

Docs/Migration-INI.md
```diff
@@ -0,0 +1,854 @@
+# Migrating from older versions - INI configuration
+This guide uses [YR mod base](https://github.com/Starkku/cncnet-client-mod-base) configuration as an example. The majority of changes also applies to to non-YR client configurations.
```
SadPencil, line 2: Could you write the commit id of the mod base? Since you're refering to a historial state and this repo keeps updating

Docs/Migration-INI.md
Metadorius: Also can you add an example on how to port one example checkbox or dropdown? something like "My cool game option" before and after migration, with all the relevant code. I think that would be more useful than the example config because the example configs could be found in the new mod base and/or TSCv6 which we should mention instead.

Docs/Migration-INI.md
```diff
@@ -0,0 +1,854 @@
...
+It is **highly recommended** to make a complete backup of your mod before starting.
+
+## Edit `ClientDefinitions.ini`
+1. Add `[Settings]->UnixLauncherExe=Launcher.sh` (script file name can be anything)
+2. Create `Launcher.sh` in root directory:
+```sh
+#!/bin/sh
+
+cd "$(dirname "$0")"
+dotnet Resources/Binaries/UniversalGL/clientogl.dll "$@"
```
SadPencil, line 13: We have migrated the client to .NET 4.8/8.0. The folder structure is updated. .NET8 builds reside in `BinariesNET8` folder.

Docs/Migration-INI.md
```diff
@@ -0,0 +1,854 @@
...
+```
+3. Add these entries in `[Settings]` (fill with your required/forbidden mod files):
+```ini
+; Comma-separated list of files required to run the game / mod that are not included in the installation.
+RequiredFiles=
+; Comma-separated list of files that cannot be present to run the game / mod without problems.
+ForbiddenFiles=
+```
```
Metadorius, line 21: Tbh this is a new feature more than a migration requirement, maybe mark it as optional.

SadPencil (changes requested):

Docs/Migration-INI.md
```diff
@@ -0,0 +1,854 @@
...
+
+## Edit `SkirmishLobby.ini`
```
Metadorius, line 23: Could you add a small recap on what changes we are doing and why? Here especially but generally too.

Metadorius (changes requested):
We should update this to account for net48 revert and merge ASAP so that we could release the update for the masses.

Docs/Migration-INI.md
MortonPL: I have been told that .NET 8.0 version won't be shipped to users.

SadPencil: I will give a precise description: net 8 builds **are** shipped to users, unless modders **actively** remove the `BinariesNET8` folder. Of course we can discuss and adjust the default behavior but this is the current behavior

Docs/Migration-INI.md
```diff
@@ -0,0 +1,972 @@
...
+
+1. Add `[Settings]->UnixLauncherExe=Launcher.sh` (script file name can be
+   anything)
+2. Create `Launcher.sh` in root directory:
+
+```sh
+#!/bin/sh
+
+cd "$(dirname "$0")"
+dotnet Resources/Binaries/UniversalGL/clientogl.dll "$@"
```
SadPencil, line 24: universal gl is net8 exclusive

Docs/Migration-INI.md
```diff
@@ -0,0 +1,972 @@
...
+It is **highly recommended** to make a complete backup of your game/mod
+before starting.
+
+## Edit `ClientDefinitions.ini`
+
+The way the client is launched on Unix systems has changed.
+
+1. Add `[Settings]->UnixLauncherExe=Launcher.sh` (script file name can be
+   anything)
+2. Create `Launcher.sh` in root directory:
```
SadPencil, line 18: game directory, not root (typically means /root)

Docs/Migration-INI.md
```diff
@@ -0,0 +1,972 @@
...
+
+### Replace binary files
+
+1. Replace contents of `Resources/Binaries` and `Resources/Compatibility` with
+   new files. This directory contains the .NET Framework 4.8 version
+   of the client.
+2. **OPTIONAL** Copy contents of downloaded `BinariesNET8` into a new directory
+   `Resources/BinariesNET8`. This directory contains the .NET 8 version
+   of the client.
+3. Replace your launcher executable in root directory (whetever it's called,
```
SadPencil, line 951: same as above. don't use root directory, especially in linux case

SadPencil (changes requested):

SadPencil (changes requested):

Docs/Migration-INI.md
MortonPL: What build should Unix users use with .NET Framework then?

SadPencil: .net framework is for windows only. Advanced linux users may explore the possibility with Mono but this is definitely beyond the scope.

For most linux users they should use .net 8 client

Rampastring: AFAIK most Linux users currently run the client with .NET 4.0 and XNA through Wine. That should hopefully be possible with .NET 4.8 too.

Metadorius:
> Also can you add an example on how to port one example checkbox or dropdown? something like "My cool game option" before and after migration, with all the relevant code. I think that would be more useful than the example config because the example configs could be found in the new mod base and/or TSCv6 which we should mention instead.

@MortonPL anything on that?

MortonPL:
It's there, see paragraph "Port custom game options" (lines 543-580). There isn't really much to do, though.

Metadorius:
> It's there, see paragraph "Port custom game options" (lines 543-580). There isn't really much to do, though.

Ah the syntax highlighting broke so I didn't notice, pardon. Then I don't have much complaints, let's address the remaining feedback from Rampa and SadPencil and merge.

Docs/Migration-INI.md
```diff
@@ -0,0 +1,972 @@
...
+```
+
+## Appendix
+
+For completion's sake, below are additional steps required for a complete
+migration (beyond INI changes) to client version [2.11.0.0][client].
+
+### Replace binary files
+
+1. Replace contents of `Resources/Binaries` and `Resources/Compatibility` with
```
SadPencil, line 943: I believe `Resources/Compatibility` directory is not released with the client?

SadPencil, line 943: Also, apart from`Resources/Binaries` folder, `Resources/client{dx, ogl, xna}.exe` are also a part of .NET 4.8 client. 

I think we can use the following tree view to better illustrate the copy process? 
```txt
Compiled
└── Ares
    └── Resources # override the `Resources` folder to update the client binaries
        ├── Binaries  # this folder contains partial .NET 4.8 client files
        ├── BinariesNET8 # this folder contains .NET 8.0 client files, where modders can either delete it, or keep it for an experimental cross-platform support
        ├── clientdx.exe
        ├── clientdx.exe.config
        ├── clientdx.pdb # .pdb file contains debug symbols. It can be either deleted or retained.
        ├── clientogl.exe
        ├── clientogl.exe.config
        ├── clientogl.pdb
        ├── clientxna.exe
        ├── clientxna.exe.config
        └── clientxna.pdb
```

Docs/Migration-INI.md
```diff
@@ -0,0 +1,972 @@
...
+   new files. This directory contains the .NET Framework 4.8 version
+   of the client.
+2. Copy contents of downloaded `BinariesNET8` into a new directory
+   `Resources/BinariesNET8`. This directory contains the .NET 8 version
+   of the client.
+3. Replace your launcher executable in game directory (whetever it's called,
+   `YRLauncher.exe`, `MentalOmegaLauncher.exe`, etc.) with version
+   [2.0.7](https://github.com/CnCNet/dta-mg-client-launcher/releases/tag/v2.0.7).
+4. Replace the second-stage updater in `Resource/Updater` with version
```
SadPencil, line 952: I think we can stress that both second-stage updaters (.NET  4.8 and .NET 8.0) can be copied in `Resource/Updater`. 

The following tree view should be helpful to readers.

```txt
<game dir>/Resources/Updater
├── Rampastring.Tools.dll # common dependencies of .NET 4.8/.NET 8.0 second-stage updater
├── SecondStageUpdater.dll # .NET 8.0 second-stage updater
├── SecondStageUpdater.exe # .NET 4.8 second-stage updater
├── SecondStageUpdater.exe.config  # .NET 4.8 second-stage updater
└── SecondStageUpdater.runtimeconfig.json # .NET 8.0 second-stage updater
```

Docs/Migration-INI.md
```diff
@@ -938,17 +936,64 @@ Location=12,189
...
+└── clientxna.pdb        # same as above
+```
+
+### Update the client launcher
+
+The client launcher (that resides in the game directory) has been updated.
+You can replace the old one with version
+[2.0.7](https://github.com/CnCNet/dta-mg-client-launcher/releases/tag/v2.0.7).
+Remember to rename it from `CncNetLauncherStub.exe` to your launcher name, i.e.
+`YRLauncher.exe`, `MentalOmegaLauncher.exe`. Rename the `.config` file
```
Metadorius, line 973: > Rename the `.config` file appropriately, i.e. `YRLauncher.exe.config`, `MentalOmegaLauncher.exe.config`.

I think that file was optional, wasn't it?

Metadorius (commented):
Apart from manual line breaks and the comment below I see no issues, great job.

Metadorius (approved):

Docs/Migration-INI.md
SadPencil: Well, long long ago, the `.exe.config` file for the launcher is not optional, as deleting it causes an additional requirement of .NET 3.5 for Windows 8/10 users, while retaining this file only requires .NET 4.0. DTA preserves this file IIRC.

Some mods like Mental Omega omitted this file so users suffer from installing another .NET runtime. Some users re-distribute the mod with a manually written `.exe.config` file (it basically writes, either try .NET 3.5 first and use .NET 4.0 as a fallback, or try .NET 4.0 first and fallback to .NET 3.5) to overcome it. If we do not include this file, the `.exe` file may have a conflict with user's `.exe.config` file.

PR #484, opened 2023-08-30
Allow INItializableWindow buttons/checkboxes to toggle visibility of other controls

### LINKED ISSUE #775: Port toggleable controls from Starkku's XNA Client fork
### Description

Probably feature has been implemented at this commit https://github.com/Starkku/xna-cncnet-client/commit/dc6d2dac045dd4578add361fcb0ed0e85fecd2ed

## Behavior example

Map markers on:

Image

Map markers off:

Image

Starkku:
Closes #775 

- It is now possible to set `$Toggles` on `XNAButton` or `XNACheckBox` & derivatives that are children of `INItializableWindow`. What this does is make the button/checkbox toggle visibility of the listed controls. Main use case would be additional panels for game options to conserve space where needed.


FS-21:
Since it has almost 1 year old I merged latest sources and this PR in a local repository.

For testing this variable $ToggleableControl that is mentioned in the main message I have declared both controls in;
```ini
[GameLobbyBase]
...
$CC29=SuperWeaponsOptionsPanel:XNAExtraPanel ;PlayerExtraOptionsPanel
$CC30=btnSuperWeaponsOptionsOpen:XNAClientButton
```
Then:
```ini
[btnSuperWeaponsOptionsOpen]
...
$ToggleableControl=SuperWeaponsOptionsPanel

[SuperWeaponsOptionsPanel]
$Width=400
$Height=200
$X=0
$Y=0; - LABEL_SPACING - DEFAULT_LBL_HEIGHT + 100
SolidColorBackgroundTexture=0,0,0,128
DrawMode=Stretched
BackgroundTexture=50.png
```

And works very well :-D

new xna-client-feature-01

MahBoiDeveloper:
@Starkku merge https://github.com/Starkku/xna-cncnet-client/pull/9 to resolve merge conflict and continue development

DXMainClient/DXGUI/Multiplayer/PlayerExtraOptionsPanel.cs
```diff
@@ -162,42 +162,42 @@ public override void Initialize()
             Visible = false;
 
             var btnClose = new XNAClientButton(WindowManager);
-            btnClose.Name = nameof(btnClose);
+            btnClose.Name = nameof(PlayerExtraOptionsPanel) + "_" + nameof(btnClose);
```
Metadorius: for INI should convert to plain strings so that accidental renames won't break compatibility

ClientGUI/INItializableWindow.cs
```diff
@@ -216,6 +216,10 @@ static string Localize(XNAControl control, string attributeName, string defaultV
                     if (kvp.Value == "Disable")
                         control.LeftClick += (s, e) => Disable();
                 }
+                else if (kvp.Key == "$ToggleableControl" && control is XNAClientButton button)
+                {
+                    button.SetToggleableControl(kvp.Value);
+                }
```
Metadorius: Where is this supposed to be useful apart from extra panel? I am not sure if this is not a niche case. Currently it's used only in that one place I think. It is kind of non-standard UI. Are there any alternative considerations, like tabs or whatever else?

ClientGUI/XNAClientButton.cs
```diff
@@ -67,5 +67,35 @@ protected override void ParseControlINIAttribute(IniFile iniFile, string key, st
...
+            {
+                if (ToggleableControl.Enabled)
+                    ToggleableControl.Disable();
+                else
+                    ToggleableControl.Enable();
+            }
+
+            base.OnLeftClick();
+        }
+
```
Metadorius: Not sure if I am a fan of
  - 1 control only;
  - modifying the `XNAClientButton` with additional state and methods/overrides instead of just attaching an event handler outside of it that would toggle the needed things.

ClientGUI/INItializableWindow.cs
Metadorius: Also how about `$Toggles=controlName1,controlName2,...`?

Metadorius (changes requested):
Extra panel changes themselves are fine, some more questions on the rest.
Docs and migration are missing, you should fix that.

Docs/INISystem.md
Metadorius, line 1: prob should be reverted and instead a migration note added

DXMainClient/DXGUI/Multiplayer/PlayerExtraOptionsPanel.cs
Starkku: Prefix only or all of it? Client already uses `nameof` a lot for setting control `Name` which is used for INI parsing, but composite names are not really used so far.

ClientGUI/XNAClientButton.cs
Starkku: I will be refactoring it so it is handled in `INItializableWindow.ReadLateAttributesForControl()` with event handlers and accepts multiple controls per button.

ClientGUI/INItializableWindow.cs
Starkku: It could hypothetically be used to toggle visibility of game option controls e.g dropdowns/checkboxes directly, altho there may be some restrictions to this e.g only two view states per button (as it works as a binary toggle). As part of [further refactoring](https://github.com/CnCNet/xna-cncnet-client/pull/484#discussion_r2581542664) I can explicitly restrict it to `XNAExtraPanel`, `XNAPanel` and derivatives or leave the restrictions out entirely.

Syntax can be changed to your suggestions, it is even more appropriate if it is not restricted to just panels.

DXMainClient/DXGUI/Multiplayer/PlayerExtraOptionsPanel.cs
Metadorius: all of it I think, just so that refactors won't break the name. I agree it is counterintuitive but I feel in the long run it will be better

Starkku:
`PlayerExtraOptionsPanel` changes have been separated from this PR.

Other changes:
- Remove bespoke logic from XNAClientButton
- Allow any control to be toggled
- Allow checkboxes to toggle controls
- Allow one control to toggle multiple controls

ClientGUI/INItializableWindow.cs
```diff
@@ -237,15 +239,52 @@ private void ReadLateAttributesForControl(XNAControl control)
...
+                    continue;
+
+                // Handle buttons & checkboxes being able to toggle other controls.
+                if (child is XNAButton || child is XNACheckBox)
+                {
+                    string toggles = childSection.GetStringValue("$Toggles", null);
+
+                    if (!string.IsNullOrWhiteSpace(toggles))
+                    {
+                        var controlnames = toggles.Split(',', StringSplitOptions.RemoveEmptyEntries);
```
Metadorius, line 254: IIRC we have an extension method for cleaned up list splitting

ClientGUI/UIHelpers.cs
```diff
@@ -0,0 +1,57 @@
...
+{
+    /// <summary>
+    /// Finds a child control matching a name and optionally a type.
+    /// </summary>
+    /// <typeparam name="T">Type of the child control to find.</typeparam>
+    /// <param name="parent">Parent control.</param>
+    /// <param name="controlName">Name of the child control.</param>
+    /// <param name="recursive">Whether or not to look for children recursively.</param>
+    /// <returns>Child control matching the given name if found, otherwise type default value.</returns>
+    public static T FindMatchingChild<T>(XNAControl parent, string controlName, bool recursive = false)
```
Metadorius, line 19: perhaps better to use extensions and place this in an appropriately named file? both of those seem to be fitting as `XNAControl` extension methods

Metadorius (changes requested):

ClientGUI/UIHelpers.cs
SadPencil: e.g., in INItializableWindow we have defined FindChild and FindChildrenStartWith. Can follow that pattern and use an extension

ClientGUI/Settings/SettingCheckBoxBase.cs
```diff
@@ -46,7 +44,7 @@ public string ParentCheckBoxName
             set
             {
                 _parentCheckBoxName = value;
-                UpdateParentCheckBox(FindParentCheckBox());
+                UpdateParentCheckBox(UIHelpers.FindMatchingChild<XNAClientCheckBox>(Parent, _parentCheckBoxName, false));
```
SadPencil, line 47: I would leave FindParentCheckBox() method and write the find child statement into the FindParentCheckBox method body

ClientGUI/INItializableWindow.cs
MahBoiDeveloper: 598a2b387fcc43bddad2b3550fe8922c3d76be84

ClientGUI/Extensions/XNAControlExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+{
+    /// <summary>
+    /// Finds a child control matching a name and optionally a type.
+    /// </summary>
+    /// <typeparam name="T">Type of the child control to find.</typeparam>
+    /// <param name="parent">Parent control.</param>
+    /// <param name="controlName">Name of the child control.</param>
+    /// <param name="recursive">Whether or not to look for children recursively.</param>
+    /// <returns>Child control matching the given name if found, otherwise type default value.</returns>
+    public static T FindMatchingChild<T>(this XNAControl parent, string controlName, bool recursive = false)
```
SadPencil: use name `FindChild`? it matches our existing method name

SadPencil (changes requested):
I don't think the name `$Toggles` is clear -- its name can suggest it toggles a checkbox to be checked or unchecked. Needs to think another name, probably `$VisibilityToggles`

SadPencil:
Renamed to `$VisibilityToggles` in a525ca7cd51cc13daef8191c90cf18f567e651b5. Feel free to suggest a better name

ClientGUI/INItializableWindow.cs
```diff
@@ -266,15 +221,52 @@ private void ReadLateAttributesForControl(XNAControl control)
...
+                                        toggleControl.Visible = !toggleControl.Visible;
+                                    };
+                                }
+                                else
+                                {
+                                    child.LeftClick += (sender, args) =>
+                                    {
+                                        toggleControl.Enabled = !toggleControl.Enabled;
+                                        toggleControl.Visible = !toggleControl.Visible;
+                                    };
```
SadPencil:
```cs
                                    child.LeftClick += (sender, args) =>
                                    {
                                        toggleControl.Enabled = !toggleControl.Enabled;
                                        toggleControl.Visible = !toggleControl.Visible;
                                    };
```

Uhh... in this case the toggleControl's visibility is *binded* to the button. `child.LeftClick` does not observe `toggleControl.Enabled/Visible` changes so possible wrong event handler was listened on.

Same for `checkBox.CheckedChanged`.

Therefore, I think we can't merge this PR right now. We need to carefully think of its usage, and provide an example. Also, the name should probably be `BindsVisibilityTo` or `BindsVisibilityFrom` depends on the implementation.

SadPencil (changes requested):

SadPencil:
> Uhh... in this case the toggleControl's visibility is binded to the button. child.LeftClick does not observe toggleControl.Enabled/Visible changes so possible wrong event handler was listened on.
>
> Same for checkBox.CheckedChanged.
>
> Therefore, I think we can't merge this PR right now. We need to carefully think of its usage, and provide an example. Also, the name should probably be BindsVisibilityTo or BindsVisibilityFrom depends on the implementation.

For example, maybe we can change to bind, whether a checkbox is *checked* (or *unchecked*), the visibility of the target controls are changed. Like `$BindsVisibilityTo=controlName1, controlName2, !controlName3`

More discussions are welcomed.

PR #494, opened 2023-11-08
Revert to .NET Framework

Metadorius:
# Description

.NET 7 provides considerably worse user experience for average desktop user over .NET Framework and brings many issues for distribution .NET Framework 4.8 is still going to be supported for a good while and supports most modern tech from C# 11 because of .NET Standard 2.0 support and many backports, and because the changes are not tremendous to go back to .NET Framework it was decided to do so.

# The proposed way

Currently I suggest to take the current develop and with power of polyfills, backports and git reverts where there's runtime/API incompatibility (like default interface implementations) make the current version work on .NET Framework 4.8. The reason is there is *a lot* of changes since the revert and reapplying them one by one would be, in my opinion, a great headache with conflicts and such, and current code is pretty stable from our limited testing with @Starkku in Project Phantom closed beta.

# TODO

- [x] Decide if we keep UniversalGL/.NET 7 target for native Linux compatibility or ditch it completely in favor of simplicity
- [x] Polyfill where needed (or revert if not possible) the .NET 4.8 incompatible code from client
  - [x] Same thing in updater, see https://github.com/CnCNet/cncnet-client-updater/pull/17
  - [x] Same thing in launcher, see https://github.com/CnCNet/dta-mg-client-launcher/pull/13
- [x] Sort out the polyfills/backports to use ([some overview](https://www.meziantou.net/polyfills-in-dotnet-to-ease-multi-targeting.htm)), I initially wanted to use Polyfill but PolySharp seems to work better? Currently some projects use one or another
- [x] Revert to the old built client file structure (exes in resources, updater name) where needed/applicable to lessen the burden of upgrading for game/mod maintainers


SadPencil:
` Polyfill where needed (or revert if not possible) the .NET 4.8 incompatible code from client: Same thing in launcher, see [link pending]`

The launcher has not introduced any .NET 4.8 incompatible codes for now. Is introducing Polyfill necessary? See: https://github.com/CnCNet/dta-mg-client-launcher/pull/13

SadPencil:
Update progress. Now, the client can be successfully built in .NET 4.8.

Several commitments of mine requires additional check: @Metadorius 

- 3d9033db4d05ba4d096033de98248540f74c86d3 removes copying Newtonsoft.Json.Bson.dll. It seems this file is unused. Keeping this line causes a file-not-found error triggered by AfterPublish.targets
- 35e56f3990ff4ff1ff4ff7435f4cc0e01620f7ef temporarily replaces Polyfill v1.32.0 with a modified Polyfill available [here](https://github.com/SadPencil/Polyfill/commit/2b3959f7cf9124c044f5f7d94ffe2986deb1da31). This is just a workaround where the implementation is ugly. Further investigation should be proceed with. See [this comment](https://github.com/SimonCropp/Polyfill/issues/12#issuecomment-1802974383) for more details.

SadPencil:
`Decide if we keep UniversalGL/.NET 7 target for native Linux compatibility or ditch it completely in favor of simplicity`

In my opinion, .NET 7 target can be preserved, as `#if !NETFRAMEWORK` is only used for forth time.

SadPencil:
Now the client runs normally

SadPencil:
> Sort out the polyfills/backports to use ([some overview](https://www.meziantou.net/polyfills-in-dotnet-to-ease-multi-targeting.htm)), I initially wanted to use Polyfill but PolySharp seems to work better? Currently some projects use one or another

Definitely Polyfill. Switching to PolySharp brings additional compile errors, while Polyfill works. Although Polyfill has missing Index/Range issue but it should be an MSBuild issue, therefore making a workaround for it is acceptable.

SadPencil:
1b987c48b522bf3e98209d7c0e7b32502b79b2e5 Changing assembly loading priority is needed so that people could just override the binaries folder without deleting it first. Explain in example: `ClientCore.dll` used to be a common dll, but not anymore at the time when the client supported both .NET 6 and .NET 4.8. If someone upgraded his client without deleting "Binaries\ClientCore.dll" first, the client fails to launch. This commit changes the priority so even if "Binaries\ClientCore.dll" is not deleted.

Belonit:
All assemblies except UniversalGL work fine at a quick glance. Below are logs of an attempt to launch UniversalGL on Windows and Linux
[ClientCrashLog_linux.txt](https://github.com/CnCNet/xna-cncnet-client/files/13467879/ClientCrashLog_linux.txt)
[ClientCrashLog_win.txt](https://github.com/CnCNet/xna-cncnet-client/files/13467880/ClientCrashLog_win.txt)

SadPencil:
TODO: update CalculateHashes()

> All assemblies except UniversalGL work fine at a quick glance. Below are logs of an attempt to launch UniversalGL on Windows and Linux [ClientCrashLog_linux.txt](https://github.com/CnCNet/xna-cncnet-client/files/13467879/ClientCrashLog_linux.txt) [ClientCrashLog_win.txt](https://github.com/CnCNet/xna-cncnet-client/files/13467880/ClientCrashLog_win.txt)

Should be fixed in f6acb03

SadPencil:
TODO: update CalculateHashes()

SadPencil:
Thanks for the help. However, the latest commit a2bbf5a by @Rans4ckeR brings some changes which, in my opinion, requires further discussions:
- [ ] This commit removes GitVersion and on the other hand, relies on GitHub CI for the versioning. This brings an issue that apart from the CI, if modders compile the client via `Build-All.ps1` by they hand, the version would be `0.0.0` as indicated by this file. I think GitHub CI is a helper rather than a necessary component. 
- [ ] This commit reverts 063c319a19d39f3eab039548ed254b554a3e0673, which brings out a problem that the dependency DLL files of .net 4.8 and .net 8 are different, despite that they have common file names. In my opinion it is better to keep the `Binaries` folder with .net 4.8 client files and move .net 8 files to another folder, otherwise we might meet a DLL hell in the future. Example crash: https://github.com/CnCNet/xna-cncnet-client/pull/494#issuecomment-1826878121, which was fixed in f6acb036f75da6fe190e9fd81b2b0fb575346381
- [ ] This commit includes several `#if NETFRAMEWORK` which seems to be less necessary. In these cases, using the old .net framework codes should be just fine.
- [ ] This commit completely overrides 3db1443769758dfed3447d28d7815383254e655b. `new Startup().Execute()` is not in a try-catch statement again. Whether this commit is correct or not requires further tests.
- [ ] In this commit, `CnCNet.ClientUpdater` is made a NuGet package. Further discussion are needed.
- [ ] This commit brings an issue in FileHashCalculator. Note that *every player* **should** get the same hash value if their client build and modding assets are the same. `#if !NETFRAMEWORK` in `FileHashCalculator.cs` might breaks the rule and remains further discussion.
- [ ] In `PreStartup.cs` file, `ApplicationConfiguration.Initialize();` is called when `#if NETFRAMEWORK` does not hold. This is not accurate since `ApplicationConfiguration` was brought in .NET 6. `#if NET6_0_OR_GREATER` is overridden by this commit.
- [ ] `using System.Linq` brings a lot of handy extension methods. However, in `Program.cs` file, this line is moved into an `#if NETFRAMEWORK` section.
- [ ] This commit removes the HighDPI related things in manifest file. Although the `app.config` file specifies these settings, I need to stress that a great number of modders might think `.config` files are useless and therefore delete them! (Think about those mods that requires .NET 3.5 just because modders have deleted the `.exe.config` file alongside with the launcher executable, for example, Mental Omega) This is why we should preserving *both* the so-called "legacy" way and the new method.
- [ ] This commit reverts 6cb7e75, where the latter commit makes `Newtonsoft.Json.Bson.dll` file optional, because in .NET framework build this file does not exist at all. Further investigation are required. Also this is exactly the reason why I am strongly opposing mixing DLLs. 
- [ ] `GetOperatingSystemVersion()` is re-implemented. Someone needs to check whether it is correct.

SadPencil:
TODO:
- [ ] figure out which dlls should be shared; double check dll dependencies
- [ ] WindowsGL build breaks on my machine. figure out why
- [ ] make PRs to target rampa's two libraries for .NET 8
- [ ] sometimes the build script fails because of "Rampastring.XNAUI..Debug" package is missing. However, all parameters have been explicitly given. This issue disappears if the script is run for a second time. figure out why

SadPencil:
WindowsGL builds (both for .NET 4.8 and .NET 8) fails because of this commit: https://github.com/Rampastring/Rampastring.XNAUI/commit/f32bcda575ed45aefcfabc19f426e5bc7e76e707

SadPencil:
Upstream dependencies prerequisites:
- [ ] https://github.com/Rampastring/Rampastring.Tools/pull/11
- [ ] https://github.com/Rampastring/Rampastring.XNAUI/pull/29
- [ ] https://github.com/Rampastring/Rampastring.XNAUI/pull/30
- [ ] And also https://github.com/CnCNet/cncnet-client-updater/pull/17, of course

After these pull requests are merged and new packages are released, those "*.(s)nupkg" files acting as temporary files can be removed from this branch

SadPencil:
All works are done. We could either wait for upstream dependencies updates or just merge this PR before the dependencies updates (preserving those NuGet packages files in `References` directory). 

Ready for review.

.github/workflows/build.yml
```diff
@@ -14,20 +14,20 @@ jobs:
...
+    - uses: actions/checkout@v4
       with:
         fetch-depth: 0
 
     - name: Setup .NET Core SDK
-      uses: actions/setup-dotnet@v3.0.3
+      uses: actions/setup-dotnet@v4
       with:
-        dotnet-version: '7.x.x'
+        dotnet-version: '8.x.x'
```
alexlambson, line 24: Is this correct? I thought we were downgrading, or am I missing something

ClientCore/ClientConfiguration.cs
```diff
@@ -382,15 +381,53 @@ public IEnumerable<string> SupplementalMapFileExtensions
...
+                    return OSVersion.UNKNOWN;
+
+                if (osVersion.Major == 5)
+                    return OSVersion.WINXP;
+
+                if (osVersion.Major == 6 && osVersion.Minor == 0)
+                    return OSVersion.WINVISTA;
+
+                if (osVersion.Major == 6 && osVersion.Minor <= 1)
+                    return OSVersion.WIN7;
```
alexlambson, line 400: Best OS

ClientCore/ClientCore.csproj
```diff
@@ -4,7 +4,7 @@
     <Description>CnCNet Client Core Library</Description>
     <Company>CnCNet</Company>
     <Product>CnCNet Client</Product>
-    <Copyright>Copyright © CnCNet, Rampastring 2011-2022</Copyright>
+    <Copyright>Copyright © CnCNet, Rampastring 2011-2024</Copyright>
```
alexlambson, line 7: I wonder if there's a way to auto update this

ClientCore/ClientCore.csproj
```diff
@@ -50,9 +50,11 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
-    <PackageReference Include="Rampastring.Tools" Version="2.0.4" />
-    <PackageReference Include="Rampastring.XNAUI.$(Engine)" Version="2.3.8" Condition="'!$(Configuration.Contains(Debug))'" />
-    <PackageReference Include="Rampastring.XNAUI.$(Engine).Debug" Version="2.3.8" Condition="'$(Configuration.Contains(Debug))'" />
+    <!-- TODO: update version after this package targets .NET 8. -->
+    <PackageReference Include="Rampastring.Tools" Version="2.0.6-local-397fda5" />
+    <!-- TODO: update version after this package targets .NET 8. -->
```
alexlambson, line 55: Nit: Do we have a ticket or GitHub task anywhere for this? I generally don't like hanging todos, even ones this obvious.

DXMainClient/Program.cs
```diff
@@ -116,25 +129,71 @@ static void Main(string[] args)
...
+        private static string SearchResourcesDir(string startupPath)
+        {
+            DirectoryInfo currentDir = new(startupPath);
+            for (int i = 0; i < 3; i++)
+            {
+                // Determine if currentDir is the "Resources" folder
+                if (currentDir.Name.ToLowerInvariant() == "Resources".ToLowerInvariant())
+                    return currentDir.FullName;
+
+                // Additional check. This makes developers to debug the client inside Visual Studio a little bit easier.
```
alexlambson, line 187: Thank you

LICENSE.md
```diff
@@ -19,7 +19,7 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 This software includes Rampastring.Tools and Rampastring.XNAUI. Their license follows:
 
-Copyright (c) 2022 Rami "Rampastring" Pasanen
+Copyright (c) 2016-2023 Rami "Rampastring" Pasanen
```
alexlambson, line 22: 2024

.github/workflows/build.yml
alexlambson: Oh I see, we're working on v8 support. I missed that thread

alexlambson (approved):
I don't see anything I had an issue with. May be worth waiting for approval from someone with stronger C# experience if you want style feedback though.

.github/workflows/build.yml
SadPencil: Even if we downgrade the client, we still need the latest toolchain from .NET 8 to make use of latest features. And we are adding .NET 8 support as well as .NET 4.8.

LICENSE.md
SadPencil: No. It's 2023 unless Rampa upgrades his library

ClientCore/ClientCore.csproj
SadPencil: Well, there are nearly thirties TODOs in the codes. Codes with unfinished TODOs are better than say nothing. And some TODOs might remain unfinished forever, because of, for example, a rare usecase.

ClientCore/ClientConfiguration.cs
MortonPL: About Windows versions, what about Windows 11? Is it included in WIN810? If yes, IMO the name should be updated to WIN81011 (or something more visually reasonable).

SadPencil: `WIN810` includes Windows 11.

This question is not related to the .NET 4.8/8.0 so I'll just mark it as solved. Could write an issue about this.

README.md
```diff
@@ -1,79 +1,92 @@
-# CnCNet Client #
+
+# CnCNet Client
```
MortonPL, line 2: Why is there an empty line?

MortonPL (approved):

DXMainClient/Program.cs
```diff
@@ -116,25 +129,71 @@ static void Main(string[] args)
...
+
+            if (commonFileInfo.Exists)
+                return Assembly.Load(AssemblyName.GetAssemblyName(commonFileInfo.FullName));
 
             return null;
         }
 #endif
+
+        // SearchResourcesDir is copied from ClientCore
+        private static string SearchResourcesDir(string startupPath)
```
Metadorius, line 178: I think should add the explanation for this and a cref to the original method to a docstring so that if someone updates one of the methods the second would need to be updated too, and a comment in the second place backreferencing that one.

Metadorius (commented):
Outstanding work, although as the OG author of the PR I can't approve that. I have no complaints apart from the one I left.

PR #508, opened 2024-02-14
Check for "default" updateconfig.ini file to copy

devo1929:


GrantBartlett (approved):

Metadorius (changes requested):
It seems you mix the hardcoded update config check and config-defined checkbox. IMO this isn't a good design, you should make a proper hardcoded toggle for update configs then. which would not require the client restart either.

devo1929:
> It seems you mix the hardcoded update config check and config-defined checkbox. IMO this isn't a good design, you should make a proper hardcoded toggle for update configs then. which would not require the client restart either.

I'm not entirely sure what you're suggesting. What checkbox are you referring to?

Metadorius:
> I'm not entirely sure what you're suggesting. What checkbox are you referring to?

Err meant the dropdown for beta version.

devo1929:
Ok, so are you suggesting that upon startup, the client reads this value for the dropdown, finds the corresponding file to copy and then do the copy here at this init step?

Metadorius:
I am suggesting to redo the dropdown completely by hardcoding it as a selector for updateconfig (not even sure if it would need to be copied) which would handle whatever this PR handles (and also not requiring restart) instead of monkeypatching the client files via a hack which I initially came up with.

devo1929:
> I am suggesting to redo the dropdown completely by hardcoding it as a selector for updateconfig (not even sure if it would need to be copied) which would handle whatever this PR handles (and also not requiring restart) instead of monkeypatching the client files via a hack which I initially came up with.

The issue is not how the "beta" or "dev" specific `updateconfig.ini` files are copied into place.  If the updater is going to rely on a specific file to provide information on where to get updates, this file cannot exist in the repository which it currently does for YR. And, if we remove it from the repo, then we need some "default" `updateconfig.ini` file that is used so that the updater has something to use at all, the same way that _any_ setting inside the client is instantiated, coming from our "settings" INI: every setting has a default value for them.

The issue that we're trying to solve here is that when someone changes their `updateconfig.ini`, whether it be by copying another instance from somewhere into the base location of `updateconfig.ini`, or we write to it or something, the client is triggered to update from that new server location. Upon getting the update, they're getting a version of the `updateconfig.ini` file that does not have the same server location as they just had before the update. As a result, the client tells them that there's an update available again and as you pointed out, this happens on a restart, regardless of whether or not we make the restart necessary upon changing their update location. The beta user might restart for other reasons and we want to get rid of this unnecessary/unwanted/invalid alert.

The solution in this PR is simple: the `updateconfig.ini` file is _removed_ from repos entirely, thus never coming in with an update, thus never being overwritten by an update. However, we still need a way for games/mods to provide specific server locations for that game's updates. Putting in a "default" `updateconfig.default.ini` file accomplishes exactly that. The client checks for the existence of the `updateconfig.ini` file and if it does exist, do nothing. If it doesn't exist and the `updateconfig.default.ini` file exists, copy the default to the base file. It's really no different than when we implement a new feature/setting inside the client itself that obviously won't exist in an existing game installation "setting" INI files; we provide a default value in code and when the "settings" INI file is written back out, it will now have that value written to it with its own key.

PR #518, opened 2024-03-10
Implement custom mission support with CampaignTagSelector

SadPencil:
This PR includes and therefore closes #367 and closes #370, avoiding maintaining too many PRs

This PR works with https://github.com/CnCNet/yrpp-spawner/pull/19

## the campaign tag selector
This PR is backward compatible as by default the option is off. Modders can enable the `CampaignTagSelectorEnabled` option to use this feature.

This feature enables modders to make a "choose your side" interface before the campaign selector. Choosing among "Act 1/Act 2/..." or any other tags is also possible.

The screenshot below shows the feature. (I am not an expert in beautifying user interfaces so it is only a working example. Fully configurable through ini files.)

img1
img2
img3

The modders define tags for missions in `battle.ini` and customize the selector window in `CampaignTagSelector.ini` using the new INItializableWindow ini format.

I use TSC v6 client as an example. The corresponding files are attached here.

[TSC.v6.example.zip](https://github.com/CnCNet/xna-cncnet-client/files/9687387/TSC.v6.example.zip)

## the custom mission support

Apart from the campaign tag selector feature brought by #370, this PR also include a custom mission support for YR and Ares/Phobos.

Users can drop third-party fan-made singleplayer missions in `Maps\CustomMissions` folder, like this:

```
CustomMissions
├── vnsepa.csf # optional
├── vnsepa.map
├── vnsepa.pal # optional
└── vnsepa.shp # optional
```

This installs a custom mission named `vnsepa`.

`vnsepa.map` file is basically a `.map` file containing campaigns, with the following new-added sectiones:

```ini
[CNCNET:MISSION:BATTLE.INI]
CD=2
BuildOffAlly=yes
Scenario=vnsepa.map ; this line will be ignored by the client
Description=vnsepa
Summary=MAP:DESCREDDAWN
Side=0
SideName=Allies
Act=1
LongDescription=Operation: The Separation - Location: Kon Tum, Central Vietnam blah blah

[CNCNET:MISSION:MISSION.INI]
Briefing=Brief:AREDDAWN
ShowBriefing=true
UIName=MAP:TITREDDAWN
LSLoadMessage=LoadMsg:AREDDAWN
LSLoadBriefing=LoadBrief:AREDDAWN
LS640BriefLocX=20
LS640BriefLocY=20
LS800BriefLocX=20
LS800BriefLocY=20
LS640BkgdName=custommission.shp ; the filename is fixed
LS800BkgdName=custommission.shp ; the filename is fixed
LS800BkgdPal=custommission.pal ; the filename is fixed
LoadScreenText.Color=LightGrey

; Map created with FinalAlert 2(tm) Mission Editor
; Get it at http://www.westwood.com
; note that all comments were truncated
```

The two sections define mission info for the client as well as YR spawner.

To show custom missions, modders need to define a button `ButtonTag_CUSTOM` in file `CampaignTagSelector.ini`.


Starkku:
I took a liberty of adding some improvements to this PR.

- `CampaignTagSelector` <-> `CampaignSelector` no longer displays fade effect between them.
- `CampaignSelector` is now `INItializableWindow`. The default config has been updated accordingly.
- If `CampaignTagSelector` is enabled in `ClientDefinitions.ini`, `CampaignSelector` has a button available to return to it.
- Map file section names `CNCNET:MISSION:BATTLE.INI` and `CNCNET:MISSION:MISSION.INI` have been changed to `ClientMissionConfig` and `GameMissionConfig` respectively.
- The logic for reading the `GameMissionConfig` section written under scenario name in `spawn.ini` has been decoupled from custom mission logic for potential future proofing.
- Client forces scenario names in `spawn.ini` to all-caps because game makes assumptions about them being so.
- Added an option (`ReturnToMainMenuOnMissionLaunch`) to control whether or not client returns to main menu when launching a mission. Defaults to true.
- Generalized the user setting control logic in `CampaignSelector` and added new INI key for them (`ResetToDefaultOnGameExit`) that if enabled resets the control and its associated setting to default value every time game process exits.



There is currently no work done on any sort of campaign progression tracker or global variable toggles. They might be better saved for another PR anyway.

SadPencil:
图片

@Starkku Do you have a migration guide on `CampaignSelector`? Making it an `INItializableWindow` brings an incompatible change which modders may feel hard to follow. Consider that CampaignSelector is hardly modified, perhaps we can revert this change for a better compatibility, and delay the `INItializableWindow` migration to when the global variable things are migrated?

nikolyn10:
For the purposes of legacy support and ease of use, it may be worth making the CampaignSelector able to create a default set of controls if it doesn't find the ones expected.

SadPencil:
> For the purposes of legacy support and ease of use, it may be worth making the CampaignSelector able to create a default set of controls if it doesn't find the ones expected. 

good thought

@Starkku is that possible?

Starkku:
Yes but it partly defeats the purpose and other `INItializableWindow`s do not do this.

SadPencil:
If the migration is not introduced now, this will be just an optional feature that people don't have to do anything to upgrade to this version. Otherwise we'll have to release 2.13.

Since 2.12.4 was the only "safe" version or a non-buggy version since 2.12.0 (the buggy here refers to some new bugs introduced in 2.12.0, not other undiscovered bugs) so I want incoming 2.12.x versions a little bit longer to make people feel safe about upgrading the client

Starkku:
Reverted `INItializableWindow` change for `CampaignSelector`. Return to campaign tag selector button now requires explicit enabling via `Enabled=true` and `Visible=true` in INI even if campaign tag selector is enabled.

DXMainClient/DXGUI/Generic/CampaignSelector.cs
```diff
@@ -28,18 +32,23 @@ public class CampaignSelector : XNAWindow
...
         {
             this.discordHandler = discordHandler;
+            this.campaignTagSelector = campaignTagSelector;
         }
 
         private DiscordHandler discordHandler;
+        private CampaignTagSelector campaignTagSelector;
 
-        private List<Mission> Missions = new List<Mission>();
+        private List<Mission> lbCampaignListMissions = new List<Mission>();
```
Metadorius: this is not listbox, hence the `lb` prefix is incorrect

DXMainClient/Domain/CustomMissionHelper.cs
```diff
@@ -0,0 +1,52 @@
...
+            string missionFileName = mission.Scenario;
+            Debug.Assert(missionFileName.EndsWith(".map", StringComparison.InvariantCultureIgnoreCase), "Mission file should have the extension \".map\".");
+
+            // copy the CSF file if exists
+            foreach ((string ext, string filename) in new (string, string)[]
+            {
+                    ("csf", ClientConfiguration.Instance.CustomMissionCsfName),
+                    ("pal", ClientConfiguration.Instance.CustomMissionPalName),
+                    ("shp", ClientConfiguration.Instance.CustomMissionShpName),
+            })
```
Metadorius: how about making this customisable and not just csf, pal, shp? just need to ensure it doesn't override any client files accidentally

not sure if better done in this PR or left for another one

DXMainClient/Domain/CustomMissionHelper.cs
```diff
@@ -0,0 +1,52 @@
...
+            {
+                string sourceFileName = missionFileName[..^".map".Length] + "." + ext;
+                if (SafePath.GetFile(SafePath.CombineFilePath(ProgramConstants.GamePath, sourceFileName)).Exists)
+                    File.Copy(SafePath.CombineFilePath(ProgramConstants.GamePath, sourceFileName), SafePath.CombineFilePath(ProgramConstants.GamePath, filename));
```
Metadorius: we have linking now which would make sense if some additional stuff like custom assets is supplied, I think could use them here?

DXMainClient/DXGUI/Generic/CampaignSelector.cs
SadPencil: 82ac5130 Rename lbCampaignListMissions to selectedMissions

DXMainClient/Domain/CustomMissionHelper.cs
SadPencil: b051a1a738928a1745f90e1c5b5f7ea8a8bee559 Create hard links for supplemental mission files

DXMainClient/Domain/CustomMissionHelper.cs
SadPencil: 34a56750d9b0cf5e14e5e89eab86bb6beabeeffa Allow customized custom mission supplemental files

SadPencil:
> Error: D:\a\xna-cncnet-client\xna-cncnet-client\DXMainClient\Domain\CustomMissionHelper.cs(32,50): error CS0121: The call is ambiguous between the following methods or properties: 'string.Split(char[]?, StringSplitOptions)' and 'string.Split(string?, StringSplitOptions)' [D:\a\xna-cncnet-client\xna-cncnet-client\DXMainClient\DXMainClient.csproj::TargetFramework=net8.0]

didn't encounter this error on my machine. will workaround it anyway

Docs/INISystem.md
```diff
@@ -536,3 +536,18 @@ TrustedDomains=  ; comma-separated list of strings,
                  ; domain names to match links and prevent the message box from appearing before they open by default browser
                  ; example: cncnet.org,github.com,moddb.com
 ```
+
+```ini
+[Settings]
+CustomMissionPath=Maps/CustomMissions ; path to the folder containing fan-made maps
+CustomMissionSupplementDefinition=csf|stringtable99.csf|pal|custommission.pal|shp|custommission.shp ; supplement files that are supposed to be copied to the game folder when a custom mission is played
```
Metadorius, line 554: What is the reason to introduce new syntax like this? Can't you do the same with `,` as a separator or simple as separate keys?

SadPencil: This seems to be the only place where the list is processed as pairs. Therefore I want to use a different separator to notify modders. And | representing pairs is also used in Win32 OpenFileDialog

Starkku: The more confusing part is that the pairs are also separated from other pairs by |

The entire system also hinges on not defining more than one pair with same file extension. It cannot currently handle that.

SadPencil: > The more confusing part is that the pairs are also separated from other pairs by |

The "pairs are also separated from other pairs" matches the same convention  in https://learn.microsoft.com/en-us/dotnet/api/microsoft.win32.filedialog.filter 
e.g.,
```cs

dlg.Filter = "Word Documents|*.doc|Excel Worksheets|*.xls|PowerPoint Presentations|*.ppt" +
             "|Office Files|*.doc;*.xls;*.ppt" +
             "|All Files|*.*";
```

> The entire system also hinges on not defining more than one pair with same file extension. It cannot currently handle that.

This can be ignored for now, left as another PR. The TS resolution things require a special handling so we can leave it as a future work.

Metadorius: I don't think non-IT modders ever encountered that, so that diminishes the example. Why not define those all as separate keys?

SadPencil: > I don't think non-IT modders ever encountered that, so that diminishes the example. Why not define those all as separate keys?

Yeah, right. Regex is *friendly* to non-IT modders but "|" symbol is not.

Whatever. So what's the prefered way to express the setting? `csf,stringtable99.csf,pal,custommission.pal,shp,custommission.shp` or something?

Metadorius: > Yeah, right. Regex is _friendly_ to non-IT modders but "|" symbol is not.

I never claimed it is, regex is hostile to anyone, even IT people lol. In that case it provided flexbility though, in this case it doesn't provide flexibility.

I thought of something like

```ini
CustomMissionFile0Extension=csf
CustomMissionFile0CopyAs=stringtable99.csf
CustomMissionFile1Extension=...
```

SadPencil: > > Yeah, right. Regex is _friendly_ to non-IT modders but "|" symbol is not.
> 
> I never claimed it is, regex is hostile to anyone, even IT people lol. In that case it provided flexbility though, in this case it doesn't provide flexibility.
> 
> I thought of something like
> 
> ```ini
> CustomMissionFile0Extension=csf
> CustomMissionFile0CopyAs=stringtable99.csf
> CustomMissionFile1Extension=...
> ```

I would prefer the following, still making the setting in oneline

```
csf:stringtable99.csf, pal:custommission.pal
```

`:` or `=`

Metadorius: That is somewhat better because it's possible to add new parameters without breaking backwards compatibility, but I am still not a fan of cramming everything in one line, it's not like we're constrained on storage space.

DXMainClient/DXGUI/Campaign/CampaignSelector.cs
```diff
@@ -714,6 +729,58 @@ public void LoadMissionsWithFilter(ISet<string> selectedTags, bool disableCustom
...
+                dd.SelectedIndex = settingsIni.GetIntValue("GameOptions", dd.Name, dd.SelectedIndex);
+
+                if (dd.SelectedIndex > -1 && dd.SelectedIndex < dd.Items.Count)
+                    dd.SelectedIndex = dd.SelectedIndex;
+            }
+
+            foreach (CampaignCheckBox cb in CheckBoxes)
+                cb.Checked = settingsIni.GetBooleanValue("GameOptions", cb.Name, cb.Checked);
 
             lbCampaignList.TopIndex = 0;
```
SadPencil, line 744: `lbCampaignList.TopIndex = 0; ` belongs to the last function

DXMainClient/DXGUI/Campaign/CampaignSelector.cs
```diff
@@ -309,7 +295,7 @@ private void BtnReturn_LeftClick(object sender, EventArgs e)
 
         private void BtnLaunch_LeftClick(object sender, EventArgs e)
         {
-            userSettings.ForEach(c => c.Save());
+            SaveSettings();
```
SadPencil, line 212: `userSettings.ForEach(c => c.Save());` line was removed. 

`SaveSettings()` method is made for the campaign dropdowns and campaign checkboxes, which is different from `userSettings` here

DXMainClient/DXGUI/Campaign/CampaignSelector.cs
SadPencil: Note: incorrect code snippet. See https://github.com/CnCNet/xna-cncnet-client/pull/518#issuecomment-3522028223 instead

DXMainClient/DXGUI/Campaign/CampaignSelector.cs
SadPencil: Note: incorrect code snippet. See https://github.com/CnCNet/xna-cncnet-client/pull/518#issuecomment-3522028223 instead

SadPencil:
Comments on 9b7d212

图片
图片

ORDINALYRDIY:
I am a player using the cncnet client, I would like to add the function of "generalized mission selector user setting controls", and hope that the PR will be merged soon.

SadPencil:
Merged in https://github.com/CnCNet/xna-cncnet-client/pull/922

PR #525, opened 2024-03-12
Fix obtaining admin privileges

### LINKED ISSUE #458: Launcher no longer asks permission for admin, instead opens up notepad with an error
Steps to reproduce:

1. Run `CnCNetYRLauncher.exe` as non-admin. The client will crash and open up Notepad with the error. 
2. Run the launcher as admin, the crash will not appear and open the client normally.

The behaviour pre net7, was it would prompt the user to run as admin.

SadPencil:
Fixes #458

Now .NET 4.8 works but .NET 8 won't


SadPencil:
Is there a bug in GitHub? I already rebased this branch to d65d5f85 but this PR still refers to 1e21247

SadPencil:
What the hell? I'll just create a new PR and close this one. Weird.

PR #527, opened 2024-03-13
Fix privileges evaluation

SadPencil:
Fixes https://github.com/CnCNet/xna-cncnet-client/issues/458 in .NET 4.8.

This PR also moves DisplayErrorAction from ClientCore to DXMainClient because this is an GUI stuff.

For .NET 8.0:

- In Windows, there's always a black console window running dotnet.exe. This is because the window can not be hidden if ShellExecute is used, while ShellExecute is an requirement of verb runas. So in .NET 8.0 there will be no automatic evaluation because a black console window is more annoying. Users should run the launcher as admin instead.
- In non-Windows, since when this error happens, the client has no privilege to write text files on game folder (the logger cannot be initialized then), a text file xna-cncnet-client-error.txt is created in temp, in order to show error dialog.

Thus, .NET 8.0 clients are not recommended for regular Windows users

### DIFF
--- a/DXMainClient/Domain/MainClientConstants.cs
+++ b/DXMainClient/Domain/MainClientConstants.cs
@@ -28,7 +28,8 @@ namespace DTAClient.Domain
 
         public static OSVersion OSId = OSVersion.UNKNOWN;
 
-        public static bool LoggerInitialized = false;
+        // TODO: remove this variable after `Logger.Initialized` property is implemented by upstream
+        public static bool LoggerInitialized { get; set; } = false;
 
         private static Action<string, string, bool> displayErrorAction = null;
         /// <summary>
--- a/DXMainClient/Domain/MainClientConstants.cs
+++ b/DXMainClient/Domain/MainClientConstants.cs
@@ -37,7 +37,14 @@ namespace DTAClient.Domain
             set => displayErrorAction = value;
         }
 
-        public static Action<string, string, bool> InitialDisplayErrorAction = (title, error, exit) =>
+        /// <summary>
+        /// Show an error in console as well as Win32 MessageBox. For non-Windows platforms, this creates and opens a temporary text file containing the error message in a GUI editor.
+        /// This action handles errors before the logger is initialized.
+        /// </summary>
+        /// <param name="title">The title.</param>
+        /// <param name="error">The error.</param>
+        /// <param name="exit">Whether the client exits.</param>
+        public static void InitialDisplayErrorAction(string title, string error, bool exit)
         {
             Console.WriteLine(title);
             Console.WriteLine();
@@ -56,9 +63,16 @@ namespace DTAClient.Domain
 #endif
             if (exit)
                 Environment.Exit(1);
-        };
+        }
 
-        public static Action<string, string, bool> DefaultDisplayErrorAction = (title, error, exit) =>
+        /// <summary>
+        /// Show an error in Win32 MessageBox. For non-Windows platforms, this launches the log file in a GUI editor.
+        /// This action handles errors after the logger is available, but XNA windows are not initialized yet.
+        /// </summary>
+        /// <param name="title">The title.</param>
+        /// <param name="error">The error.</param>
+        /// <param name="exit">Whether the client exits.</param>
+        public static void DefaultDisplayErrorAction(string title, string error, bool exit)
         {
             Logger.Log(FormattableString.Invariant($"{(title is null ? null : title + Environment.NewLine + Environment.NewLine)}{error}"));
 #if WINFORMS
@@ -69,7 +83,7 @@ namespace DTAClient.Domain
 
             if (exit)
                 Environment.Exit(1);
-        };
+        }
 
         public static void Initialize()
         {
--- a/DXMainClient/Domain/MainClientConstants.cs
+++ b/DXMainClient/Domain/MainClientConstants.cs
@@ -61,7 +61,7 @@ namespace DTAClient.Domain
                 ProcessLauncher.StartShellProcess(ProgramConstants.LogFileName);
             else
             {
-                string tempfile = SafePath.CombineFilePath(Path.GetTempPath(), "xna-cncnet-client-error.txt");
+                string tempfile = SafePath.CombineFilePath(Path.GetTempPath(), "xna-cncnet-client-error.log");
                 using (StreamWriter writer = new StreamWriter(tempfile))
                 {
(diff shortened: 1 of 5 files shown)


DXMainClient/PreStartup.cs
```diff
@@ -278,19 +267,28 @@ private static void CheckPermissions()
...
 
             string error = string.Format(("You seem to be running {0} from a write-protected directory.\n\n" +
-                "For {1} to function properly when run from a write-protected directory, it needs administrative priveleges.\n\n" +
-                "Would you like to restart the client with administrative rights?\n\n" +
-                "Please also make sure that your security software isn't blocking {1}.").L10N("Client:Main:AdminRequiredText"), MainClientConstants.GAME_NAME_LONG, MainClientConstants.GAME_NAME_SHORT);
+                "For {1} to function properly when run from a write-protected directory, it needs administrative priveleges.").L10N("Client:Main:AdminRequiredText1"),
+                MainClientConstants.GAME_NAME_LONG, MainClientConstants.GAME_NAME_SHORT);
+
+            string question = string.Format(("Would you like to restart the client with administrative rights?\n\n" +
+                "Please also make sure that your security software isn't blocking {0}.").L10N("Client:Main:AdminRequiredText2"), MainClientConstants.GAME_NAME_SHORT);
```
Metadorius, line 274: Is the `security software` part really relevant only for the restart prompt? If not - it should be moved out of that part.

DXMainClient/Domain/MainClientConstants.cs
```diff
@@ -20,8 +27,54 @@ public static class MainClientConstants
...
+            Logger.Log(FormattableString.Invariant($"{(title is null ? null : title + Environment.NewLine + Environment.NewLine)}{error}"));
+#if WINFORMS
+            MessageBox.Show(error, title, MessageBoxButtons.OK, MessageBoxIcon.Error);
+#else
+            ProcessLauncher.StartShellProcess(ProgramConstants.LogFileName);
+#endif
+
+            if (exit)
+                Environment.Exit(1);
+        };
```
Metadorius, line 72: Kind of don't understand the distinction and duplication of the funcs here (also if the func is not changed it may be put as a regular static method), would need an explanation (could be latter put in some docstrings/comments if everything is ok).

Metadorius (changes requested):

DXMainClient/PreStartup.cs
SadPencil: Moved this sentence to `error` from `question` string

DXMainClient/Domain/MainClientConstants.cs
SadPencil: Should I reduce these two error actions to one, by determining if the logger has been initialized yet?

Metadorius: Was about to suggest that, it seems more appropriate to just check if the logger is inited inside one action.

SadPencil: Updated

DXMainClient/PreStartup.cs
```diff
@@ -82,6 +82,7 @@ public static void Initialize(StartupParams parameters)
 
             Logger.Initialize(clientUserFilesDirectory.FullName, clientLogFile.Name);
             Logger.WriteLogFile = true;
+            MainClientConstants.LoggerInitialized = true;
```
Metadorius, line 85: Isn't there a built in way to check if logger is inited? If yes - would prefer that instead, if not - then this should work as is but should be implemented in future as a part of logger class (maybe throw a TODO on the `LoggerInitialized`).

DXMainClient/PreStartup.cs
```diff
@@ -278,19 +268,28 @@ private static void CheckPermissions()
                 return;
 
             string error = string.Format(("You seem to be running {0} from a write-protected directory.\n\n" +
-                "For {1} to function properly when run from a write-protected directory, it needs administrative priveleges.\n\n" +
-                "Would you like to restart the client with administrative rights?\n\n" +
-                "Please also make sure that your security software isn't blocking {1}.").L10N("Client:Main:AdminRequiredText"), MainClientConstants.GAME_NAME_LONG, MainClientConstants.GAME_NAME_SHORT);
+                "For {1} to function properly when run from a write-protected directory, it needs administrative priveleges.\n\n"+
+                "Please also make sure that your security software isn't blocking {0}.").L10N("Client:Main:AdminRequiredText1"),
```
Metadorius, line 272:
```suggestion
                "Please also make sure that your security software isn't blocking {0}.").L10N("Client:Main:AdminRequiredExplanation"),
```

DXMainClient/PreStartup.cs
```diff
@@ -278,19 +268,28 @@ private static void CheckPermissions()
                 return;
 
             string error = string.Format(("You seem to be running {0} from a write-protected directory.\n\n" +
-                "For {1} to function properly when run from a write-protected directory, it needs administrative priveleges.\n\n" +
-                "Would you like to restart the client with administrative rights?\n\n" +
-                "Please also make sure that your security software isn't blocking {1}.").L10N("Client:Main:AdminRequiredText"), MainClientConstants.GAME_NAME_LONG, MainClientConstants.GAME_NAME_SHORT);
+                "For {1} to function properly when run from a write-protected directory, it needs administrative priveleges.\n\n"+
```
Metadorius, line 271:
```suggestion
                "For {1} to function properly when run from a write-protected directory, it needs administrative priveleges.\n\n" +
```

DXMainClient/PreStartup.cs
```diff
@@ -278,19 +268,28 @@ private static void CheckPermissions()
...
+                "Please also make sure that your security software isn't blocking {0}.").L10N("Client:Main:AdminRequiredText1"),
+                MainClientConstants.GAME_NAME_LONG, MainClientConstants.GAME_NAME_SHORT);
+
+            string question = string.Format("Would you like to restart the client with administrative rights?".L10N("Client:Main:AdminRequiredText2"), MainClientConstants.GAME_NAME_SHORT);
```
Metadorius, line 275:
```suggestion
            string question = "Would you like to restart the client with administrative rights?".L10N("Client:Main:AdminRequiredRestartPrompt");
```

Metadorius (changes requested):
Re-check the formatted string, it looks wrong to me, sorry missed it on the first review.

Metadorius (approved):

PR #534, opened 2024-03-15
Fix `BtnSaveLoadGameOptions` button naming

MahBoiDeveloper:
This button is like a black sheep in the client configuration files. All buttons in the config files have `btn` in lower case, for example `btnLaunchGame`, but `BtnSaveLoadGameOptions` is an exception to this rule. It was a typo, I suppose.

After this PR config `GameLobbyBase.ini` should be modified to lower case `BtnSaveLoadGameOptions` to `btnSaveLoadGameOptions`, for example:

```diff
Before
-$CC05=BtnSaveLoadGameOptions:XNAClientButton

After
+$CC05=btnSaveLoadGameOptions:XNAClientButton
```


Metadorius (approved):

Metadorius:
Not sure why you're bringing this up 1 year after it was merged. The button is quite new and wasn't used by a lot of clients, so it's fine the way it is

PR #544, opened 2024-04-27
Make SettingCheckBoxBase and SettingDropDownBase can be i18n

frg2089:
Make SettingCheckBoxBase and SettingDropDownBase be translatable.

### DIFF
--- a/DTAConfig/Settings/SettingCheckBoxBase.cs
+++ b/DTAConfig/Settings/SettingCheckBoxBase.cs
@@ -1,7 +1,9 @@
 ﻿using System;
+using ClientCore.I18N;
 using ClientGUI;
 using Rampastring.Tools;
 using Rampastring.XNAUI;
+using Rampastring.XNAUI.XNAControls;
 
 namespace DTAConfig.Settings
 {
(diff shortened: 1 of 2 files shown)


DTAConfig/Settings/FileSettingCheckBox.cs
```diff
@@ -57,6 +59,9 @@ public override void GetAttributes(IniFile iniFile)
                 enabledFiles = FileSourceDestinationInfo.ParseFSDInfoList(section, "EnabledFile");
                 disabledFiles = FileSourceDestinationInfo.ParseFSDInfoList(section, "DisabledFile");
             }
+
+            Text = Name.L10N($"Client:DTAConfig:{Name}:Text");
+            ToolTipText = string.Empty.L10N($"Client:DTAConfig:{Name}:ToolTip");
```
SadPencil, line 64: L10N($"INI:xxx") for dynamic strings that are not relevant to the client

SadPencil (changes requested):

Metadorius (changes requested):
I don't understand why are the changes needed. Look up TranslationINIParser, it already reads localizable attributes, including `Text` and `ToolTip`, without modifying the actual controls.

PR #546, opened 2024-06-07
make compiler error

frg2089:
It should be a compiler error not a syntax error

[CS1029](https://learn.microsoft.com/dotnet/csharp/language-reference/compiler-messages/cs1029)

image


pzhlkj6612 (approved):
Interesting code line.

SadPencil (approved):

PR #547, opened 2024-06-07
Allows passing arguments to gamemd.exe

### LINKED ISSUE #471: Allow the client launcher to pass on command line arguments to the game it is launching.
Allow the client launcher to pass on command line arguments to the game it is launching. This would allow users to create a shortcut to the main client launcher executable, and hard code these to the `ClientDefintions.ini`.

For example;
`TiberianSun.exe -GAME_ARGS=-DEVELOPER`.

Vinifera implements a new "developer" feature to assist mod creators in their process, but mod creators need to keep modifying `ExtraCommandLineParams=` in `ClientDefintions.ini` to remove this before they push their update, sometimes this is not always done and results in an update being pushed where the end users enter the developer mode and another update is required to be pushed to revert this. Support for passing on custom arguments like this via the client launcher would allow mod creators to leave `ExtraCommandLineParams=` untouched, and then create a local shortcut for this to run the game in the developer mode.

Open to other suggestions as to how this could be implemented.

frg2089:
Allows passing arguments to gamemd.exe using something like `clientdx.exe -- <options>`.

```powershell
clientdx.exe -MULTIPLEINSTANCE -- -SPAWN # Pass the -SPAWN argument to gamemd.exe
```

Close #471

### DIFF
--- a/ClientGUI/GameProcessLogic.cs
+++ b/ClientGUI/GameProcessLogic.cs
@@ -69,6 +69,9 @@ namespace ClientGUI
 
             string extraCommandLine = ClientConfiguration.Instance.ExtraExeCommandLineParameters;
 
+            if (Environment.CommandLine.Contains(" -- "))
+                extraCommandLine += Environment.CommandLine[(Environment.CommandLine.IndexOf("--") + 2)..];
+
             SafePath.DeleteFileIfExists(ProgramConstants.GamePath, "DTA.LOG");
             SafePath.DeleteFileIfExists(ProgramConstants.GamePath, "TI.LOG");
             SafePath.DeleteFileIfExists(ProgramConstants.GamePath, "TS.LOG");


SadPencil:
Personally I am in favor of #535 as it is most straightforward to specify arguments to the same place where the executable name is specified

SadPencil:
Apologies for my last reply. But, I think it is still better to provide a separate .ini file to override this argument. I don't think modders as in #471 expects a command-line

frg2089:
> Apologies for my last reply. But, I think it is still better to provide a separate .ini file to override this argument. I don't think modders as in #471 expects a command-line

并非一定要在命令行中打开，也可以在快捷方式的目标属性中添加参数。

我这样设计它也是出于这种目的

SadPencil:
> > Apologies for my last reply. But, I think it is still better to provide a separate .ini file to override this argument. I don't think modders as in #471 expects a command-line
> 
> It's not always necessary to open it in the command line, but you can also add parameters to the target properties of the shortcut.
> 
> I designed it this way for this purpose as well

Another reason I am against it is that a client can only have one set of unnamed command-line arguments, and therefore it needs a strong reason to occupy it. I know in some designs like 'docker run' or something else all unnamed command-line arguments after '--' are passed to the program inside but I doubt whether it is also reasonable for this client. The difference is that xna-cncnet-client does not automatically runs the game executable like 'docker run': launching the game executable is not the *only* feature -- e.g, users could alternatively launch the game(md).exe file in CnCNet YR for a classic campaign, or launch the map editor. Through your design, it is not straightforward for users to tell whether the unnamed parameters be passed to all these executables or just the gaming executable. That's why I am opposing it.

Metadorius:
What about anti-cheat? You should realize that passing different command args by different players can lead to cheating undetected by XNA client. Imagine passing `-INCLUDE -INHERITANCE` when it isn't on in a Phobos mod, and the map would have the code that only works with those tags. The anticheat won't go off but the cheats will be still on only for the people with those tags working.

SadPencil:
I'll close this one for now. According to the conversation we need an INI configurable commandline argument.

PR #549, opened 2024-07-09
Fix current selected map is not preserved during game mode changes

SadPencil:
Currently, previous map is only preserved if *both* game mode *and* map are identical before and after changes. So it won't work as people usually expected.

This PR corrects this behavior so that as long as there exists a `GameModeMap` whose map is identical before and after changes, regardless of the game mode, it would be selected. Note: if a `GameModeMap` matches both game mode and the map (i.e., switching to/from favorite map), it would have the top priority.


SadPencil:
https://github.com/CnCNet/xna-cncnet-client/blame/56638037c79c8b2f611aac6c32223d042e123ede/DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs#L625

Metadorius (approved):

PR #554, opened 2024-09-07
Ignore case when selecting default translation code

SadPencil:
Although the constructor `new CultureInfo(string)` is [defined](https://learn.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.-ctor?view=net-8.0#system-globalization-cultureinfo-ctor(system-string)) as case-insensitive, the client does not fully comply with this convention right now.

As a result, when translators created a folder named `zh-hans` (while the normalized name should be `zh-Hans`), the client will not be able to select this translation as the default.

This PR fixes this issue, allowing the folder name be case-insensitive.

### DIFF
--- a/DTAConfig/OptionPanels/DisplayOptionsPanel.cs
+++ b/DTAConfig/OptionPanels/DisplayOptionsPanel.cs
@@ -679,7 +679,7 @@ namespace DTAConfig.OptionPanels
                 ddi => (string)ddi.Tag == UserINISettings.Instance.ClientTheme);
             ddClientTheme.SelectedIndex = selectedThemeIndex > -1 ? selectedThemeIndex : 0;
 
-            foreach (string localeCode in [UserINISettings.Instance.Translation, Translation.GetDefaultTranslationLocaleCode(), ProgramConstants.HARDCODED_LOCALE_CODE])
+            foreach (string localeCode in new string[] { UserINISettings.Instance.Translation, Translation.GetDefaultTranslationLocaleCode(), ProgramConstants.HARDCODED_LOCALE_CODE })
             {
                 int selectedTranslationIndex = ddTranslation.Items.FindIndex(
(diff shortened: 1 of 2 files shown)


DTAConfig/OptionPanels/DisplayOptionsPanel.cs
```diff
@@ -679,20 +679,20 @@ public override void Load()
...
 
-            int selectedTranslationIndex = ddTranslation.Items.FindIndex(
-                ddi => (string)ddi.Tag == UserINISettings.Instance.Translation);
-
-            if (selectedTranslationIndex > -1)
-            {
-                ddTranslation.SelectedIndex = selectedTranslationIndex;
-            }
-            else
+            foreach (string localeCode in new string[] { UserINISettings.Instance.Translation, Translation.GetDefaultTranslationLocaleCode(), ProgramConstants.HARDCODED_LOCALE_CODE })
```
Metadorius, line 682:
```suggestion
            foreach (string localeCode in [UserINISettings.Instance.Translation, Translation.GetDefaultTranslationLocaleCode(), ProgramConstants.HARDCODED_LOCALE_CODE])
```

SadPencil: I remember the compiler would complain about writing `[]` in a foreach loop. Not pretty sure. Will commit and wait for the bot compiling the codes

SadPencil:
```
Error: D:\a\xna-cncnet-client\xna-cncnet-client\DTAConfig\OptionPanels\DisplayOptionsPanel.cs(682,43): error CS9176: There is no target type for the collection expression. [D:\a\xna-cncnet-client\xna-cncnet-client\DTAConfig\DTAConfig.csproj::TargetFramework=net8.0]
```

PR #568, opened 2024-10-04
Print detailed client version in the log

SadPencil:
This PR prints detailed client version ("ProductVersion") in the log, for example, `2.12.0-feat-detailed-client-version.1+9.Branch.feat-detailed-client-version.Sha.54da550909971152332380ca79fe1b732501ec3a`. Containing the git commit should be extremely helpful to minimize the mess and determine the client version if a user uses a non-release build.


Metadorius (approved):
LGTM but maybe we should keep both simple version and detailed version?

Also we really should introduce separation between non-released version and released version. Can be as little as a different message in the UI indicating detailed info.

PR #570, opened 2024-10-09
Game listings fetched by IRC List

GrantBartlett:
image

### Changes
- Request LIST of game lobby, should return somewhat instantly the list of games
- Parse the topic of the channel to give the CnCNet lobby game information (as opposed to using ctcp)
- Updates topic if game options have changed

PR #571, opened 2024-10-10
WIP: Feature/map peview game info

GrantBartlett:
**Show map preview in Game information hover**
image

- Adds map preview to game info hover
- Renders ping images like in game lobby

It's branched from https://github.com/CnCNet/xna-cncnet-client/pull/570 so waiting on this before updating


ismail-doitbig:
If it's a custom map does it automatically download the .map file?
I don't want to clutter my pc with these map files, the map name is good info, I'd like to give consent first

And would I be able to see the players in the map?

GrantBartlett:
> If it's a custom map does it automatically download the .map file? I don't want to clutter my pc with these map files, the map name is good info, I'd like to give consent first
> 
> And would I be able to see the players in the map?

It would probably just be a blank preview, unknown yet

PR #575, opened 2024-10-15
Reduce default timeout to 10 seconds

SadPencil:
The web client was created by `new WebClient()` without assigning a timeout value. It seems that the default timeout value is undocumented, and the debugger told me the default value is 100000 ms (note: 100 seconds!) No wonder why some people with poor Internet condition complained about it takes too long to wait for their clients to initialize.

This PR proposed the following changes:
- By default, the timeout value would be 10 seconds instead of 100 seconds
- As an exception, the timeout value of uploading/downloading a map is set to 100 seconds
- Remove the obsolete `TWebClient` introduced 8 years ago ([here](https://github.com/CnCNet/xna-cncnet-client/commit/d33e7c2a3a55381d8dd98dc5bc9ee491af7e6575)) since `ExtendedWebClient` was introduced 6 years ago ([here](https://github.com/CnCNet/xna-cncnet-client/commit/bb3dbc83c77f64a4864d497ec68ee10da7e246f2))
- Remove the weird statement `webClient.Proxy = null` introduced 8 years ago ([here](https://github.com/CnCNet/xna-cncnet-client/commit/d33e7c2a3a55381d8dd98dc5bc9ee491af7e6575)). Since now WebClients are initialized everywhere and no one explicitly clear the proxy setting like here, removing this statement will make the client behaves consistently -- leaving the proxy setting as system default.

### DIFF
--- a/DXMainClient/Domain/Multiplayer/CnCNet/CnCNetPlayerCountTask.cs
+++ b/DXMainClient/Domain/Multiplayer/CnCNet/CnCNetPlayerCountTask.cs
@@ -52,11 +52,11 @@ namespace DTAClient.Domain.Multiplayer.CnCNet
             {
                 // Don't fetch the player count if it is explicitly disabled
                 // For example, the official CnCNet server might be unavailable/unstable in a country with Internet censorship,
-                // which causes lags in the splash screen. In the worst case, say if packets are dropped, it waits until timeouts --- 30 seconds
+                // which causes lags in the splash screen. In the worst case, say if packets are dropped, it waits until timeouts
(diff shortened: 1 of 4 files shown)


ismail-doitbig:
Sorry this might be off topic but I noticed a bunch of trollers on purpose disconnect their cable for two seconds and then reconnect again, they do this until everyone votes to kick them or people just leave because it's too annoying..

Can something be done about this?

E.g. If the cumulative delay is 10s over a 60s time period the person should be kicked so trollers that disconnect 3 seconds each, after the 4th time will get autokicked

If this is possible I would be forever thankful!!

Metadorius (approved):

PR #582, opened 2024-11-24
When there is only 1 player, the selection of random map will randomize to all maps

Flactine:
When the number of players is 1, the random map will be assigned to any map, regardless of the max players of the map.
Used to randomly check maps.


Flactine:
Well, I was just trying to change a few things in my own library, and the Pull Request was affected... I'll write another one later

PR #583, opened 2024-11-25
Randomize all maps when there is only one player

Flactine:
When the number of players is 1, the random map will be assigned to any map, regardless of the max players of the map.
Used to randomly check maps.

### DIFF
--- a/DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
+++ b/DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
@@ -742,6 +742,11 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
 
         private List<Map> GetMapList(int playerCount)
         {
+            if (playerCount == 1)
+			{
+				List<Map> allMaps = GameMode?.Maps.ToList() ?? new List<Map>();
+				return allMaps;
+			}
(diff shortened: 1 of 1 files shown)


DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
```diff
@@ -742,6 +742,11 @@ private void PickRandomMap()
 
         private List<Map> GetMapList(int playerCount)
         {
+            if (playerCount == 1)
+			{
+				List<Map> allMaps = GameMode?.Maps.ToList() ?? new List<Map>();
+				return allMaps;
+			}
```
SadPencil, line 749: Add a line break after }

SadPencil (changes requested):
Note: please address the review comments before your PR gets merged

SadPencil:
@Flactine This is a reminder that please address the review comment since we want to merge this PR.

Flactine (commented):
Good

SadPencil:
> Good

You need to **address** the review comment, which means you need to pushing more commits so that your codes does not violate the requirements. Not hitting the "Resolve conversation" button -- it just sent the information that **you** have addressed the comment, rather than automatically let GitHub solve it.

SadPencil:
@Flactine After you have addressed all concerns, please hit "Resolve conversation" button and then re-request reviews. See the screenshot below.

SadPencil (approved):

PR #590, opened 2024-12-03
Implementing file linking for the client file system

MahBoiDeveloper:
The current client doesn't use links for with the file system operations such as applying and reverting file changes for `FileSettingDropDown` or `FileSettingCheckBox` and operates with files by copy-pasting them. PR is designed to fix this problem by upgrading the `KeepChanges` file operation options to use hard links. Creating a hard link to the large file is much faster than copying it.

> [!TIP]
> To see links from Windows Explorer, please install [this extension](https://schinagl.priv.at/nt/hardlinkshellext/linkshellextension.html).

> [!IMPORTANT]
> When you open a hardlinked file for edit, it depends on the editor associated to the file if the file either gets
>
>    opened, changed, deleted, and saved ( ==> link broken; the other one is not modified )
>    opened, changed, and saved back ( ==> link alive; the other one is modified  )
> 
> And therefore, we **do not recommend** editing a file that is acting as/referred from a hard link, unless both behaviors are acceptable to you.

UPD 2024-05-12:
`FileSettingDropDown` and `FileSettingCheckBox` supports hard links with fallback to file copying if client is unable to create link for file operation options `KeepChanges` and `AlwaysOverwrite_LinkAsReadOnly`.

UPD 2024-06-12:
Now links work with `ddraw.dll` and translation game files.

Todo list:
- [x] Add OS platform and .NET detecting mechanism more accurate i.e., need to separate the runner (Windows with .NET Framework 4.8 and any Linux distro (with WINE or Mono) or Windows with .NET 8.0).
- [x] Change name for `CreateSymbolicLink` to `CreateLinkToFile`.
- [x] Handle exceptions when link deploy is unavailable.
- [x] Make link (symbolic) uses relative paths.
- [x] Make source file is read only if link established (link logic moved to 2 new `FileOperationOption`: `ImmutableLink` and `Link`).
- [x] Move link method to the different class.
- [x] Improve translation game files copying with new linking method.
- [x] Make `ddraw.dll` link instead file copy.
- [x] Update documentation.
    - [x] Mention that ReShade files better to link instead of copy them due to antivirus checks when files copying.
- [x] * Record speed of changing for large files with symlink and not. <br/>

<details>
<summary>Speed of changing.</summary>

Tested on [Tiberian Sun Client](https://www.moddb.com/mods/russian-dubbing-and-translation/downloads/tiberian-sun-client-606-russian-dub-2024) translation by [TheMonarchist](https://www.moddb.com/members/themonarchist). In this translation the most biggest archives are `MOVIES01.mix`, `MOVIES02.mix`, `MOVIES03.mix` more than 400MB *each* with 77 files total.

Hard drive: SSD.

<details>
<summary>Translations section demonstration from `ClientDefinitions.ini`.</summary>

```ini
[Translations]
; Example of usage
; You need put file from %CD%/Resources/%LOCALE%/FROM to %CD%/WHERE
; So you should use keys and values like this:
;   GameFileX=FROM,WHERE,OPTIONAL_PARAMETER
; where:
;   X                  - number from 0 to infinity,
;   FROM               - path to file into language folder,
;   WHERE              - path to file into destination folder,
;   OPTIONAL_PARAMETER - read the doc to understand.

; Localized in-game fonts
GameFile0=MIX/ecache98.mix,MIX/ecache98.mix
GameFile1=MIX/expand98.mix,MIX/expand98.mix

; In-game menu text
GameFile2=DLL/Language_1024x720.dll,Resources/Language_1024x720.dll
GameFile75=DLL/language_640x480.dll,Resources/language_640x480.dll
GameFile76=DLL/language_800x600.dll,Resources/language_800x600.dll

; In-game mission text and units naming
GameFile3=INI/LangFS.ini,INI/LangFS.ini
GameFile4=INI/tutorial.ini,INI/tutorial.ini

; TS GDI missions with localized briefings
GameFile5=Missions/gdi1a.map,Maps/Missions/gdi1a.map
GameFile6=Missions/gdi1adshp.map,Maps/Missions/gdi1adshp.map
GameFile7=Missions/gdi2a.map,Maps/Missions/gdi2a.map
GameFile8=Missions/gdi3a.map,Maps/Missions/gdi3a.map
GameFile9=Missions/gdi3b.map,Maps/Missions/gdi3b.map
GameFile10=Missions/gdi4a.map,Maps/Missions/gdi4a.map
GameFile11=Missions/gdi5a.map,Maps/Missions/gdi5a.map
GameFile12=Missions/gdi5b.map,Maps/Missions/gdi5b.map
GameFile13=Missions/gdi5c.map,Maps/Missions/gdi5c.map
GameFile14=Missions/gdi6a.map,Maps/Missions/gdi6a.map
GameFile15=Missions/gdi6b.map,Maps/Missions/gdi6b.map
GameFile16=Missions/gdi7a.map,Maps/Missions/gdi7a.map
GameFile17=Missions/gdi8a.map,Maps/Missions/gdi8a.map
GameFile18=Missions/gdi9a.map,Maps/Missions/gdi9a.map
GameFile19=Missions/gdi9b.map,Maps/Missions/gdi9b.map
GameFile20=Missions/gdi9c.map,Maps/Missions/gdi9c.map
GameFile21=Missions/gdi9d.map,Maps/Missions/gdi9d.map
GameFile22=Missions/gdi10a.map,Maps/Missions/gdi10a.map
GameFile23=Missions/gdi10b.map,Maps/Missions/gdi10b.map
GameFile24=Missions/gdi11a.map,Maps/Missions/gdi11a.map
GameFile25=Missions/gdi12a.map,Maps/Missions/gdi12a.map

; TS Nod missions with localized briefings
GameFile26=Missions/nod1a.map,Maps/Missions/nod1a.map
GameFile27=Missions/nod2a.map,Maps/Missions/nod2a.map
GameFile28=Missions/nod3a.map,Maps/Missions/nod3a.map
GameFile29=Missions/nod3b.map,Maps/Missions/nod3b.map
GameFile30=Missions/nod4a.map,Maps/Missions/nod4a.map
GameFile31=Missions/nod4b.map,Maps/Missions/nod4b.map
GameFile32=Missions/nod5a.map,Maps/Missions/nod5a.map
GameFile33=Missions/nod6a.map,Maps/Missions/nod6a.map
GameFile34=Missions/nod6b.map,Maps/Missions/nod6b.map
GameFile35=Missions/nod6c.map,Maps/Missions/nod6c.map
GameFile36=Missions/nod7a.map,Maps/Missions/nod7a.map
GameFile37=Missions/nod7b.map,Maps/Missions/nod7b.map
GameFile38=Missions/nod8a.map,Maps/Missions/nod8a.map
GameFile39=Missions/nod9a.map,Maps/Missions/nod9a.map
GameFile40=Missions/nod9b.map,Maps/Missions/nod9b.map
GameFile41=Missions/nod10a.map,Maps/Missions/nod10a.map
GameFile42=Missions/nod11a.map,Maps/Missions/nod11a.map
GameFile43=Missions/nod12a.map,Maps/Missions/nod12a.map
GameFile44=Missions/nod12b.map,Maps/Missions/nod12b.map

; TS tutorial missions with localized briefings
GameFile45=Missions/tsdemo1.map,Maps/Missions/tsdemo1.map
GameFile46=Missions/tsdemo2.map,Maps/Missions/tsdemo2.map

; FS GDI missions with localized briefings
GameFile47=Missions/fsgdi01.map,Maps/Missions/fsgdi01.map
GameFile48=Missions/fsgdi02.map,Maps/Missions/fsgdi02.map
GameFile49=Missions/fsgdi03.map,Maps/Missions/fsgdi03.map
GameFile50=Missions/fsgdi04.map,Maps/Missions/fsgdi04.map
GameFile51=Missions/fsgdi05.map,Maps/Missions/fsgdi05.map
GameFile52=Missions/fsgdi06.map,Maps/Missions/fsgdi06.map
GameFile53=Missions/fsgdi07.map,Maps/Missions/fsgdi07.map
GameFile54=Missions/fsgdi08.map,Maps/Missions/fsgdi08.map
GameFile55=Missions/fsgdi09.map,Maps/Missions/fsgdi09.map

; FS Nod missions with localized briefings
GameFile56=Missions/fsnod01.map,Maps/Missions/fsnod01.map
GameFile57=Missions/fsnod02.map,Maps/Missions/fsnod02.map
GameFile58=Missions/fsnod03.map,Maps/Missions/fsnod03.map
GameFile59=Missions/fsnod04.map,Maps/Missions/fsnod04.map
GameFile60=Missions/fsnod05.map,Maps/Missions/fsnod05.map
GameFile61=Missions/fsnod06.map,Maps/Missions/fsnod06.map
GameFile62=Missions/fsnod07.map,Maps/Missions/fsnod07.map
GameFile63=Missions/fsnod08.map,Maps/Missions/fsnod08.map
GameFile64=Missions/fsnod09.map,Maps/Missions/fsnod09.map

; Dubbed movies
GameFile65=MIX/MOVIES01.mix,MIX/MOVIES01.mix
GameFile66=MIX/MOVIES02.mix,MIX/MOVIES02.mix
GameFile67=MIX/MOVIES03.mix,MIX/MOVIES03.mix

; Dubbed voiceovers
GameFile68=MIX/e01vox01.mix,MIX/e01vox01.mix
GameFile69=MIX/e01vox02.mix,MIX/e01vox02.mix
GameFile70=MIX/sidecd01.mix,MIX/sidecd01.mix
GameFile71=MIX/sidecd02.mix,MIX/sidecd02.mix
GameFile72=MIX/Sounds.mix,MIX/Sounds.mix
GameFile73=MIX/speech01.mix,MIX/speech01.mix
GameFile74=MIX/speech02.mix,MIX/speech02.mix
```

</details>

Without PR:

https://github.com/user-attachments/assets/4a1d995d-3a1d-4f91-8292-988c2c7c77f5

With PR:

https://github.com/user-attachments/assets/c1f6801d-9fe2-4e97-832d-d4e064de5595
</details>


SadPencil:
Well, I have different opinion on this proposal.

Symlinks or hard links are different from Copy-On-Write: either file be modified means both files be modified. And unfortunately, NTFS does not support CoW at all.

Long long ago, I wrote a software [here](https://github.com/SadPencil/mo3-mod-manager) utilizing NTFS hard links to minimize the disk space consumption with multiple mods. The result is, users that are not familiar with advanced computer skills are struggling with it, and sometimes files that aren't supposed be modified are modified due to the reason mentioned above. A small proportion of users love this software since they are capable of these knowledge, but the number of them is too small.

Should you insist on this feature, I suggest you explicitly specify known text file extensions, and make sure mklink never acts on text files (since users are tend to modify them) or any files less than e.g., 1MB. Making links for DLL and EXE files are relatively safe (I don't know whether Resource Hacker directly modifys an EXE file or renames it first, i.e., whether modifying a link will make an influence on the source file. You also need to investigate this point)

SadPencil:
Also, even if `#if NETFRAMEWORK` is written, I suggest you make use of `SupportedOSPlatform` and `RuntimeInformation.IsOSPlatform(OSPlatform.Windows)` before invoking Win32 APIs since people could also running a .net framework program with wine.

DTAConfig/Settings/FileSourceDestinationInfo.cs
```diff
@@ -151,6 +134,27 @@ public void Revert()
...
+        );
+#endif
+
+        /// <summary>
+        ///  Creates a symbolic link from the source file to the destination file.
+        /// </summary>
+        private void CreateSymbolicLink(string source, string destination)
+        {
+#if NETFRAMEWORK
+            CreateHardLink(destination, source, IntPtr.Zero);
```
SadPencil, line 153: Apart from the comments above, symlinks and hard links are two different concepts. You should call `CreateSymbolicLinkW` to make this function correctly behaves

SadPencil (changes requested):

SadPencil:
Also, there is a small chance that users might run the client in a FAT32 filesystem so creating links are not always works. You should use `Marshal.GetLastWin32Error()` to determine if mklink fails and fallback to regular copying. Similarly, you need a try-catch for the `File.CreateSymbolicLink()` function

MahBoiDeveloper:
> Should you insist on this feature, I suggest you explicitly specify known text file extensions, and make sure mklink never acts on text files (since users are tend to modify them) or any files less than, e.g., 1MB. Making links for DLL and EXE files are relatively safe (I don't know whether Resource Hacker directly modifys an EXE file or renames it first, i.e., whether modifying a link will make an influence on the source file. You also need to investigate this point)

This is a pretty good idea to unable this feature for text files and keep only for binary files. At this point @Metadorius suggest to add a new file operation option (e.g., `Hardlink`).

@SadPencil also thanks for advises about platform detecting and FAT32 detecting mechanisms and linking issues, i'll try to implement this in the next commit.

SadPencil:
Also, it might be better to set the read-only flag on both source file and target symlink/hard link file? At least read-only property can act as an implicit reminder

SadPencil:
Also, for symlinks, I suggest using relative paths. Since users might copy/rename/zip their game folder.

MahBoiDeveloper:
> Also, could you list some example large files copied by `FileSettingDropDown` or `FileSettingCheckBox` (filenames, file types, and sizes)?

In this branch exists `DXMainClient/Resources/DTA/ReShade Files/opengl32.dll` that makes client lag when it copies and gives no lag with creating hard link. More large example haven't tested yet.

SadPencil:
> > Also, could you list some example large files copied by `FileSettingDropDown` or `FileSettingCheckBox` (filenames, file types, and sizes)?
> 
> In this branch exists `DXMainClient/Resources/DTA/ReShade Files/opengl32.dll` that makes client lag when it copies and gives no lag with creating hard link. More large example haven't tested yet.

How large this file is? If it's less than 1MB, probably it's due to an anti-virus software. Anyway, making links for DLL files is fair.

Does your PR also help on applying a DDRAW.dll file? (don't link the corresponding ini file)

MahBoiDeveloper:
> How large is this file? If it's less than 1MB, probably it's due to an anti-virus software. Anyway, making links for DLL files is fair.
> 
> Does your PR also help on applying a DDRAW.dll file? (don't link the corresponding ini file)

PR right now supports only `FileSettingCheckBox` and `FileSettingDropDown`.

DTAConfig/Settings/FileSourceDestinationInfo.cs
SadPencil: > thing is, you can't create symlinks on Windows as a regular user, we're currently investigating if that is solvable by adequate means

At least one should never create a hard link in a function explicitly named `CreateSymbolicLink`

MahBoiDeveloper:
> Also, even if `#if NETFRAMEWORK` is written, I suggest you make use of `SupportedOSPlatform` and `RuntimeInformation.IsOSPlatform(OSPlatform.Windows)` before invoking Win32 APIs since people could also running a .net framework program with wine in linux.

AFAIK official CnCNet team statement is Linux users should use .NET 8.0 version client instead of .NET 4.8. I can see no reason not to add another layer of security checks, but then why does .NET 8.0 version even exist?

SadPencil:
> > Also, even if `#if NETFRAMEWORK` is written, I suggest you make use of `SupportedOSPlatform` and `RuntimeInformation.IsOSPlatform(OSPlatform.Windows)` before invoking Win32 APIs since people could also running a .net framework program with wine in linux.
> 
> AFAIK official CnCNet team statement is Linux users should use .NET 8.0 version client instead of .NET 4.8. I can see no reason not to add another layer of security checks, but then why does .NET 8.0 version even exist?

- Modders decide whether including .NET 8 clients in their mods. If a modder decides not to include .NET 8 clients, a linux user might choose using Mono to prevent hash mismatch warning
- The client does not actively refuse to launch when mono is detected right now. Imagine the client goes 99% well in mono without showing any warnings and it only fails on minor conditions like this. It will waste so many hours for a user to finally figure it out
- Currently, the source codes of this client has already employed such checks for almost every Windows-specific cases. It does not cost more than a few lines so why breaking this.
- The function that makes symlinks/hardlinks should be highly reusable. So a robust implementation will probably make us easier to copy codes to other projects in case it's needed.

SadPencil:
Also,
> you can't create symlinks on Windows as a regular user
suppose this statement is true. You should also invoke Win32API to create hard links instead of symlinks even in .NET 8, right?

MahBoiDeveloper:
> Also,
> 
> > you can't create symlinks on Windows as a regular user
> 
> suppose this statement is true. You should also invoke Win32API to create hard links instead of symlinks even in .NET 8, right?

Yes. .NET 8 has only `File.CreateSymbolicLink` method that doesn't work, if you start program process without admin rights, because it creates symbolic links and not any hard links or junctions (soft links). .NET 4.8 doesn't have any API methods to create this links.

MahBoiDeveloper:
> Make source and link is read-only.

Win32 API doesn't support to make hard links are read only. 
Reference: https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createhardlinkw

I suppose it possible to implement if mark original file as read only.

SadPencil:
> > Make source and link is read-only.
> 
> 
> 
> Win32 API doesn't support to make hard links are read only. 
> 
> Reference: https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createhardlinkw
> 
> 
> 
> I suppose it possible to implement if mark original file as read only.

A hard link is non-distinguishable from its source -- it's a file. Set the file as readonly

DTAConfig/Settings/FileSourceDestinationInfo.cs
```diff
@@ -151,6 +134,35 @@ public void Revert()
...
+        /// </summary>
+        private void CreateLinkToFile(string source, string destination)
+        {
+#if NETFRAMEWORK
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                bool result = CreateHardLink(destination, source, IntPtr.Zero);
+                if (!result)
+                    throw new Exception(Marshal.GetExceptionForHR(Marshal.GetHRForLastWin32Error()).Message);
+            }
```
SadPencil, line 154: missing else

SadPencil (changes requested):

DTAConfig/Settings/FileSourceDestinationInfo.cs
```diff
@@ -100,17 +101,7 @@ public void Apply()
...
-                        if (File.Exists(CachedPath))
-                            File.Copy(CachedPath, DestinationPath, false);
-                        else
-                            File.Copy(SourcePath, DestinationPath, false);
-                    }
-
-                    Directory.CreateDirectory(Path.GetDirectoryName(CachedPath));
-                    File.Copy(DestinationPath, CachedPath, true);
-
+                    CreateLinkToFile(sourcePath, destinationPath);
```
SadPencil, line 104: Don't forget fallback to regular copying

SadPencil (changes requested):

DTAConfig/Settings/FileSourceDestinationInfo.cs
MahBoiDeveloper: if you mean add `File.CreateSymbolicLink` -- it isn't possible due to .NET 4.8 limitations. `File.CreateSymbolicLink` needs .NET 6 or newer.
Reference: https://learn.microsoft.com/en-us/dotnet/api/system.io.file.createsymboliclink?view=net-8.0

```cs
else
    File.CreateSymbolicLink(destination, source); // this will get compilation error in .NET 4.8
```

SadPencil: throw an Exception

MahBoiDeveloper: > throw an Exception

`bool result` contains the result of working `CreateHardLink`. And `!result` triggers when it is a `false` value.

SadPencil: No. Currently, your method will not throw an exception for .NET 4.8 on non-Windows platform.

MahBoiDeveloper: > No. Currently, your method will not throw an exception for .NET 4.8 on non-Windows platform.

Maybe better remove OS check for .NET 4.8? For Windows it would work as always, and WINE on Linux could wrap it.

SadPencil: There is also a mono case. 

Don't remove if-windows here. Throw an exception. The reasons were listed earlier

SadPencil:
> @Metadorius @SadPencil What do you think about moving `CreateLinkToFile` method to the `File` class extension?

I am not sure if this is a C# 13 feature or not.

DTAConfig/Settings/FileSourceDestinationInfo.cs
MahBoiDeveloper: > There is also a mono case.
> 
> Don't remove if-windows here. It's better to explicitly throw an exception. The reasons were listed earlier

Ok, what if for Linux i would be use `link` from `libc.so`?

https://www.man7.org/linux/man-pages/man2/link.2.html

```cs
else
{
    int result = link(destination, source);
    if (result != 0)
        throw new Exception(string.Format("Unable to create hard link to file {0}", source));
}
```

DTAConfig/Settings/FileSourceDestinationInfo.cs
MahBoiDeveloper: > Don't forget fallback to regular copying

Why though? This implementation doesn't need that undefined behavior, IMO.

If you want, i can revert changes for `KeepChanges` to original file copy and add new `ImmutableLink` option for links.

DTAConfig/Settings/FileSourceDestinationInfo.cs
SadPencil: > > There is also a mono case.
> > Don't remove if-windows here. It's better to explicitly throw an exception. The reasons were listed earlier
> 
> Ok, what if for Linux i would be use `link` from `libc.so`?
> 
> https://www.man7.org/linux/man-pages/man2/link.2.html
> 
> ```cs
> else
> {
>     int result = link(source, destination);
>     if (result != 0)
>         throw new Exception(string.Format("Unable to create hard link to file {0}", source));
> }
> ```

I won't recommend so since you will have to dynamically link both glibc (with a minimum version number) and musl-libc for a wide Linux support. This causes a lot of efforts in .NET.

DTAConfig/Settings/FileSourceDestinationInfo.cs
SadPencil: with fallback -- you are proposing a feature that at least fallback to previous behavior. linking is an "optimization" to copying.

without fallback -- you are breaking the client for partial users

Not all file systems support linking. An "ImmutableLink" option is also not a good idea because nor us developer or the modder should decide whether all users must use linking or not.

DTAConfig/Settings/FileSourceDestinationInfo.cs
```diff
@@ -151,6 +134,35 @@ public void Revert()
                         $"Invalid {nameof(FileOperationOptions)} value of {FileOperationOptions}");
             }
         }
+
+        [DllImport("Kernel32.dll", SetLastError = true, CharSet = CharSet.Unicode)]
```
SadPencil, line 176: Note:

> Also, even if #if NETFRAMEWORK is written, I suggest you make use of SupportedOSPlatform and RuntimeInformation.IsOSPlatform(OSPlatform.Windows) before invoking Win32 APIs since people could also running a .net framework program with wine or mono in linux.

`SupportedOSPlatform` is still missing. You can also find `DefaultDllImportSearchPaths` attributes in the client's source codes.

SadPencil (changes requested):
~~Your `CreateLinkToFile()` still do nothing on certain case. Please make sure this function either copy/linking files as intended, or throw an exception. Silent failures or skips are not acceptable.~~ hold on, you have wrapped the else statement inside an `#if`, never mind then.

DTAConfig/Settings/FileSourceDestinationInfo.cs
Metadorius: > An "ImmutableLink" option is also not a good idea because nor us developer or the modder should decide whether all users must use linking or not.

Why not though? Linking is a different behavior from copying.

SadPencil:
> Should you insist on this feature, I suggest you explicitly specify known text file extensions, and make sure mklink never acts on text files

I suggest you provide a way to whitelist/blacklist files that should be/never be linked. For example, never link .ini files. Or, only link .dll, .mix files.

DTAConfig/Settings/FileSourceDestinationInfo.cs
```diff
@@ -117,6 +118,19 @@ public void Apply()
                     File.Copy(SourcePath, DestinationPath, true);
                     break;
 
+                case FileOperationOptions.ImmutableLink:
+                    if (!File.Exists(DestinationPath))
+                    {
+                        FileInfo info = new FileInfo(SourcePath);
+                        info.IsReadOnly = true;
+                        CreateLinkToFile(sourcePath, destinationPath);
```
SadPencil, line 126: > A hard link is non-distinguishable from its source -- it's a file. Set both files as readonly, same as setting regular files

You can also set the destination file as read only

DTAConfig/Settings/FileSourceDestinationInfo.cs
```diff
@@ -140,6 +154,13 @@ public void Revert()
                     }
                     break;
 
+                case FileOperationOptions.ImmutableLink:
+                    FileInfo info = new FileInfo (SourcePath);
+                    info.IsReadOnly = false;
```
SadPencil, line 159: I think disabling readonly is unnecessary since the source file are meant to be readonly

SadPencil:
Well, I still felt kinds of awkward for your `CreateLinkToFile()` method. I'll make a PR to your repo

DTAConfig/Settings/FileSourceDestinationInfo.cs
SadPencil: > > An "ImmutableLink" option is also not a good idea because nor us developer or the modder should decide whether all users must use linking or not.
> 
> Why not though? Linking is a different behavior from copying.

Linking without a fallback succeeds on partial users and fails on other users.

SadPencil: > > An "ImmutableLink" option is also not a good idea because nor us developer or the modder should decide whether all users must use linking or not.
> 
> Why not though? Linking is a different behavior from copying.

Also, current file operation options includes: overwrite on mismatch, don't overwrite, etc. These are different behaviors about *whether* to copy. If we want to differ linking from copying, we need to add a new group of options. e.g., overwrite on mismatch with copying, overwrite on mismatch with linking

MahBoiDeveloper:
> Well, I still felt kinds of awkward for your `CreateLinkToFile()` method. I'll make a PR to your repo

Thanks, new implementation solve problem with  links in FAT32 and exFAT file systems.

ClientCore/FileHelper.cs
```diff
@@ -0,0 +1,93 @@
...
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                if (!CreateHardLink(destination, source, IntPtr.Zero))
+                    Marshal.ThrowExceptionForHR(Marshal.GetHRForLastWin32Error());
+            }
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux) || RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                if (link(source, destination) != 0)
+                    throw new IOException(string.Format("Unable to create hard link at {0} with the following error code: {1}".L10N("Client:DTAConfig:CreateHardLinkFailed"), destination, Marshal.GetLastWin32Error()));
```
Metadorius, line 85: Split that line

DTAConfig/DirectDrawWrapper.cs
```diff
@@ -163,12 +163,20 @@ public bool IsCompatibleWithOS(OSVersion os)
...
         {
+            string ddrawDllSourcePath = SafePath.CombineFilePath(ProgramConstants.GetBaseResourcePath(), ddrawDLLPath);
+            string ddrawDllTargetPath = SafePath.CombineFilePath(ProgramConstants.GamePath, "ddraw.dll");
+
             if (!string.IsNullOrEmpty(ddrawDLLPath))
             {
-                File.Copy(SafePath.CombineFilePath(ProgramConstants.GetBaseResourcePath(), ddrawDLLPath), SafePath.CombineFilePath(ProgramConstants.GamePath, "ddraw.dll"), true);
+                FileHelper.CreateHardLinkFromSource(ddrawDllSourcePath, ddrawDllTargetPath);
+                new FileInfo(ddrawDllSourcePath).IsReadOnly = false;
+                new FileInfo(ddrawDllTargetPath).IsReadOnly = false;
```
Metadorius, line 173: Oversight? should be true

DTAConfig/OptionPanels/DisplayOptionsPanel.cs
```diff
@@ -812,7 +812,7 @@ public override bool Save()
             IniSettings.Translation.Value = (string)ddTranslation.SelectedItem.Tag;
 
             // copy translation files to the game directory
-            foreach (TranslationGameFile tgf in ClientConfiguration.Instance.TranslationGameFiles)
+            ClientConfiguration.Instance.TranslationGameFiles.AsParallel().ForAll(tgf =>
```
Metadorius, line 815: Will this wait for all the tasks to complete?

Metadorius (changes requested):
seems to be good otherwise

DTAConfig/OptionPanels/DisplayOptionsPanel.cs
MahBoiDeveloper: Yes, as previous code.

DTAConfig/DirectDrawWrapper.cs
MahBoiDeveloper: Oops, yes, typo.

ClientCore/FileHelper.cs
MahBoiDeveloper: Done

Docs/INISystem.md
```diff
@@ -462,6 +462,7 @@ Valid file operation options available for files defined for `FileSettingCheckBo
 - `OverwriteOnMismatch`: Overwrites the destination file with source file only if they are different.
 - `DontOverwrite`: Never overwrites the destination file with source file if destination file is already present.
 - `KeepChanges`: Carries over the destination file with any changes manually made to by caching the file if deleted by disabling the option and then re-enabling it.
+- `LinkAsReadOnly`: Try to make a hard link to the source file, and if you get an error — copy the file. Recommended to use with any binary source files such as `opengl32.dll`, `d3d9.dll`, `dxgi.dll` and don't recommended use option with text files. While link is established, source file and target file has property `Read only` which protects original file and created link from edits. [More details](https://github.com/CnCNet/xna-cncnet-client/pull/590).
```
Metadorius, line 465:
```suggestion
- `LinkAsReadOnly`: Try to make a hard link (will look the same as the file but the content of the file will be shared) to the source file (copies the file as a fallback if the linking fails). Recommended to use with any binary source files such as `opengl32.dll`, `d3d9.dll`, `dxgi.dll` and not recommended to use with text files. While link is established, source file and target file has property `Read-only` which protects original file and created link from edits.
```

DTAConfig/Settings/FileSourceDestinationInfo.cs
```diff
@@ -162,6 +186,7 @@ public enum FileOperationOptions
         AlwaysOverwrite,
         OverwriteOnMismatch,
         DontOverwrite,
-        KeepChanges
+        KeepChanges,
+        LinkAsReadOnly
```
SadPencil, line 190: I have only one comment remains: suggest naming this new option as `AlwaysOverwrite_LinkAsReadOnly`

SadPencil (changes requested):

DTAConfig/Settings/FileSourceDestinationInfo.cs
MahBoiDeveloper: Done by d69c8f8.

SadPencil (approved):

PR #595, opened 2024-12-30
Add Steamworks SDK

GrantBartlett:
Registers the game as playing on steam.

Configurable inside ClientDefs, `SteamGameId=2229850`

Notes:
- Switched to [Facepunch ](https://wiki.facepunch.com/steamworks/Setting_Up)
- [Specific steam api dll needs to be used](https://partner.steamgames.com/downloads/steamworks_sdk_148a.zip): Issue here: https://github.com/Facepunch/Facepunch.Steamworks/issues/699#issuecomment-2047888712
- Added to `DXMainClient/Resources/steam_api64.dll` - Need clarification on where / how this should be copied to correct location

## Screenshots
image
image
image


ClientCore/ClientConfiguration.cs
```diff
@@ -184,6 +184,8 @@ public void RefreshSettings()
 
         public string LocalGame => clientDefinitionsIni.GetStringValue(SETTINGS, "LocalGame", "DTA");
 
+        public int SteamGameId => clientDefinitionsIni.GetIntValue(SETTINGS, "SteamGameId", 2229850);
```
Metadorius, line 187: I don't think this should be configured to the default value of YR one explicitly. you should check if it's a YR or Ares config via ifdef and use that as default, elif TS - use Steam TS one, else assign a missing ID value (idk if 0 is correct for Steam?) and check before even initializing.

Metadorius (changes requested):

Metadorius:
@MahBoiDeveloper please review where the docs for that should be put, if any at all?

DXMainClient/Startup.cs
```diff
@@ -153,6 +155,25 @@ public void Execute()
                 fhc.CalculateHashes();
             }
 #endif
+            // Load native steam binaries.
+            SteamNative.Initialize();
```
SadPencil, line 159: Also, does the Steam SDK work (or at least doesn't crash the client) in Linux? It's better to test it. If not (I don't know whether it works or not), an if statement should be introduced to skip the Steam API if the OS is not Windows.

ClientCore/ClientConfiguration.cs
GrantBartlett: image
image
What about this instead

Directory.Packages.props
```diff
@@ -25,6 +25,9 @@
     <PackageVersion Include="Rampastring.XNAUI.UniversalGL.Debug" Version="$(RampastringXNAUIVersion)" />
     <PackageVersion Include="Rampastring.XNAUI.WindowsXNA.Debug" Version="$(RampastringXNAUIVersion)" />
     <PackageVersion Include="SixLabors.ImageSharp" Version="2.1.9" />
+    <PackageVersion Include="SteamworksSharp" Version="1.0.8" />
+    <PackageVersion Include="SteamworksSharp.Native" Version="1.0.8" />
+    <PackageVersion Include="SteamworksSharp.Native.Windows_x64" Version="1.0.8" />
```
Metadorius, line 30: image
image

Uhh I don't think we should be using this. This is a pretty random wrapper wrote by a random person 6 years ago.

For a 1:1 Steamworks mappings use `Steamworks.NET`. For a preoprly C# one I think `Facepunch.Steamworks` should be used (based on some quick reading up I would pick that, even if it's outdated). Please research this question a bit more.

Metadorius (changes requested):

ClientCore/ClientConfiguration.cs
Metadorius: Why? It's not hard to change the default value for something sane based on the build config, and I don't think having am ability for people change steam app ID is something that should be exposed a lot. Besides, there are only two such games that are on Steam and are supported by the client - TS and YR.

MahBoiDeveloper:
> @MahBoiDeveloper please review where the docs for that should be put, if any at all?

It is not so complex feature like Discord Rich Presence, and documentation doesn't need for that, i suppose.

ClientCore/ClientConfiguration.cs
MahBoiDeveloper: > What about this instead

Why Steam Game Identityficator can be set from `ClientDefinitions.ini`? Is there possible situation when modder publish its mod in the Steam as the standalone game? If not -- better to hardcode YR and TS Steam IDs as Kerbiter said.

In addition to the above, i believe a button to enable/disable feature (like DRP) is needed.

MahBoiDeveloper (approved):

SadPencil (approved):

SadPencil:
Ah wait. The common DLL list

SadPencil (changes requested):
Update the common DLL list since new DLLs are introduced. https://github.com/CnCNet/xna-cncnet-client/issues/605

Saint146:
With this players can do "complete" cncnet steam integration, if they set a launch option in game's steam properties to something like this:
`"C:\Program Files\steam\steamapps\common\Command & Conquer Red Alert II\CnCNetYRLauncher.exe" %command%`
(obviously changing the path according to proper game and wherever they have it installed)

Maybe it can be useful to add this info to the website? For instance, to the guides https://cncnet.org/yuris-revenge/how-to-play, https://cncnet.org/red-alert-2/how-to-play, https://cncnet.org/tiberian-sun/how-to-play

SadPencil:
Must `steam_api64.dll` be placed into `Resources` folder, not `Resources\Binaires` folder?

SadPencil:
Another question: launcher the client in multi instance mode. Since they will initialize the Steam API with the same ID, we need to check at least no crashes in this situation.

SadPencil:
I just realized that maybe we need to re-license the client to GPL v3 with an exception that allowing the client linking with the steam api dll file since it's not compatible with GPL. https://www.gnu.org/licenses/gpl-faq.en.html#GPLIncompatibleLibs

SadPencil:
> @SadPencil this has been live in YR now for a while with no confirmed reported issues. Just an FYI :)

Can you test the situation below?

> launch the client in multi instance mode. Since they will initialize the Steam API with the same ID, we need to check at least no crashes in this situation.

DXMainClient/Startup.cs
```diff
@@ -156,6 +157,25 @@ public void Execute()
...
+                    SteamClient.Init(2229880);
+#endif
+                }
+                catch (System.Exception e)
+                {
+                    Logger.Log("Steam init failed: " + e.Message);
+                    // Couldn't init for some reason (steam is closed etc)
+                }
+            }
+#endif
```
MahBoiDeveloper, line 178: Why this is Windows only? What about Linux?

CommonAssemblies.txt
```diff
@@ -1,6 +1,7 @@
 ClientUpdater.dll
 ClientUpdater.pdb
 DiscordRPC.dll
+Facepunch.Steamworks.Win64.dll
```
MahBoiDeveloper, line 4: Does Steamworks have 32 bit version of their dll? Or 64 bit version only available?

DXMainClient/Startup.cs
SadPencil: limited time to investigate *shipping* and calling the steamworks library in linux

CommonAssemblies.txt
SadPencil: yes but it seems to act as a conflict package (`Facepunch.Steamworks`) with the win64 one

currently *we* only support win64 so we can directly use the `Facepunch.Steamworks` package with minimal changes. For a full support with multiple platforms and operating systems we have to build a special NuGet package shipping the native libraries from Steam as well as modifying this C# wrapper `Facepunch.Steamworks`. It's time consuming.

Metadorius: > we only support win64

What's the source on that claim? We build under either AnyCPU or 32-bit for XNA build (IIRC).

SadPencil: I mean we only support Windows x64 for the Steam integration for now, due to some restrictions on Facepunch.Steamworks as well as the native Steamworks dll files. We can of course spend a lot of time to build a better NuGet package providing support for multiple architectures while it's much more time consuming

Metadorius: Ah. Is it verified that the client still works in 32-bit mode, perhaps disabling steamworks integration then?

SadPencil:
> Ah. Is it verified that the client still works in 32-bit mode, perhaps disabling steamworks integration then?

@Metadorius Yes. The client will ignore the calling error.

图片

MahBoiDeveloper:
AFAIK Steam doesn't support Windows 7. Have you tested the x86 case with 32-bit Windows 10?

DXMainClient/Startup.cs
```diff
@@ -156,6 +157,25 @@ public void Execute()
...
+            if (UserINISettings.Instance.SteamIntegration)
+            {
+                try
+                {
+                    Logger.Log("Steam init called");
+#if YR || ARES
+                    SteamClient.Init(2229850);
+#else
+                    SteamClient.Init(2229880);
+#endif
```
Metadorius, line 170: This has to be updated with unified client build in mind

Metadorius (changes requested):
LGTM otherwise

SadPencil:
> AFAIK Steam doesn't support Windows 7. Have you tested the x86 case with 32-bit Windows 10?

No but it should behaves the same -- an exception is thrown because of invoking an amd64 dll in an x86-32 machine.

DXMainClient/Startup.cs
MahBoiDeveloper, line 170:
```suggestion
                if (ClientConfiguration.Instance.ClientGameType == ClientType.Ares || ClientConfiguration.Instance.ClientGameType == ClientType.YR)
                    SteamClient.Init(2229850);

                if (ClientConfiguration.Instance.ClientGameType == ClientType.TS)
                    SteamClient.Init(2229880);
#endif
```

SadPencil: `if TS` --> `else if TS` ?

MahBoiDeveloper: > `if TS` --> `else if TS` ?

Updated

DXMainClient/Startup.cs
```diff
@@ -156,6 +157,24 @@ public void Execute()
...
+#if ISWINDOWS
+            if (UserINISettings.Instance.SteamIntegration)
+            {
+                try
+                {
+                    Logger.Log("Steam init called");
+                if (ClientConfiguration.Instance.ClientGameType == ClientType.Ares || ClientConfiguration.Instance.ClientGameType == ClientType.YR)
+                    SteamClient.Init(2229850);
+                else if (ClientConfiguration.Instance.ClientGameType == ClientType.TS)
+                    SteamClient.Init(2229880);
```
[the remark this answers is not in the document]
SadPencil: Dismiss

PR #600, opened 2025-01-01
Fix game lobby being translated, bump protocol

GrantBartlett:


SadPencil (approved):

Metadorius (approved):

ClientCore/ProgramConstants.cs
```diff
@@ -26,7 +26,7 @@ public static class ProgramConstants
 
         public const string QRES_EXECUTABLE = "qres.dat";
 
-        public const string CNCNET_PROTOCOL_REVISION = "R10";
+        public const string CNCNET_PROTOCOL_REVISION = "R11";
```
Metadorius, line 29: Why?

Metadorius (changes requested):

ClientCore/ProgramConstants.cs
GrantBartlett: Its live right now, and not bumping it would mean other players would still see the game and join?

Metadorius: Well, it isn't the intended use case of protocol version but I guess that also works.

PR #613, opened 2025-01-03
Disable readonly property before deleting files

### LINKED ISSUE #614: Possible wrong implementation of `RenameAndMerge` in `(pre)updateexec` scripts
https://github.com/CnCNet/xna-cncnet-client/commit/fc939a06ff978b51daa6563eaa15a28cf48319ec#diff-0d2ea0a16eec1c4752818efaced94ca47df0f506a743e4407f0f6d667ac5daa0R894

图片

SadPencil:
Also, fixes #614

### DIFF
--- a/ClientUpdater/Updater.cs
+++ b/ClientUpdater/Updater.cs
@@ -916,12 +916,12 @@ public static class Updater
             {
                 Logger.Log("Updater: " + fileName + ": Renaming directory '" + key + "' to '" + newDirectoryName + "'");
 
-                DirectoryInfo directory = SafePath.GetDirectory(GamePath, key);
+                DirectoryInfo srcDirectory = SafePath.GetDirectory(GamePath, key);
 
-                // TODO
-
-                if (directory.Exists)
-                    directory.MoveTo(SafePath.CombineDirectoryPath(GamePath, newDirectoryName));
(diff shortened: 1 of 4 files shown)


ClientUpdater/Updater.cs
```diff
@@ -851,7 +855,13 @@ private static void ExecuteScript(string fileName)
...
             try
             {
-                SafePath.DeleteFileIfExists(GamePath, key);
+                FileInfo fileInfo = SafePath.GetFile(GamePath, key);
+
+                if (fileInfo.Exists)
+                {
+                    fileInfo.IsReadOnly = false;
+                    fileInfo.Delete();
+                }
```
MahBoiDeveloper, line 864: Instead of constantly checking `IsReadOnly` before deleting, it is better to modify the Rampastring.Tools class `SafePath` to support disabling the read-only property in delete methods.

MahBoiDeveloper (approved):

ClientUpdater/Updater.cs
SadPencil: I agree but let's deal with the emergency first

MahBoiDeveloper:
Changes doesn't have a negative effect on 2nd stage updater and client.

PR #624, opened 2025-01-05
Upgrade default CnCNet service URLs to HTTPS

GrantBartlett:


SadPencil:
This can be firstly tested by modifying networkdefinitions.ini file. Does the client work as intended?

GrantBartlett:
> This can be firstly tested by modifying networkdefinitions.ini file. Does the client work as intended?

Have tested, and have had no issues

SadPencil:
Great. Please globally search the source code since I remember these four URLs appear more than once in the codes (although they would be overridden by network definitions, we can still modify them as well)

GrantBartlett:
> Great. Please globally search the source code since I remember these four URLs appear more than once in the codes (although they would be overridden by network definitions, we can still modify them as well)

Yeah there were a couple, have updated

SadPencil (approved):

PR #630, opened 2025-01-08
Add map preview cache

### LINKED ISSUE #626: Map preview lagging/ crashing for some users
Need a crash log most likely and someone to double check the [GameInformationPanel ](https://github.com/CnCNet/xna-cncnet-client/blob/develop/DXMainClient/DXGUI/Multiplayer/GameInformationPanel.cs)changes.

SadPencil:


DXMainClient/DXGUI/Multiplayer/MapTextureCacheManager.cs
```diff
@@ -0,0 +1,75 @@
...
+            if (mapTextures.TryGetValue(map, out Image image))
+                return image;
+
+            if (missedMaps.Count < MaxCacheSize)
+                missedMaps.TryAdd(map, 0);
+
+            return null;
+        }
+
+        private async Task MapTextureLoadingService(CancellationToken cancellationToken)
```
Metadorius, line 41: don't think the functions should be called like so, the functions is "do something", not "a thing"

DXMainClient/DXGUI/Multiplayer/MapTextureCacheManager.cs
```diff
@@ -0,0 +1,75 @@
...
+namespace DTAClient.DXGUI.Multiplayer
+{
+    public class MapTextureCacheManager : IDisposable
+    {
+        public const int MaxCacheSize = 100;
+        public const int SleepIntervalMS = 100;
+
+        private readonly ConcurrentDictionary<Map, Image> mapTextures = [];
+
+        private readonly ConcurrentDictionary<Map, byte> missedMaps = [];
```
Metadorius, line 20: 1. Why is this a dictionary when this should be a set? If the reason is that there's no concurrent set it's better to use some empty type like `System.Reactive.Unit`, for instance, as it better conveys the purpose.
2. IMO this should be brought higher and renamed to `pendingMaps`, it's not `missing`, the maps are just waiting to be processed.

DXMainClient/DXGUI/Multiplayer/MapTextureCacheManager.cs
```diff
@@ -0,0 +1,75 @@
...
+
+                        missedMaps.TryRemove(missedMap, out _);
+
+                        if (mapTextures.ContainsKey(missedMap))
+                            continue;
+
+                        Image image = await Task.Run(missedMap.ExtractMapPreview);
+                        mapTextures.TryAdd(missedMap, image);
+                    }
+
```
Metadorius, line 67:
```suggestion
                    }
```

DXMainClient/DXGUI/Multiplayer/GameInformationPanel.cs
```diff
@@ -43,7 +43,7 @@ public GameInformationPanel(WindowManager windowManager, MapLoader mapLoader)
 
         private GenericHostedGame game = null;
 
-        private bool disposeTextures = false;
+        private bool disposeTextures = true;
```
Metadorius, line 46: This just caught my attention, I don't think there should be such a flag, just the no map preview texture should be set as a singleton, and then when disposing the texture should be compared to the no map preview one.

DXMainClient/DXGUI/Multiplayer/GameInformationPanel.cs
```diff
@@ -194,17 +194,28 @@ public void SetInfo(GenericHostedGame game)
...
+                if (map != null)
+                {
+                    if (map.IsPreviewTextureAvailableAsFile())
+                    {
+                        mapTexture = map.LoadPreviewTexture();
+                        disposeTextures = true;
+                    }
+                    else
+                    {
+                        mapTexture = AssetLoader.TextureFromImage(mapLoader.MapTextureCacheManager.GetMapTextureIfAvailable(map));
```
Metadorius, line 208: Does this mean the first show of the preview will always show null texture? When will it be updated? IMO there should be a callback that will update the info upon loading the map.

DXMainClient/DXGUI/Multiplayer/MapTextureCacheManager.cs
```diff
@@ -0,0 +1,75 @@
...
+        public const int SleepIntervalMS = 100;
+
+        private readonly ConcurrentDictionary<Map, Image> mapTextures = [];
+
+        private readonly ConcurrentDictionary<Map, byte> missedMaps = [];
+
+        private readonly CancellationTokenSource cancellationTokenSource = new();
+
+        public MapTextureCacheManager() =>
+            Task.Run(() => MapTextureLoadingService(cancellationTokenSource.Token));
```
Metadorius, line 25: There's no error handling and no task tracking. If your task errors out - there won't be any map previews anymore provided and the exception will not be caught.

Metadorius (changes requested):

DXMainClient/DXGUI/Multiplayer/GameInformationPanel.cs
SadPencil: The `disposeTextures` pattern is directly copied from `MapPreviewBox` class. In that class, `PreloadMapPreviews` is such a texture that should not be disposed. 

I think we can leave this design as a future work since the client already has such a pattern for a long time?

DXMainClient/DXGUI/Multiplayer/MapTextureCacheManager.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        public void Dispose() =>
+            cancellationTokenSource?.Cancel();
+
+        public Image GetMapTextureIfAvailable(Map map)
```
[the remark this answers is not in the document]
SadPencil: I will write `#nullable enable` on file `MapTextureCacheManager.cs`, if we decide to use nullable pattern for new files

DXMainClient/DXGUI/Multiplayer/GameInformationPanel.cs
Metadorius: I thought that's a trivial change to do so though, isn't that right?

DXMainClient/DXGUI/Multiplayer/MapTextureCacheManager.cs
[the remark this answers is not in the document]
Metadorius: IMO nullables should be enabled across all projects.

DXMainClient/DXGUI/Multiplayer/MapTextureCacheManager.cs
```diff
@@ -0,0 +1,75 @@
...
+        {
+            if (mapTextures.TryGetValue(map, out Image image))
+                return image;
+
+            if (missedMaps.Count < MaxCacheSize)
```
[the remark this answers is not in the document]
SadPencil: Limiting RAM usage?

DXMainClient/DXGUI/Multiplayer/GameInformationPanel.cs
SadPencil: Should be but it might require additional tests. Will trying migrating it after I am available again

DXMainClient/DXGUI/Multiplayer/GameInformationPanel.cs
SadPencil: Yes. Currently it will be updated the second time it is visited.

Will trying implementing a callback after I am available again (~weeks perhaps)

DXMainClient/DXGUI/Multiplayer/MapTextureCacheManager.cs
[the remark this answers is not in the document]
Metadorius: IIRC the main complaint was too strict StyleCop warnings, but alright. Those warnings are actually useful in this case IMO.

DXMainClient/DXGUI/Multiplayer/MapTextureCacheManager.cs
SadPencil: So what do you suggest? I think we can't introduce RX here right now just because of introducing the unit type

DXMainClient/DXGUI/Multiplayer/MapTextureCacheManager.cs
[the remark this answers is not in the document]
SadPencil: I know it's helpful. The point is there are too many warnings that would draw our attention from new warnings introduced

DXMainClient/DXGUI/Multiplayer/MapTextureCacheManager.cs
SadPencil: will fix it

DXMainClient/DXGUI/Multiplayer/MapTextureCacheManager.cs
SadPencil: Will fix it by catching and recording exceptions in the while loop

DXMainClient/DXGUI/Multiplayer/MapTextureCacheManager.cs
SadPencil: huh? No changes were made

Metadorius: Ugh, I tried to remove the extra newline. Seems GitHub ate that.

DXMainClient/DXGUI/Multiplayer/MapTextureCacheManager.cs
Metadorius: Well, I don't see a big problem with introducing Rx, but I guess you could make an empty type or just copy the unit type from `System.Reactive`, and later we could just replace it when we introduce Rx.

DXMainClient/DXGUI/Multiplayer/GameInformationPanel.cs
```diff
@@ -194,17 +194,28 @@ public void SetInfo(GenericHostedGame game)
...
+                        disposeTextures = true;
```
Metadorius, line 209: I am not sure you should dispose the cached textures though?

SadPencil: Otherwise it would occupy VRAMs for those map previews. And a lot of users have limited VRAM. So only the bitmap is cached in RAM

DXMainClient/DXGUI/Multiplayer/MapTextureCacheManager.cs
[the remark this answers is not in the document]
Metadorius: You missed the point. The RAM usage regulation is going to happen in the main dictionary which actually holds the map previews. AFAIK the map list is already in memory, so there wouldn't be a point to also regulate this as the most it will save is a few bytes. Also I don't think hitting 100 maps limit in 100ms is possible at all, and the "incoming" dictionary is drained regularly.

DXMainClient/DXGUI/Multiplayer/GameInformationPanel.cs
Metadorius: Oh right, it's not just a copy, nvm then.

DXMainClient/DXGUI/Multiplayer/MapTextureCacheManager.cs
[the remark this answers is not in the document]
SadPencil: Then do you suggest also removing `MaxCacheSize` limits for the dictionary that holds bitmaps or not?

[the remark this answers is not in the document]
Metadorius: No, only from the pending/missing maps.

Metadorius (commented):
Actually hold on, maybe it would be better to use something like a weak reference for the cache instead of manually doing it?

Also I am not sure if clearing all of the maps when the limit is reached is a good idea, I would shave off the not used ones. Some maps are going to appear more frequently so they won't be deleted.

SadPencil:
> Actually hold on, maybe it would be better to use something like a weak reference for the cache instead of manually doing it?
> 
> Also I am not sure if clearing all of the maps when the limit is reached is a good idea, I would shave off the not used ones. Some maps are going to appear more frequently so they won't be deleted.

Thanks for letting me know `WeakReference`. Will investigate

pzhlkj6612:
> ... I would shave off the not used ones. Some maps are going to appear more frequently so they won't be deleted.

LRU.

SadPencil:
> > ... I would shave off the not used ones. Some maps are going to appear more frequently so they won't be deleted.
> 
> LRU. 

I would prefer leaving GC to decide using WeakReference

Metadorius:
No, he has a point. GC doesn't care about time of last use, but we do. LRU cache seems to be a common technique and an effective measure for keeping only actually used things in memory, so why not?

SadPencil:
> No, he has a point. GC doesn't care about time of last use, but we do. LRU cache seems to be a common technique and an effective measure for keeping only actually used things in memory, so why not?

Okay. I just searched and find out WeakReference will be cleared out at every collection. Will implement an LRU

PR #633, opened 2025-01-09
Modify GameInformationPanel to optionally display as an invite

11EJDE11:
Currently when you receive an invite to join a game, you can only get information about the game host and game name. This PR modifies game invite requests to show as a GameInformationPanel instead of ChoiceNotificationBox so you get the nice map thumbnail and all the other information that is normally shown about a game.

It adds Accept/Decline buttons to the GameInformationPanel which are visible if [panel].IsInvite is set.

image

First timer here so apologies if I am going about this incorrectly.

### DIFF
--- a/DXMainClient/DXGUI/Multiplayer/GameInvitePanel.cs
+++ b/DXMainClient/DXGUI/Multiplayer/GameInvitePanel.cs
@@ -1,0 +1,113 @@
+﻿using System;
+using System.Diagnostics;
+using System.Windows.Forms;
+
+using ClientCore;
+using ClientCore.Extensions;
+
+using ClientGUI;
+
+using DTAClient.Domain.Multiplayer;
+
+using Microsoft.Xna.Framework;
+using Microsoft.Xna.Framework.Graphics;
+
+using Rampastring.XNAUI;
+using Rampastring.XNAUI.XNAControls;
+
+namespace DTAClient.DXGUI.Multiplayer
+{
+    /// <summary>
+    /// A UI panel that displays information about a hosted CnCNet or LAN game.
+    /// </summary>
+    public class GameInvitePanel : XNAPanel
+    {
+        public GameInvitePanel(WindowManager windowManager, MapLoader mapLoader)
+            : base(windowManager)
+        {
+            this.mapLoader = mapLoader;
+            DrawMode = ControlDrawMode.UNIQUE_RENDER_TARGET;
+        }
+
+        private MapLoader mapLoader;
+
+        private GenericHostedGame game = null;
+
+        private const int buttonWidth = 75;
+        private const int buttonHeight = 30;
+        private const int padding = 10;
+
+        private XNALabel lblInviteHeading;
+        private XNAClientButton btnInviteAccept;
+        private XNAClientButton btnInviteDecline;
+        private GameInformationPanel panelGameInformation;
+
+        public event Action AcceptInvite;
+        public event Action DeclineInvite;
+
+        public override void Initialize()
+        {
+            panelGameInformation = new GameInformationPanel(WindowManager, mapLoader);
+            panelGameInformation.Name = nameof(panelGameInformation);
+            panelGameInformation.BackgroundTexture = AssetLoader.LoadTexture("cncnetlobbypanelbg.png");
+            panelGameInformation.DrawMode = ControlDrawMode.UNIQUE_RENDER_TARGET;
+            panelGameInformation.Initialize();
+            panelGameInformation.ClearInfo();
+            panelGameInformation.Disable();
+            panelGameInformation.InputEnabled = false;
+            panelGameInformation.Alpha = 0.5f;
+            panelGameInformation.AlphaRate = 0.5f;
+            AddChild(panelGameInformation);
+
+            lblInviteHeading = new XNALabel(WindowManager);
+            lblInviteHeading.FontIndex = 1;
+            lblInviteHeading.Text = "INVITATION TO JOIN GAME".L10N("Client:Main:InviteHeading");
+            lblInviteHeading.X = (panelGameInformation.Width / 2) - (lblInviteHeading.Width/2);
+            lblInviteHeading.Y = ((lblInviteHeading.Height + padding) / 2) - (lblInviteHeading.Height/2);
+
+            ClientRectangle = new Rectangle(0, 0, panelGameInformation.Width, padding + lblInviteHeading.Height  + panelGameInformation.Height  + padding + buttonHeight + padding);
+            BackgroundTexture = AssetLoader.CreateTexture(new Color(0, 0, 0, 255), 1, 1);
+            PanelBackgroundDrawMode = PanelBackgroundImageDrawMode.STRETCHED;
+
+            AddChild(lblInviteHeading);
+            lblInviteHeading.Visible = true;
+            panelGameInformation.Y = lblInviteHeading.Height + padding;
+
+            btnInviteAccept = new XNAClientButton(WindowManager);
+            btnInviteAccept.Text = "Accept".L10N("Client:Main:InviteAccept");
+            btnInviteAccept.ClientRectangle = new Rectangle(ClientRectangle.Width / 2 - buttonWidth - (padding / 2), panelGameInformation.Y + panelGameInformation.Height + padding, buttonWidth, buttonHeight);
+            btnInviteAccept.LeftClick += (s, e) => AcceptInvite?.Invoke();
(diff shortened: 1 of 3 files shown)


Metadorius:
Transmitting this on behalf of @Rampastring who doesn't have access to GitHub at the moment:

> Could be done better by creating a game invitation panel/window that has the game information panel as a child

11EJDE11:
Thanks for the feedback Saint, Meta, and Rampa. A game invitation panel makes much more sense. I have made the requested changes. 

A bug has also been fixed where if you accept an invite to a game lobby that you are already a part of, it will no longer leave lobby and attempt to join again. Previously this would quite often fail.

There is a slight change to the layout from my last commit:
image

DXMainClient/DXGUI/Multiplayer/GameInvitePanel.cs
```diff
@@ -0,0 +1,113 @@
...
+            btnInviteDecline.Visible = true;
+            btnInviteDecline.Name = "btnInviteDecline";
+            AddChild(btnInviteDecline);
+            base.Initialize();
+        }
+
+        public void SetInfo(GenericHostedGame game)
+        {
+
+            this.game = game;
```
Metadorius, line 97:
```suggestion
        {
            this.game = game;
```

DXMainClient/DXGUI/Multiplayer/GameInvitePanel.cs
```diff
@@ -0,0 +1,113 @@
...
+            btnInviteAccept.Visible = true;
+            btnInviteAccept.Name = "btnInviteAccept";
+            AddChild(btnInviteAccept);
+
+            btnInviteDecline = new XNAClientButton(WindowManager);
+            btnInviteDecline.Text = "Decline".L10N("Client:Main:InviteDecline");
+            btnInviteDecline.ClientRectangle = new Rectangle(ClientRectangle.Width / 2 + (padding / 2), btnInviteAccept.Y, buttonWidth, buttonHeight);
+            btnInviteDecline.LeftClick += (s, e) => DeclineInvite?.Invoke();
+            btnInviteDecline.Visible = true;
+            btnInviteDecline.Name = "btnInviteDecline";
```
Metadorius, line 89: Could use `nameof` here

DXMainClient/DXGUI/Multiplayer/GameInvitePanel.cs
```diff
@@ -0,0 +1,113 @@
...
+            panelGameInformation.AlphaRate = 0.5f;
+            AddChild(panelGameInformation);
+
+            lblInviteHeading = new XNALabel(WindowManager);
+            lblInviteHeading.FontIndex = 1;
+            lblInviteHeading.Text = "INVITATION TO JOIN GAME".L10N("Client:Main:InviteHeading");
+            lblInviteHeading.X = (panelGameInformation.Width / 2) - (lblInviteHeading.Width/2);
+            lblInviteHeading.Y = ((lblInviteHeading.Height + padding) / 2) - (lblInviteHeading.Height/2);
+
+            ClientRectangle = new Rectangle(0, 0, panelGameInformation.Width, padding + lblInviteHeading.Height  + panelGameInformation.Height  + padding + buttonHeight + padding);
```
Metadorius, line 68:
```suggestion
            ClientRectangle = new Rectangle(0, 0, panelGameInformation.Width, padding + lblInviteHeading.Height  + panelGameInformation.Height + padding + buttonHeight + padding);
```

DXMainClient/DXGUI/Multiplayer/GameInvitePanel.cs
```diff
@@ -0,0 +1,113 @@
...
+            DrawMode = ControlDrawMode.UNIQUE_RENDER_TARGET;
+        }
+
+        private MapLoader mapLoader;
+
+        private GenericHostedGame game = null;
+
+        private const int buttonWidth = 75;
+        private const int buttonHeight = 30;
+        private const int padding = 10;
```
Metadorius, line 38: Can't recall if there are already constants for those globally, please check.

DXMainClient/DXGUI/Multiplayer/GameInvitePanel.cs
```diff
@@ -0,0 +1,113 @@
...
+    public class GameInvitePanel : XNAPanel
+    {
+        public GameInvitePanel(WindowManager windowManager, MapLoader mapLoader)
+            : base(windowManager)
+        {
+            this.mapLoader = mapLoader;
+            DrawMode = ControlDrawMode.UNIQUE_RENDER_TARGET;
+        }
+
+        private MapLoader mapLoader;
```
Metadorius, line 32:
```suggestion
        private readonly MapLoader mapLoader;
```

Metadorius (changes requested):
Looking good so far, I have a couple of questions though.

1. Have you checked whether you can customize it through the INI?
2. Not sure what special logic you have for alpha?
3. Is the panel updated on game changes if the user doesn't accept the invite for long enough?
   - Not necessarily in scope of this PR, but what happens if the game gets abandoned and the window is still open?

I also suggest to perhaps remove borders from the inner panel somehow and re-add information on who invited you.

pzhlkj6612:
Hey, @11EJDE11! A pro tip about GitHub: 

> To quickly incorporate more than one suggested change into a single commit, you can also apply suggested changes as a batch. ...
> 
> *from: [Reviewing proposed changes in a pull request - GitHub Docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request)*

11EJDE11:
Thanks for the review, Metadorius.

1) Regarding the INIs - not too sure if they can be customised through there. I'm not calling ReadINIForControl so I assume not? I don't believe it was doing anything with the INIs before though either.

2) Alpha stuff was a leftover. I've removed it now.

3.1) I have made a change so the panel will update when the game information updates.

3.2) The invite will disappear if the game has closed or the host is offline.

I've removed the border from the inner panel and have added the invitee's name back in (note: invites can only be from hosts and the host name is already on the inner panel).

Also I am new to Github and all this jazz - do I click Resolve Conversations on the ones above? I have switched to nameof as you suggested, and I am now using the button constants for size. There is no constant I could see for the padding amount so I have copied the amount from another area that had padding.

And thanks pzhlkj6612 - I could have sworn that was greyed out! Please keep the tips coming; I need them. Pretty sure I've done the last commit a bit weird. Not sure why there is a second one for changing the map to readonly. When I pushed it said I was behind so had to update to the latest branch. Did that and there was a conflict, chose the newer/local copy and then it put that one line into another commit. Pushed that and here we are.

New panel:
image

Metadorius:
@11EJDE11 You're welcome!

> Regarding the INIs - not too sure if they can be customised through there. I'm not calling ReadINIForControl so I assume not? I don't believe it was doing anything with the INIs before though either.

This should be done. Generally the point is to have all of the client customizable since there are many users (as in modders and game maintainers) who use it and they all have different customization needs. , For example, look at GameCreationWindow -- it is customizable via INI.

I am not sure if I am a fan of combining title that is all in caps and the username.

> do I click Resolve Conversations on the ones above? 

Yeah.

> There is no constant I could see for the padding amount so I have copied the amount from another area that had padding.

I propose to introduce one. It is used in GameInformationPanel as well.

> I could have sworn that was greyed out!

That's the GitHub unintuitivity for you. It is only available in the files tab.

pzhlkj6612:
> > Regarding the INIs ...
> 
> ... who use it and they all have different customization needs. ...

Another example is, I'd like to hide the "GAME INFORMATION" line in the dialog. I can achieve it easily if you provide INI-ish way to customize it.

DXMainClient/DXGUI/Multiplayer/GameInvitePanel.cs
11EJDE11: Yes, fixed width and height. Padding will come in another commit/PR as there is no constant currently.

DXMainClient/DXGUI/Multiplayer/GameInvitePanel.cs
11EJDE11: Done, thanks.

Metadorius:
Any progress on this?

11EJDE11:
Will take another look.

SadPencil:
Might have a conflict with https://github.com/CnCNet/xna-cncnet-client/issues/977 so I removed the milestone for now

PR #634, opened 2025-01-09
Fix skirmish lobby selecting wrong map when the list is sorted

Saint146:
Not the most optimal but fast fix of the small issue that annoyed me for months.
Currently, if you turn on sorting of the map list in the skirmish lobby, select some map and restart the client, the map changes to a completely different one. If the sorting is off, selected map is kept between restarts correctly.


Saint146:
Well the map list is being sorted when adding to the listbox, but when searching for map we got the index from the unsorted list.
That led to the wrong map being selected. You can easily reproduce it in the current version

SadPencil (approved):

Metadorius (approved):

PR #637, opened 2025-01-11
Add map thumbnail preview to multiplayer game lobby map list hover

11EJDE11:
Still a bit of a draft but figured I'd submit now and get some initial feedback.

- Do others want this, or just me?
- Would this be better behind a toggle? For example, right click the map list and choose "Show previews"
- Add it to skirmish map list box too?
- Is there other information that should show?

Like the other recent PR from Grant that adds map previews, this preview is only shown for maps with the texture file. The reason for this was lag on large map previews. I've never dealt with threading before - do you think it's possible to load the preview in another thread in order to have all previews working rather than those with just the texture file?

**Here it is currently:**
XNAMapPreviewGameLobby

### DIFF
--- a/DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
+++ b/DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
@@ -17,6 +17,7 @@ using DTAClient.DXGUI.Multiplayer.CnCNet;
 using DTAClient.Online.EventArguments;
 using ClientCore.Extensions;
 using TextCopy;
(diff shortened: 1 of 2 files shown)


pzhlkj6612:
Looks cool!

> * Do others want this, or just me?

I want this as a player who is always lost in the map list.

<br>

It would be nice if we highlighted the map with a dark color (configurable) where the mouse is hovering.

CnC-RaVaGe:
You could just display a temp full-size preview over the current map preview as well, especially with your map preview title, would look cleaner and a larger preview is always better.

PR #638, opened 2025-01-12
Fix issues when running 2.11.0.0+ in Wine.

remgodow:
This PR fixes two issues occuring after upgrading YR client to 8.65+ when running in Wine (Linux).

.NET Framework 4.8 version hangs in WMIC when generating online id. Changing from async function to a manual thread (like in CheckSystemSpecifications()) solves the issue.

.NET 8 version throws SecurityException which crashes the app when checking directory permissions. Catching the exception solves the issue.

From SteamOS/Linux user perspective it is easier to keep using Windows executables, which worked up until 2.11.0.0, so it would be nice if these fixes could be merged.

Thanks.

### DIFF
--- a/DXMainClient/PreStartup.cs
+++ b/DXMainClient/PreStartup.cs
@@ -17,6 +17,7 @@ using System.Runtime.InteropServices;
 using System.Runtime.Versioning;
 using ClientCore.I18N;
 using System.Globalization;
+using System.Security;
 using System.Transactions;
 
 namespace DTAClient
@@ -345,11 +346,18 @@ namespace DTAClient
                         if (ntAccount == null)
                             continue;
 
-                        if (principal.IsInRole(ntAccount.Value))
+                        try
                         {
-                            if (fsAccessRule.AccessControlType == AccessControlType.Deny)
-                                return false;
-                            isInRoleWithAccess = true;
+                            if (principal.IsInRole(ntAccount.Value))
+                            {
+                                if (fsAccessRule.AccessControlType == AccessControlType.Deny)
+                                    return false;
+                                isInRoleWithAccess = true;
+                            }
+                        }
+                        catch (SecurityException)
+                        {
+                            //IsInRole may throw when running in Wine
                         }
                     }
                 }
--- a/DXMainClient/PreStartup.cs
+++ b/DXMainClient/PreStartup.cs
@@ -357,7 +357,8 @@ namespace DTAClient
                         }
                         catch (SecurityException)
                         {
-                            //IsInRole may throw when running in Wine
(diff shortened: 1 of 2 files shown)


Metadorius (approved):

DXMainClient/Startup.cs
```diff
@@ -62,7 +62,8 @@ public void Execute()
                 thread.Start();
             }
 
-            GenerateOnlineIdAsync();
```
SadPencil, line 65: Could you explain why making this method `GenerateOnlineIdAsync` synchronous again?

DXMainClient/PreStartup.cs
```diff
@@ -345,11 +346,18 @@ private static bool UserHasDirectoryAccessRights(string path, FileSystemRights a
...
+                            {
+                                if (fsAccessRule.AccessControlType == AccessControlType.Deny)
+                                    return false;
+                                isInRoleWithAccess = true;
+                            }
+                        }
+                        catch (SecurityException)
+                        {
+                            //IsInRole may throw when running in Wine
                         }
```
SadPencil, line 361: Make the code straightforward. Add a return statement in the catch block instead of relying on the final return statement which is too far away and wastes people more time to figure out the fact that `isInRoleWithAccess` will not be further modified during line 360 -- 369.

SadPencil (changes requested):

SadPencil:
LGTM except adding an early return in the catch block

DXMainClient/PreStartup.cs
remgodow: `isInRoleWithAccess` may be modified in the next `foreach (AurhorizationRole role in roles)` iteration, which is why I did not use an early return. I thought about adding `continue`, but it generates a warning about redundant code, so I opted for a comment insted.
If you think that an explicit `continue` statement would improve code readability, I can add it.

SadPencil: Thank you for clarifying this. Either an explicit `continue` or a comment line in the catch block should be fine.

remgodow: Done.

DXMainClient/Startup.cs
```diff
@@ -62,7 +62,8 @@ public void Execute()
...
+            Thread onlineIdThread = new Thread(GenerateOnlineId);
+            onlineIdThread.Start();
```
Metadorius, line 66:
```suggestion
            // Using tasks here causes crashes on Wine for some reason
            Thread onlineIdThread = new Thread(GenerateOnlineId);
            onlineIdThread.Start();
```

PR #642, opened 2025-01-21
Fix parsing error with IPv6 addresses in CnCNetTunnel

11EJDE11:
I saw in my logs a parsing error when encountering tunnels with IPV6 addresses. I am assuming the fix is to parse these correctly rather than discard tunnels with IPV6 addresses (I haven't tried connecting to any so don't know if they cause errors). 

For reference, the error that was appearing in my logs:
21.01. 19:30:52.509    Fetching tunnel server info.
21.01. 19:30:52.824    Parsing tunnel information failed: System.FormatException: Input string was not in a correct format.
   at System.Number.StringToNumber(String str, NumberStyles options, NumberBuffer& number, NumberFormatInfo info, Boolean parseDecimal)
   at System.Number.ParseInt32(String s, NumberStyles style, NumberFormatInfo info)
   at DTAClient.Domain.Multiplayer.CnCNet.CnCNetTunnel.Parse(String str) in C:\Dev\xna-cncnet-client\DXMainClient\Domain\Multiplayer\CnCNet\CnCNetTunnel.cs:line 39
Parsed string: 240e:33d:343:bf80::1017:50000;China;CN;yuki;0;0;200;0;0;0;2;0

### DIFF
--- a/DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetGameLoadingLobby.cs
+++ b/DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetGameLoadingLobby.cs
@@ -16,6 +16,7 @@ using Rampastring.XNAUI.XNAControls;
 using System;
 using System.Collections.Generic;
 using System.Text;
+using System.Linq;
 
 namespace DTAClient.DXGUI.Multiplayer.CnCNet
 {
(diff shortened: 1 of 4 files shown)


11EJDE11:
I've made some additional fixups. At the limit of what I can do now as IPV6 in general doesn't work for me.

Metadorius:
Does the tunnel get ignored if it's IPv6? AFAIK IPv6 tunnels are not supported by our spawner ATM.

11EJDE11:
Yeah it gets ignored. You can set a game to an IPV6 tunnel but it won't launch (you get prompt to change the tunnel) and won't show for other players (afaik because it fails with the parsing).

Metadorius:
I don't think it should be shown then if it can't be used. It's a bad UX to show stuff that is not usable in this case.

11EJDE11:
Sorry, forgot IPV6 doesn't work for me in general - that'd be why I get the prompt to change tunnel. Perhaps it would work for others with this PR.

Metadorius:
> Sorry, forgot IPV6 doesn't work for me in general - that'd be why I get the prompt to change tunnel. Perhaps it would work for others with this PR.

Point is, from my knowledge none of the games support IPv6, if an IPv6 tunnel is happened to be in master list - currently that's an error and it should probably be discarded and logged.

PR #643, opened 2025-01-21
Add optional parameter to [ForcedOptions] to make changes optional

11EJDE11:
Currently with maps you can force options with:

[ForcedOptions]
chkNoSpy=true
chkShortGame=false
chkBalancePatch=true
...etc

Doing this will force the checkboxes to the value provided (true=ticked, false=unticked) **_and_** lock the checkbox (checkBox.AllowChecking = False).

This PR adds an optional second parameter that sets the value for checkbox.AllowChecking

So essentially it allows maps to force options, but still allows hosts to change them.

[ForcedOptions]
chkNoSpy=true,false
chkShortGame=false
chkBalancePatch=true,true
...etc

image

### DIFF
--- a/DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
+++ b/DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
@@ -2235,6 +2235,7 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
 
             PlayerExtraOptionsPanel?.UpdateForMap(Map);
         }
+        
         private void ApplyForcedCheckBoxOptions(List<GameLobbyCheckBox> optionList,
     List<KeyValuePair<string, string>> forcedOptions)
         {
@@ -2251,7 +2252,6 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
                 }
             }
         }
-
 
         private void ApplyForcedDropDownOptions(List<GameLobbyDropDown> optionList,
             List<KeyValuePair<string, int>> forcedOptions)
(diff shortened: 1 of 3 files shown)


11EJDE11:
Also it was user "786" on Discord that requested something like this in ra2yr-updates-chat about a week ago.

DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
```diff
@@ -2235,22 +2235,24 @@ protected virtual void ChangeMap(GameModeMap gameModeMap)
...
-                    checkBox.AllowChecking = false;
+                    checkBox.Checked = bool.Parse(optionParts[0]);
+                    checkBox.AllowChecking = optionParts.Length > 1 ? bool.Parse(optionParts[1]) : false;
+
                     optionList.Remove(checkBox);
                 }
             }
         }
 
+
```
SadPencil, line 2255: This line break should be moved in front of `private void ApplyForcedCheckBoxOptions`

SadPencil (changes requested):
Besides a coding style issue, others are fine

SadPencil (approved):

Metadorius (changes requested):
IMO piggybacking off `ForcedOptions` is unintuitive. Why not make an identical section `RecommendedOptions`?

Also what would be the behavior when a user switches to a map with recommended options, then switches to another map where there are no recommended options? Will those options be kept?

Also I don't think subtly changing options without any indication (force option has a clear indication that it makes the checkbox unavailable) is good. It will confuse the user.

PR #647, opened 2025-01-30
Add L10N to the components in the XNAMessageBox

MahBoiDeveloper:
As MarkJFox mentioned, there is a bug where the client doesn't show the translated component name in the install message boxes. PR should fix this.

<details>
<summary>Before:</summary>

image
image

</details>

<details>
<summary>After:</summary>

Снимок экрана 2025-01-30 173133
Снимок экрана 2025-01-30 173309
Снимок экрана 2025-01-30 173426

</details>


SadPencil (approved):

Metadorius (commented):
Why translate every time instead of translating the source field/property once?

MahBoiDeveloper:
> Why translate every time instead of translating the source field/property once?

Alternative solution is to use `.L10N` in `CustomComponent` class, but it gives error because it cannot use string extensions in `ClientUpdater` project and include other client's namespaces. I don't know how to fix such dependencies and decided to use simpler idea.

Metadorius (approved):

PR #658, opened 2025-02-04
Disable chat message audio when game is running

11EJDE11:
As discussed somewhere on Discord.


SadPencil (approved):

Metadorius:
I am not sure that is a proper solution. You're trying to solve the problem of someone's spam coming through and repeatedly making sounds AFAIK, but you're also screwing up legit players chatting in the room when they, for example, didn't launch for some reason.

You should instead deal with the core issue, that is allowing to spam too many messages in a quick succession.

11EJDE11:
Understood, yeah it is a rather pedestrian fix. I still think something like this could be wanted by some. The lag with IRC can be bad. There's also the case where two users get out of the game early and sit in the lobby talking; all remaining players in game can hear the beeps.

Metadorius:
> There's also the case where two users get out of the game early and sit in the lobby talking;

Is this a frequent/valid case? Genuine question since all my client exposure is from playing with a group of friends and not in FFA, so such never happened to me.

11EJDE11:
We normally do 3v3 and 4v4 with a friendly bunch of lads so it's more common for us I would imagine.

Also to your point earlier - If the game fails to launch (anti virus blocking it for instance), or they quit out on the loading screen, the code to enable the audio will still run. I can't think of any cases where it wouldn't be enabled again.

Metadorius:
How about a toggle in client options to control whether the chat sound is muted while in-game then?

11EJDE11:
Sounds good to me. I'll look at into it when I get a mo.

11EJDE11:
I've added a setting in. Figured while I was in there I should add some constants for control positioning - hope that's OK.

Before/After with these changes:
image

DTAConfig/OptionPanels/AudioOptionsPanel.cs
```diff
@@ -10,6 +10,20 @@ namespace DTAConfig.OptionPanels
 {
     class AudioOptionsPanel : XNAOptionsPanel
     {
+        private const int VOLUME_MIN = 0;
+        private const int VOLUME_MAX = 10;
+        private const float VOLUME_SCALE_FLOAT = 10.0f;
```
SadPencil, line 15: It should be okay always using double. We are not programming in GPU so there's no situation requires a float.

SadPencil: I just realized that the UI library requires a float instead of a double as the paramater. Never mind then.

DTAConfig/OptionPanels/AudioOptionsPanel.cs
```diff
@@ -139,27 +155,36 @@ public override void Initialize()
...
             chkMainMenuMusic.CheckedChanged += ChkMainMenuMusic_CheckedChanged;
+            AddChild(chkMainMenuMusic);
 
             chkStopMusicOnMenu = new XNAClientCheckBox(WindowManager);
             chkStopMusicOnMenu.Name = "chkStopMusicOnMenu";
-            chkStopMusicOnMenu.ClientRectangle = new Rectangle(
-                lblScoreVolume.X, chkMainMenuMusic.Bottom + 24, 0, 0);
             chkStopMusicOnMenu.Text = "Don't play main menu music in lobbies".L10N("Client:DTAConfig:NoLobbiesMusic");
+            chkStopMusicOnMenu.ClientRectangle = new Rectangle(
+                lblScoreVolume.X, chkMainMenuMusic.Y + chkMainMenuMusic.Height + CHECKBOX_SPACING, 0, 0);
```
Metadorius, line 179:
```suggestion
                lblScoreVolume.X, chkMainMenuMusic.Bottom + CHECKBOX_SPACING, 0, 0);
```

DTAConfig/OptionPanels/AudioOptionsPanel.cs
```diff
@@ -10,6 +10,20 @@ namespace DTAConfig.OptionPanels
 {
     class AudioOptionsPanel : XNAOptionsPanel
     {
+        private const int VOLUME_MIN = 0;
+        private const int VOLUME_MAX = 10;
+        private const float VOLUME_SCALE_FLOAT = 10.0f;
+        private const int VOLUME_SCALE_INT = 10;
+        private const double VOLUME_SCALE_DOUBLE = 10;
```
Metadorius, line 17: Could do

```suggestion
        private const int VOLUME_SCALE = 10;
```
and then just use `/ (float)VOLUME_SCALE_INT` and `/ (double)VOLUME_SCALE_INT` where needed.

Metadorius (changes requested):
Looks good apart from few minor things

SadPencil (approved):

Metadorius (approved):

DXMainClient/DXGUI/Multiplayer/GameLobby/MultiplayerGameLobby.cs
```diff
@@ -264,9 +264,6 @@ protected override void StartGame()
             if (fsw != null)
                 fsw.EnableRaisingEvents = true;
 
-            for (int pId = 0; pId < Players.Count; pId++)
-                Players[pId].IsInGame = true;
-
```
SadPencil: Removed in https://github.com/CnCNet/xna-cncnet-client/pull/680

PR #659, opened 2025-02-05
Fix concurrency issue in LoadCustomMaps by using a thread-safe collection

11EJDE11:
Had an "IndexOutOfRangeException" today on **localMapSHAs.Add(map.SHA1);** and I think this is the fix for it. 
I don't know anything when it comes to threading so hopefully one of the propeller heads around here can verify this.


SadPencil (approved):
Looks like the order of elements does not matter in `localMapSHAs`.

SadPencil:
> I don't know anything when it comes to threading so hopefully one of the propeller heads around here can verify this.
Indeed, when concurrency issue occurs, the exception message will be confusing. And since `localMapSHAs` is added by a task (either multiple writes, or read+write could have a concurrentcy issue), we should indeed use concurrent collections here

11EJDE11:
Thanks for the explanation, SadPencil. Setting a breakpoint on one of those lines blew my mind.

PR #665, opened 2025-02-13
Profanity filter changes

11EJDE11:
- Adds a new profanity filter.
- Adds a setting to CnCNet options panel to enable/disable word censoring (on by default - may need changing?).
- Censors words in channel names (the original, uncensored name remains accessible in code).
- Censors words in chat messages (the original, uncensored message remains accessible in code).
- Updates CnCNetOptionPanel to use constants for positioning.

The new profanity filter will filter l33tspeak variations of words, repeating characters, and basic bypass attempts. For example, these are all considered profane:

- Shit
- Sh!t
- Ss5shhhhii11tttt
- S_H_I-T
- 5_h_1_t

It's a regex based filter system. Hoping the comments make it easy to follow. On my somewhat dated laptop it can censor 10,000 medium sized strings in 200ms, and checking if a string is offensive will be quicker than that.

Censoring usernames isn't included in this PR. That's a different beast and will need more time.
The word "Admin" is now only considered profane when choosing your username.

### DIFF
--- a/ClientCore/ProfanityFilter.cs
+++ b/ClientCore/ProfanityFilter.cs
@@ -11,14 +11,20 @@ namespace ClientCore
     {
         private const char CENSOR_CHAR = '*';
         private readonly Regex _combinedRegex;
-        
-        public ProfanityFilter()
-        {
+
+        /// <summary>
+        /// Creates a new profanity filter with a default set of censored words.
+        /// </summary>
+        /// <param name="extraWords">Any extra words to be considered profane.</param>
+        /// 
+        public ProfanityFilter(IEnumerable<string> extraWords = null)
+        {        
             var defaultWords = new[]
             {
-                "fuck", "shit", "cunt", "nigger", "cock",
-                "hitler", "pussy", "akbar", "allahu", "paki"
+                "fuck", "shit", "cunt", "nigger", "nigga", "niggr", "cock",
+                "hitler", "pussy", "akbar", "allahu", "paki", "twat"
             };
+            var allWords = defaultWords.Concat(extraWords ?? Enumerable.Empty<string>());
 
             //for each bad word
             //      for each letter in the bad word,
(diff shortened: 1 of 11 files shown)


ClientCore/ProfanityFilter.cs
```diff
@@ -1,90 +1,97 @@
...
-                "admin*",
-                "allahu*",
-                "akbar",
-                "twat",
-                "cock",
-                "pussy",
-                "hitler*",
-                "anal"
+                "fuck", "shit", "cunt", "nigger", "nigga", "niggr", "cock",
+                "hitler", "pussy", "akbar", "allahu", "paki", "twat"
```
Metadorius, line 25: With client being localized now, I think it's high time to add localization support here somehow

ClientCore/ProfanityFilter.cs
```diff
@@ -1,90 +1,97 @@
...
+                        {
+                            case 'a': return "[aA4@]+"; //in between [] are leetspeak variations.
+                            case 'b': return "[bB8]+";  //and a + matches repeating letters
+                            case 'e': return "[eE3]+";
+                            case 'g': return "[gG69]+";
+                            case 'i': return "[iI1!]+";
+                            case 'o': return "[oO0]+";
+                            case 's': return "[sS5$]+";
+                            case 't': return "[tT7]+";
+                            case 'z': return "[zZ2]+";
```
Metadorius, line 51: *A person with a Russian keyboard walks into a bar and writes "пigger"*

In other words, I think this isn't a robust method and only works for English language. What I would do is to have lookalike characters placed in a set, for example. You'd have a list of sets and just query each one of them, and build a regex out of this. Maybe it's also a good idea to add localization support here too somehow, because otherwise it would be an uphill battle not worth taking.

Metadorius (commented):
As it is I think it's an uphill battle, and a quite tough one. I can totally see the developments of this having false positives while also not covering all of the workarounds.
1. Are there any ready made profanity detection libraries that we could use to not reinvent the wheel?
2. I would look into something to handle all unicode lookalike characters. Doing this manually would be a daunting task, maybe there's some ready-made solution
3. Maybe offloading this onto client localizators will be a good idea?

PR #666, opened 2025-02-14
Add setting to disable in-game chat

11EJDE11:
Add setting to disable in-game chat
Add layout constants to GameOptionsPanel
Fix incorrectly named trbScrollRate

https://github.com/CnCNet/cncnet-yr-client-package/issues/456

### DIFF
--- a/DTAConfig/OptionPanels/GameOptionsPanel.cs
+++ b/DTAConfig/OptionPanels/GameOptionsPanel.cs
@@ -12,6 +12,16 @@ namespace DTAConfig.OptionPanels
 {
     class GameOptionsPanel : XNAOptionsPanel
     {
+        private const int PADDING_X = 12;
+        private const int PADDING_Y = 14;
+        private const int TRACKBAR_X_PADDING = 32;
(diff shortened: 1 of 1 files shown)


Metadorius:
Is `DisableChat` a YR only thing? If so - it shouldn't be hardcoded. We actually have a syntax for custom checkboxes and dropdowns present, please see this for an example: https://github.com/Starkku/cncnet-client-mod-base/blob/master/ClientFiles/Resources/OptionsWindow.ini

GrantBartlett:
> Add setting to disable in-game chat Add layout constants to GameOptionsPanel Fix incorrectly named trbScrollRate
> 
> [CnCNet/cncnet-yr-client-package#456](https://github.com/CnCNet/cncnet-yr-client-package/issues/456)

```
[chkChatEnabled]
Location=242,0
Text=Chat Enabled
ToolTip=When enabled, you can send and receive messages in-game. When disabled, messaging is unavailable.
DefaultValue=False
SettingSection=Options
SettingKey=DisableChat
```

Untested, copied from the beta branch in the yr client package, but something like this should work I think

Issue #641, opened 2025-01-18
After a failed game start, players are shown as playing


CCCP84:
Another problem. When I started the game, one player at that moment left. For me it looks like all the players are not loading. For the players it looks like the host is not loading. After everyone returns to the room, opposite the players the icon is lit as if the players are still in the game, although they have already came back. To fix this they all have to leave and come back into the room.
I had this problem already few times.

![Image](https://github.com/user-attachments/assets/0163ac1e-f75d-45db-9e49-bfac712b8056)

11EJDE11:
#680 has fixed this for me so I will close.

PR #667, opened 2025-02-19
Fix crash when starting game with null map

11EJDE11:
This issue occurs when all players are marked as Ready, and the host quickly switches to a map that some players don't have before clicking Start. Players missing the map will crash.

The crash happens in `GameLobbyBase.cs` within `WriteSpawnIni` at the `if (Map.IsCoop)` check. This fix prevents the game reaching that line with a null map.

GameProcessExited is called to inform the host that the player has left the game. Without this, the host would still consider the player as IsInGame which would cause issues (I think) similar to #641.

The match statistics changes are because GameProcessExited calls them which won't work with a null map. 
If it's preferred, instead of calling GameProcessExited and making changes to match statistics, we could just call the relevant lines of GameProcessExited directly.

```
            ClearReadyStatuses();
            CopyPlayerDataToUI();
            UpdateDiscordPresence(true);
```

Not sure which is best or preferred.

### DIFF
--- a/DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
+++ b/DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
@@ -1231,6 +1231,16 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
         /// </summary>
         protected override void GameProcessExited()
         {
+            ResetGameState();
+        }
+
+        protected void GameStartAborted()
+        {
+            ResetGameState();
+        }
+
+        protected void ResetGameState() 
+        {
             base.GameProcessExited();
 
             channel.SendCTCPMessage("RETURN", QueuedMessageType.SYSTEM_MESSAGE, 20);
@@ -1260,7 +1270,7 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
 
             if (Map == null)
             {
-                GameProcessExited();
+                GameStartAborted();
                 return;
             }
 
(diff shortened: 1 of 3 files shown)


DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
```diff
@@ -1258,6 +1258,12 @@ private void NonHostLaunchGame(string sender, string message)
             if (sender != hostName)
                 return;
 
+            if (Map == null)
+            {
+                GameProcessExited();
```
SadPencil, line 1263: Whenever this line is executed, does it always mean the game process is exited? If not, even if the implementation is the same, we should not call such a method. I suggest you made another method, something like `HandleNullMap()` or something

SadPencil (changes requested):

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
11EJDE11: No, the game process isn't always exited - in this case the game process has never even started. But the game host will think they are in the game and the only way the host knows the player has returned from the game is if they receive a RETURN message which happens through GameProcessExited. 
I'm happy to split it out to another method (AbortGameStart?), but any thoughts instead on renaming GameProcessExited to ResetGameState and just using that?

SadPencil: > No, the game process isn't always exited - in this case the game process has never even started. But the game host will think they are in the game and the only way the host knows the player has returned from the game is if they receive a RETURN message which happens through GameProcessExited. I'm happy to split it out to another method (AbortGameStart?), but any thoughts instead on renaming GameProcessExited to ResetGameState and just using that?

Maybe keeping all 3 methods:
```cs
GameStartAborted() => ResetGameState();
GameProcessExited() => ResetGameState();
```

SadPencil (approved):

Metadorius (approved):

Issue #664, opened 2025-02-12
Players' ready status when AutoReady is enabled is not always accurate on host's side.


11EJDE11:
This has been mentioned on Discord a few times as of late. 

In my experience:
It can occur without having started a game.
It's just the host that shows the statuses incorrectly. Non hosts it is correct.
IIRC, this issue used to crop up even prior to the large client update, just less frequently.

I've done some analysis and can replicate the problem if the host quickly goes from map > null map > map. I think this is what's happening:

>When a map change happens, the host queues a Game Options (GO) message to be sent to non-hosts with the new map. If the map is null then the host also resets the ready statuses.

>Because the host changed the map quickly, the queue eliminates the first GO message (containing the null map) and only sends the second (containing the valid map).

>This results in [LastMapChangeWasInvalid](https://github.com/CnCNet/xna-cncnet-client/blob/e5034eb92fc2a493c9cd1e061839c26575b33d22/DXMainClient/DXGUI/Multiplayer/GameLobby/MultiplayerGameLobby.cs#L1123) not being set to True for non-hosts as the only message they received indicated a valid map. The result of which is they won't send their ready status to the host (who as mentioned, has already reset the ready statuses).

I'm not sure how best to fix. RequestReadyStatus with every map change?

There was a commit that [modified LastMapChangeWasInvalid last month](https://github.com/CnCNet/xna-cncnet-client/commit/a0f2d15044faad99f2d3e1d0d6b36ea0be0fc8d1).
I don't really think that that is to blame. I just think this issue is cropping up more because the client is sending null maps more often (searching for a map with gibberish used to hold the last map, but with recent updates it changes it to null).

SadPencil:
> If the map is null then the host also resets the ready statuses.

Not sure, but whether it would help if the host does not reset the ready status? For a null map, the client only prevents the host launching the game, without clearing these statuses.

PR #670, opened 2025-02-19
Fix search selecting null maps, that also caused autoready issues

GrantBartlett:
Should fix auto ready issues as described here https://github.com/CnCNet/xna-cncnet-client/issues/664

## Before

https://github.com/user-attachments/assets/f6775b85-fa4e-45cb-a2cb-e16e0ff126ef

## After (with the change applied)

https://github.com/user-attachments/assets/0977ece5-6e14-4dd0-ae4f-9df159373e1c

## Auto ready fixed

https://github.com/user-attachments/assets/8218e980-907d-4392-869b-00f9fb8efe24


SadPencil:
This line was introduced to allow users switching the game mode and automatically select the same map in that mode. Simply removing this line will break such a thing. We need to make this line conditionally executes

SadPencil:
Code updated. Can you test it again?

SadPencil (approved):

PR #684, opened 2025-04-24
Fix crash when the player themself is not in the player list

### LINKED ISSUE #683: Random crash after leftclick autoready - on v8.70.0
[ClientCrashLog_2025_04_19_19_42.txt](https://github.com/user-attachments/files/19822293/ClientCrashLog_2025_04_19_19_42.txt)

Image

SadPencil:
Fixes #683 

```
19.04. 19:42:40.783    KABOOOOOOM!!! Info:
19.04. 19:42:40.784    Type: System.NullReferenceException
19.04. 19:42:40.784    Message: Object reference not set to an instance of an object.
19.04. 19:42:41.122    Source: clientdx
19.04. 19:42:41.122    TargetSite.Name: RequestReadyStatus
19.04. 19:42:41.125    Stacktrace:    at DTAClient.DXGUI.Multiplayer.GameLobby.CnCNetGameLobby.RequestReadyStatus() in C:\Dev\xna-cncnet-client\DXMainClient\DXGUI\Multiplayer\GameLobby\CnCNetGameLobby.cs:line 736
   at DTAClient.DXGUI.Multiplayer.GameLobby.MultiplayerGameLobby.ChkAutoReady_CheckedChanged(Object sender, EventArgs e) in C:\Dev\xna-cncnet-client\DXMainClient\DXGUI\Multiplayer\GameLobby\MultiplayerGameLobby.cs:line 392
```

### DIFF
--- a/DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
+++ b/DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
@@ -707,6 +707,9 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
             }
 
             PlayerInfo pInfo = Players.Find(p => p.Name == ProgramConstants.PLAYERNAME);
+            if (pInfo == null)
+                return;
+
             int readyState = 0;
 
             if (chkAutoReady.Checked)


MahBoiDeveloper (approved):

SadPencil:
> Are you sure it's not treating a symptom instead of a root cause?

Even if we might need to investigate why such thing could happen (maybe something related to the remote server?), such a fix should not be delayed because:

there are *already* ~10 occurrence of searching a player that matches the PLAYERNAME and followed by an if-null-return. The codes here is the only one without an if-null-return.

Metadorius (approved):

PR #691, opened 2025-05-03
Fix potential crash when client map name is not a valid Win32 file name

MahBoiDeveloper:
Fix crash when client tries to save map file with map name containing special symbols not allowed for files on NTFS.

Crash log: [ClientCrashLog_2025_05_02_23_55.txt](https://github.com/user-attachments/files/20024447/ClientCrashLog_2025_05_02_23_55.txt)

Map file to recreate crash: [14ac45ba3197c2f19a36c4269a0f263aee2ce157.zip](https://github.com/user-attachments/files/20024448/14ac45ba3197c2f19a36c4269a0f263aee2ce157.zip)

### DIFF
--- a/ClientCore/Extensions/StringExtensions.cs
+++ b/ClientCore/Extensions/StringExtensions.cs
@@ -106,7 +106,7 @@ public static class StringExtensions
         }, StringComparer.InvariantCultureIgnoreCase);
 
         if (reservedFileNames.Contains(filename))
-            filename += "_map";
+            filename += "_";
 
         return filename;
     }
--- a/ClientCore/Extensions/StringExtensions.cs
+++ b/ClientCore/Extensions/StringExtensions.cs
@@ -77,4 +77,12 @@ public static class StringExtensions
         => string.IsNullOrEmpty(defaultValue)
             ? defaultValue
             : Translation.Instance.LookUp(key, defaultValue, notify);
+
+    public static string DeleteSpecialSymbols(this string defaultValue)
+    {
+        foreach (char ch in "/\\:*?<>|")
+            defaultValue = defaultValue.Replace(ch, ' ');
+
+        return defaultValue;
+    }
 }
--- a/ClientCore/Extensions/StringExtensions.cs
+++ b/ClientCore/Extensions/StringExtensions.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using ClientCore.I18N;
 
 namespace ClientCore.Extensions;
@@ -79,6 +79,7 @@ public static class StringExtensions
             : Translation.Instance.LookUp(key, defaultValue, notify);
 
     public static string DeleteSpecialSymbols(this string defaultValue)
+    public static string ToWin32FileName(this string defaultValue)
     {
         foreach (char ch in "/\\:*?<>|")
             defaultValue = defaultValue.Replace(ch, ' ');
--- a/ClientCore/Extensions/StringExtensions.cs
+++ b/ClientCore/Extensions/StringExtensions.cs
@@ -1,4 +1,6 @@
 ﻿using System;
+using System.Collections.Generic;
+
 using ClientCore.I18N;
 
 namespace ClientCore.Extensions;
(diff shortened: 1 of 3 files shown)


SadPencil:
Technically, these symbols are not allowed as valid Win32 file names. NTFS does not have such an restriction.

SadPencil (changes requested):
Also, I am not sure whether we should, for example, rename something like "con" (a disallowed file name if the file extension is none) into "con1". Since map files have extensions maybe we can just leave a TODO comment in the source codes

ClientCore/Extensions/StringExtensions.cs
```diff
@@ -77,4 +77,12 @@ public static string L10N(this string defaultValue, string key, bool notify = tr
         => string.IsNullOrEmpty(defaultValue)
             ? defaultValue
             : Translation.Instance.LookUp(key, defaultValue, notify);
+
+    public static string DeleteSpecialSymbols(this string defaultValue)
```
SadPencil, line 81: This function name is too general.

Try something like `AsWin32FileName()`

SadPencil (changes requested):

ClientCore/Extensions/StringExtensions.cs
```diff
@@ -77,4 +77,42 @@ public static string L10N(this string defaultValue, string key, bool notify = tr
...
+        {
+            "CON",
+            "PRN",
+            "AUX",
+            "NUL",
+            "COM1", "COM2", "COM3", "COM4", "COM5", "COM6", "COM7", "COM8", "COM9", "COM¹", "COM²", "COM³",
+            "LPT1", "LPT2", "LPT3", "LPT4", "LPT5", "LPT6", "LPT7", "LPT8", "LPT9", "LPT¹", "LPT²", "LPT³"
+        };
+
+        foreach(var reserved in reservedFileNames)
```
SadPencil, line 107: there should be a space between foreach and left bracket

also I think it is better to use a HashSet (public static readonly IReadOnlySet) or an ImmutableList (public static readonly ImmutableList) instead of a list here

SadPencil (changes requested):

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
```diff
@@ -1653,6 +1653,7 @@ private void MapSharer_MapDownloadComplete(object sender, SHA1EventArgs e) =>
         private void MapSharer_HandleMapDownloadComplete(SHA1EventArgs e)
         {
             string mapFileName = MapSharer.GetMapFileName(e.SHA1, e.MapName);
+            mapFileName = mapFileName.ToWin32FileName();
```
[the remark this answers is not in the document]
SadPencil: Dismiss this comment

ClientCore/Extensions/StringExtensions.cs
```diff
@@ -77,4 +77,42 @@ public static string L10N(this string defaultValue, string key, bool notify = tr
...
+        {
+            if (defaultValue == reserved)
```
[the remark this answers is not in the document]
SadPencil: Nice catch!

ClientCore/Extensions/StringExtensions.cs
MahBoiDeveloper: > also I think it is better to use a HashSet (public static readonly IReadOnlySet) or an ImmutableList (public static readonly ImmutableList) instead of a list here

Ok, where should i place this hashset?

SadPencil: > > also I think it is better to use a HashSet (public static readonly IReadOnlySet) or an ImmutableList (public static readonly ImmutableList) instead of a list here
> 
> Ok, where should i place this hashset?

Sorry, I just realized that IReadOnlySet is not support in .NET 4.8. Therefore a static public class member is not a good idea anymore.

Speaking out of the current project, for a .net 8 project, we can write something like this:

```cs

public static class StringExtensions
{
    public static readonly IReadOnlySet<string> InvalidWin32FileNames = new HashSet<string>();
}
```

SadPencil (approved):

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
```diff
@@ -1653,6 +1653,7 @@ private void MapSharer_MapDownloadComplete(object sender, SHA1EventArgs e) =>
         private void MapSharer_HandleMapDownloadComplete(SHA1EventArgs e)
         {
             string mapFileName = MapSharer.GetMapFileName(e.SHA1, e.MapName);
```
Metadorius, line 1655: Are you sure you shouldn't modify `GetMapFileName` instead?

ClientCore/Extensions/StringExtensions.cs
```diff
@@ -77,4 +79,35 @@ public static string L10N(this string defaultValue, string key, bool notify = tr
...
+            "CON",
+            "PRN",
+            "AUX",
+            "NUL",
+            "COM1", "COM2", "COM3", "COM4", "COM5", "COM6", "COM7", "COM8", "COM9", "COM¹", "COM²", "COM³",
+            "LPT1", "LPT2", "LPT3", "LPT4", "LPT5", "LPT6", "LPT7", "LPT8", "LPT9", "LPT¹", "LPT²", "LPT³"
+        }, StringComparer.InvariantCultureIgnoreCase);
+
+        if (reservedFileNames.Contains(filename))
+            filename += "_map";
```
SadPencil: @MahBoiDeveloper I just realized that as a general utility function defined in `StringExtensions` class, appending `_map` as the suffix is not appropriate as it would be an application-related behavior. I changed it to `filename += "_";`

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
MahBoiDeveloper: Thank you for advice. Check 01975a6ed9c5310a8b2cfd9302267da2405c5c4d.

ClientCore/Extensions/StringExtensions.cs
MahBoiDeveloper: > @MahBoiDeveloper I just realized that as a general utility function defined in `StringExtensions` class, appending `_map` as the suffix is not appropriate as it would be an application-related behavior. I changed it to `filename += "_";`

Ok, thanks for help!

SadPencil (approved):

Metadorius (approved):

PR #699, opened 2025-05-05
Update nuget dependencies 

MahBoiDeveloper:


SadPencil:
Can you also try upgrading PolyFill to the latest?

MahBoiDeveloper:
> Can you also try upgrading PolyFill to the latest?

I've got an error `The operation is invalid due to the current state of the object.` while trying update it to any new version. I cannot even delete package because it also gives an error `The calculated object originating from the imported file “C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\IDE\CommonExtensions\Microsoft\NuGet\NuGet.targets” cannot be modified.`

MahBoiDeveloper:
Ok, version 7.29.0 throws exception about that Polyfill requires C# 13, not 12. Version 6.9.0 launches without any error.

SadPencil:
> Ok, version 7.29.0 throws exception about that Polyfill requires C# 13, not 12. Version 6.9.0 launches without any error.

Perhaps we need to modify `global.json` file in order to bring C# 13 support. GitHub CI set up .NET SDK according to this file. Otherwise, 6.9.0 also seems fine.

MahBoiDeveloper:
image

image

Hmm, [C# 13 requires .NET 9](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-versioning). Ok, i am downgrade Polyfill.

SadPencil:
> image
> 
> image
> 
> Hmm, [C# 13 requires .NET 9](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-versioning). Ok, i am downgrade Polyfill.

I remember the language requirement was not completely correct. But it does require a newer .NET SDK (not the .NET target version)

Example: this program uses C# 12 in .NET 4.0. https://github.com/CnCNet/dta-mg-client-launcher/blob/master/CnCNet.LauncherStub.csproj

MahBoiDeveloper:
> I remember the language requirement was not completely correct. But it does require a newer .NET SDK (not the .NET target version)
> Example: this program uses C# 12 in .NET 4.0. https://github.com/CnCNet/dta-mg-client-launcher/blob/master/CnCNet.LauncherStub.csproj

However, people will need to install .NET 9.0 to enable C# 13, as well as .NET 8.0 and C# 12. As VS2022 installs .NET 9.0 by default this is not a problem, but changing client .NET requirements to 9.0 will need to be discussed.

PR #704, opened 2025-05-09
Fix `spawnmap.ini` encoding getting messed up

MahBoiDeveloper:
The bug was first discovered in 2019 when I was translating Twisted Insurrection into Russian. The bug was caused by the fact that `spawnmap.ini` is always created in UTF8 encoding, even if the original file was in a different one, for example in CP1251.

Important map code in UTF8:

image

Important map code in CP1251:

image

Before patch:

image

After patch:

image

### DIFF
--- a/ClientCore/FileHelper.cs
+++ b/ClientCore/FileHelper.cs
@@ -90,5 +90,21 @@ namespace ClientCore
                 throw new PlatformNotSupportedException();
             }
         }
+
+        public static Encoding GetEncoding(string filename)
+        {
+            Encoding encoding = Encoding.UTF8;
+
+            using (FileStream fs = File.OpenRead(filename))
+            {
+                Ude.CharsetDetector cdet = new Ude.CharsetDetector();
+                cdet.Feed(fs);
+                cdet.DataEnd();
+                if (cdet.Charset != null)
+                    encoding = Encoding.GetEncoding(cdet.Charset);
+            }
+
+            return encoding;
+        }
     }
 }
--- a/DXMainClient/Domain/Multiplayer/Map.cs
+++ b/DXMainClient/Domain/Multiplayer/Map.cs
@@ -15,6 +15,7 @@ using Point = Microsoft.Xna.Framework.Point;
 using Utilities = Rampastring.Tools.Utilities;
 using static System.Collections.Specialized.BitVector32;
 using System.Diagnostics;
+using System.Text;
 
 namespace DTAClient.Domain.Multiplayer
 {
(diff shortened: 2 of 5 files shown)


Metadorius (approved):

ClientCore/FileHelper.cs
```diff
@@ -90,5 +90,21 @@ public static void CreateHardLinkFromSource(string source, string destination, b
                 throw new PlatformNotSupportedException();
             }
         }
+
+        public static Encoding GetEncoding(string filename)
+        {
+            Encoding encoding = Encoding.UTF8;
```
SadPencil, line 96: The default encoding in IniFile is `new UTF8Encoding(false)` (without BOM) instead of Encoding.UTF8 (with BOM).

DXMainClient/Program.cs
```diff
@@ -100,6 +102,9 @@ static void InitializeApplicationConfiguration()
 #endif
         static void Main(string[] args)
         {
+            // https://stackoverflow.com/questions/3967716/how-to-find-encoding-for-1251-codepage
+            Encoding.RegisterProvider(CodePagesEncodingProvider.Instance);
```
SadPencil, line 106: This line was executed in `ClientCore.PlatformShim.EncodingExt`. Since that static initialization method does not guarantee to be executed, I guess it is okay to execute it in the main method.

DXMainClient/Domain/Multiplayer/Map.cs
```diff
@@ -711,11 +712,15 @@ public Texture2D LoadPreviewTexture()
...
-            var mapIni = new IniFile(CompleteFilePath);
+            Encoding encoding = FileHelper.GetEncoding(CompleteFilePath);
+
+            var mapIni = new IniFile(CompleteFilePath, encoding);
 
             if (!string.IsNullOrEmpty(ExtraININame))
             {
-                var extraIni = new IniFile(SafePath.CombineFilePath(ProgramConstants.GamePath, "INI", "Map Code", ExtraININame));
+                string extraIniPath = SafePath.CombineFilePath(ProgramConstants.GamePath, "INI", "Map Code", ExtraININame);
+                encoding = FileHelper.GetEncoding(extraIniPath);
```
SadPencil, line 722: Is it TS the only platform that requires a non Unicode encoding on this file? I don't remember the spawner for YR or Phobos has this requirements. Therefore, I suggest wrapping it with an `if-TS`.

Detecting file encoding from a text file is basically "guessing", so we should avoid such a behavior if we are certain the program that processes the text file in UTF8, with or without BOM.

ClientCore/FileHelper.cs
MahBoiDeveloper: Done by 2336d2f74966e9a646000cb504ac2cf7d66af953.

DXMainClient/Program.cs
MahBoiDeveloper: Done by 2336d2f74966e9a646000cb504ac2cf7d66af953.

DXMainClient/Domain/Multiplayer/Map.cs
MahBoiDeveloper: Done by 0435e605f66a171bacc63843ea251f02969af45e.

DXMainClient/Program.cs
SadPencil: I am not suggesting changes in this comment

DXMainClient/Domain/Multiplayer/Map.cs
SadPencil: The file helper should provide the general, game-irrelevant methods.

If-TS should be applied directly here, instead of in the ClientCore

SadPencil:
Done in f6be8ae8ab75d684ae7552b62cd51185a75b6fdf

DXMainClient/Domain/Multiplayer/Map.cs
Metadorius: some YR mods still rely on non-Unicode for translating their stuff, and it is possible in principle (albeit is more problematic than unicode, but Unicode input is only propely supported in Phobos, so there's that), so I don't think that is valid necessarily

PR #715, opened 2025-05-13
Use mapdb staging on beta branch

alexlambson:


MahBoiDeveloper (approved):

SadPencil (commented):
Does this title indicates it's a testing or temporary service? They can be defined in the network definition file if so.

SadPencil:
I will close this PR for now. Feel free to reopen it if it will be the default setting for everyone.

alexlambson:
We're probably going to do something with the subdomains instead so we can test all clients

PR #721, opened 2025-05-16
Refactor control names in client's code base

MahBoiDeveloper:
Important fix: changed name for `CheaterWindow` from `CheaterScreen`.

### DIFF
--- a/ClientGUI/DarkeningPanel.cs
+++ b/ClientGUI/DarkeningPanel.cs
@@ -21,7 +21,7 @@ namespace ClientGUI
 
         public override void Initialize()
         {
-            Name = "DarkeningPanel";
+            Name = nameof(DarkeningPanel);
 
             SetPositionAndSize();
             
--- a/ClientGUI/XNAMessageBox.cs
+++ b/ClientGUI/XNAMessageBox.cs
@@ -104,7 +104,7 @@ namespace ClientGUI
             btnOK.IdleTexture = AssetLoader.LoadTexture("75pxbtn.png");
             btnOK.HoverTexture = AssetLoader.LoadTexture("75pxbtn_c.png");
             btnOK.HoverSoundEffect = new EnhancedSoundEffect("button.wav");
-            btnOK.Name = "btnOK";
+            btnOK.Name = nameof(btnOK);
             btnOK.Text = "OK".L10N("Client:ClientGUI:ButtonOK");
             btnOK.LeftClick += BtnOK_LeftClick;
             btnOK.HotKey = Keys.Enter;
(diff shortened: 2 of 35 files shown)


SadPencil:
Perhaps it's better to retain the old `CheaterScreen` name for compatibility

Metadorius:
> Perhaps it's better to retain the old `CheaterScreen` name for compatibility

I agree

MahBoiDeveloper:
Ok, then let's rename `CheaterWindow` class to `CheaterScreen`.

SadPencil:
> Ok, then let's rename `CheaterWindow` class to `CheaterScreen`.

I don't think its proper to rename the class neither as all classes here ends with `Window`

Metadorius:
Just keep the names as is.

Actually, I'd like to raise a concern. We probably should not use nameof in this case *at all*, because someone can rename the variable in code and impact the client functionality this way. This is not how it should be done, is it?

MahBoiDeveloper:
> Just keep the names as is.
> 
> Actually, I'd like to raise a concern. We probably should not use nameof in this case _at all_, because someone can rename the variable in code and impact the client functionality this way. This is not how it should be done, is it?

Changes in names that already use `nameof` should be mentioned in release message text instead of deprecating renaming at all.

Metadorius:
> Changes in names that already use `nameof` should be mentioned in release message text instead of deprecating renaming at all.

You do realise you're just adding friction for almost no real benefit, right? Yes, ideally, they'd all be consistently named, but they have not been consistently named for quite some time, and changing that out of perfectionism is an ass move to our consumers.

SadPencil:
> > Changes in names that already use `nameof` should be mentioned in release message text instead of deprecating renaming at all.
> 
> You do realise you're just adding friction for almost no real benefit, right? Yes, ideally, they'd all be consistently named, but they have not been consistently named for quite some time, and changing that out of perfectionism is an ass move to our consumers.

Either `nameof(AName)` or `"AName"` will show in git diff if the rename to a class was processed, so we can examine whether this rename is intended or not.

I do think `nameof(AName)` is better **if** the name is identical to the class name. It tells people the name is the class name without need to manually check whether the string equals. For the `CheaterWindow`, use `"CheaterScreen"` (not a `nameof`) to demonstrate the name is different than the class name.

SadPencil:
@Metadorius I suggest accepting this PR based on the readability advantage of nameof. The CheatWindow renaming thing is reverted by @MahBoiDeveloper and this PR brings only coding style modifications now.

Metadorius:
> Either `nameof(AName)` or `"AName"` will show in git diff if the rename to a class was processed, so we can examine whether this rename is intended or not.

What if someone misses that?

With previous approach there's just less slippery ground, you don't even need to think about compatibility. Now you need one more thing to keep in mind and to check for.

What does this PR improve? I don't see it improving anything, only introducing potential traps for future contributors. I don't see any readability advantages in this change personally, it's clear to see that the name is exactly the same by just, well, reading it.

PR #728, opened 2025-05-21
Avoid the abuse of random generator initialization

SadPencil:
This PR uses dependency injection to provide a shared Random class whose seed is NOT generated from the system time but a reliable RNG from [System.Security.Cryptography.RandomNumberGenerator](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.randomnumbergenerator). 

This PR solves the issues below:
- Calling `new Random()` with in a short time will get an RNG with the same seed. This makes `RandomizeChannelName()` waste some times.
- Different players might have a chance to get and register the same channel if the time is close enough in `RandomizeChannelName()`


SadPencil:
On my machine it works well. I'll merge it

PR #732, opened 2025-05-23
Fix potential bug with different lengths of message array

MahBoiDeveloper:


Metadorius (approved):

SadPencil:
Different lengths? Aren't they all the same 1024?

PR #736, opened 2025-05-27
Refactor handling errors in `Connection` class

MahBoiDeveloper:


DXMainClient/Online/Connection.cs
```diff
@@ -240,28 +243,57 @@ private void HandleComm()
...
                 {
-                    Logger.Log("Disconnected from CnCNet due to a socket error. Message: " + ex.ToString());
                     errorTimes++;
 
-                    if (errorTimes > MAX_RECONNECT_COUNT)
+                    if (errorTimes > MAX_ERROR_COUNT)
                     {
                         const string errorMessage = "Disconnected from CnCNet after reaching the maximum number of connection retries.";
-                        Logger.Log(errorMessage);
+                        Logger.Log(errorMessage + "Message: " + ex.ToString());
```
Metadorius, line 259: forgot newlines

DXMainClient/Online/Connection.cs
```diff
@@ -240,28 +243,57 @@ private void HandleComm()
...
                     continue;
                 }
+                catch (Exception ex)
+                {
+                    errorTimes++;
+
+                    if (errorTimes > MAX_ERROR_COUNT)
+                    {
+                        const string errorMessage = "Disconnected from CnCNet due to an internal error.";
+                        Logger.Log(errorMessage + "Message: " + ex.ToString());
```
Metadorius, line 274: same

Metadorius (changes requested):

DXMainClient/Online/Connection.cs
MahBoiDeveloper: Done by 99c91b85dd32d2d17d013f1049602e89001129f3.

DXMainClient/Online/Connection.cs
MahBoiDeveloper: Done by 99c91b85dd32d2d17d013f1049602e89001129f3.

Metadorius:
> it might be good actively handle `ObjectDisposedException` (The NetworkStream is closed)?

I don't understand what benefit will that bring, from a brief read it only happens when the object is disposed, which should not happen and is a coding error.

SadPencil:
> > it might be good actively handle `ObjectDisposedException` (The NetworkStream is closed)?
> 
> I don't understand what benefit will that bring, from a brief read it only happens when the object is disposed, which should not happen and is a coding error.

It seems 6bf2db20a683f24f01753efc93a29454686ab5e1 solves my concern so I don't require changes anymore. Even if this coding error should not happen, it seems better to immediately return an exception instead of let the client retries for 8 times.

Issue #687, opened 2025-04-30
Multiple message boxes shown simultaneously results in a loop of button sound and unclickable buttons


SadPencil:
`MainMenu.cs` file:

```cs
public void PostInit()
{

    CheckRequiredFiles();
    CheckForbiddenFiles();
    CheckIfFirstRun();
}
```

These 3 methods can each create an XNAMessageBox. For example, let's say a required file is missing while another forbidden file is here. There will be two message boxes shown simultaneously.

Moving the cursor into some areas result in an infinite loop playing the button sound, clicking the ok button makes no effect.

![Image](https://github.com/user-attachments/assets/a2256c28-29d8-4041-acbd-ee2f31ac42e5)

What's worse, if the 2nd message box completely overrides the first, and the two ok buttons happen to overlap each other, the user will be so confusing since he/she would have no way to close the message box.

![Image](https://github.com/user-attachments/assets/4e4b4150-d43d-45c0-aa09-fd6f8ddc92cd)

PR #763, opened 2025-06-25
Handle multiple links in chat message

### LINKED ISSUE #762: Improve chat links handling
### Description

The current implementation can only handle one link per message (the first one). However, users and bots in the chat can send messages containing several links.

MahBoiDeveloper:
New implementation can handle one or more links in single message.

Example: 
image

Closes #762

### DIFF
--- a/DXMainClient/DXGUI/Multiplayer/ChatListBox.cs
+++ b/DXMainClient/DXGUI/Multiplayer/ChatListBox.cs
@@ -1,16 +1,12 @@
 ﻿using System;
-using System.Linq;
 
-using ClientCore;
 using ClientCore.Extensions;
 
-using ClientGUI;
-
+using DTAClient.DXGUI.Generic;
 using DTAClient.Online;
 
 using Microsoft.Xna.Framework;
 
-using Rampastring.Tools;
 using Rampastring.XNAUI;
 using Rampastring.XNAUI.XNAControls;
 
@@ -34,57 +30,15 @@ namespace DTAClient.DXGUI.Multiplayer
 
             // Get the clicked links
             string[] links = Items[SelectedIndex].Text?.GetLinks();
-            
-            if (links == null) 
+
+            if (links == null)
                 return;
 
             if (links.Length == 0 || links.Length > 1)
                 return;
 
             string link = links[0];
-
-            // Determine if the links is trusted
-            bool isTrusted = false;
-            try
-            {
-                string domain = new Uri(link).Host;
-                var trustedDomains = ClientConfiguration.Instance.TrustedDomains.Concat(ClientConfiguration.Instance.AlwaysTrustedDomains);
-                isTrusted = trustedDomains.Contains(domain, StringComparer.InvariantCultureIgnoreCase)
-                    || trustedDomains.Any(trustedDomain => domain.EndsWith("." + trustedDomain, StringComparison.InvariantCultureIgnoreCase));
-            }
-            catch (Exception ex)
-            {
-                isTrusted = false;
-                Logger.Log($"Error in parsing the URL \"{link}\": {ex.ToString()}");
-            }
-
-            if (isTrusted)
-            {
-                ProcessLink(link);
-                return;
-            }
-
-            // Show the warning if the links is not trusted
-            var msgBox = new XNAMessageBox(WindowManager,
-                "Open Link Confirmation".L10N("Client:Main:OpenLinkConfirmationTitle"),
-                """
-                    You're about to open a link shared in chat.
-
-                    Please note that this link hasn't been verified,
-                    and CnCNet is not responsible for its content.
-
-                    Would you like to open the following link in your browser?
-                    """.L10N("Client:Main:OpenLinkConfirmationText")
-                + Environment.NewLine + Environment.NewLine + link,
-                XNAMessageBoxButtons.YesNo);
-            msgBox.YesClickedAction = (msgBox) => ProcessLink(link);
-            msgBox.Show();
-        }
-
-        private void ProcessLink(string link)
-        {
-            if (link != null)
-                ProcessLauncher.StartShellProcess(link);
+            URLHandler.OpenLink(WindowManager, link);
         }
 
         public void AddMessage(string message)
--- a/DXMainClient/DXGUI/Multiplayer/ChatListBox.cs
+++ b/DXMainClient/DXGUI/Multiplayer/ChatListBox.cs
@@ -27,47 +27,50 @@ namespace DTAClient.DXGUI.Multiplayer
             if (SelectedIndex < 0 || SelectedIndex >= Items.Count)
                 return;
 
-            // Get the clicked link
-            string link = Items[SelectedIndex].Text?.GetLink();
-            if (link == null)
+            // Get the clicked links
+            string[] links = Items[SelectedIndex].Text?.GetLinks();
+            if (links == null)
                 return;
 
-            // Determine if the link is trusted
-            bool isTrusted = false;
-            try
+            foreach (string link in links)
             {
-                string domain = new Uri(link).Host;
-                var trustedDomains = ClientConfiguration.Instance.TrustedDomains.Concat(ClientConfiguration.Instance.AlwaysTrustedDomains);
-                isTrusted = trustedDomains.Contains(domain, StringComparer.InvariantCultureIgnoreCase)
-                    || trustedDomains.Any(trustedDomain => domain.EndsWith("." + trustedDomain, StringComparison.InvariantCultureIgnoreCase));
+                // Determine if the links is trusted
+                bool isTrusted = false;
+                try
+                {
+                    string domain = new Uri(link).Host;
+                    var trustedDomains = ClientConfiguration.Instance.TrustedDomains.Concat(ClientConfiguration.Instance.AlwaysTrustedDomains);
+                    isTrusted = trustedDomains.Contains(domain, StringComparer.InvariantCultureIgnoreCase)
+                        || trustedDomains.Any(trustedDomain => domain.EndsWith("." + trustedDomain, StringComparison.InvariantCultureIgnoreCase));
+                }
+                catch (Exception ex)
+                {
+                    isTrusted = false;
+                    Logger.Log($"Error in parsing the URL \"{link}\": {ex.ToString()}");
+                }
+
+                if (isTrusted)
+                {
+                    ProcessLink(link);
+                    continue;
+                }
+
+                // Show the warning if the links is not trusted
+                var msgBox = new XNAMessageBox(WindowManager,
+                    "Open Link Confirmation".L10N("Client:Main:OpenLinkConfirmationTitle"),
+                    """
+                    You're about to open a link shared in chat.
+
+                    Please note that this link hasn't been verified,
+                    and CnCNet is not responsible for its content.
+
+                    Would you like to open the following link in your browser?
+                    """.L10N("Client:Main:OpenLinkConfirmationText")
+                    + Environment.NewLine + Environment.NewLine + link,
+                    XNAMessageBoxButtons.YesNo);
+                msgBox.YesClickedAction = (msgBox) => ProcessLink(link);
+                msgBox.Show();
             }
-            catch (Exception ex)
-            {
-                isTrusted = false;
-                Logger.Log($"Error in parsing the URL \"{link}\": {ex.ToString()}");
-            }
-
-            if (isTrusted)
-            {
-                ProcessLink(link);
-                return;
-            }
-
-            // Show the warning if the link is not trusted
-            var msgBox = new XNAMessageBox(WindowManager,
-                "Open Link Confirmation".L10N("Client:Main:OpenLinkConfirmationTitle"),
-                """
-                You're about to open a link shared in chat.
-
-                Please note that this link hasn't been verified,
-                and CnCNet is not responsible for its content.
-
-                Would you like to open the following link in your browser?
-                """.L10N("Client:Main:OpenLinkConfirmationText")
-                + Environment.NewLine + Environment.NewLine + link,
-                XNAMessageBoxButtons.YesNo);
-            msgBox.YesClickedAction = (msgBox) => ProcessLink(link);
-            msgBox.Show();
         }
 
         private void ProcessLink(string link)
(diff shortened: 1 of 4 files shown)


DXMainClient/DXGUI/Multiplayer/ChatListBox.cs
```diff
@@ -27,47 +27,50 @@ private void ChatListBox_DoubleLeftClick(object sender, EventArgs e)
...
-            if (link == null)
+            // Get the clicked links
+            string[] links = Items[SelectedIndex].Text?.GetLinks();
+            if (links == null)
                 return;
 
-            // Determine if the link is trusted
-            bool isTrusted = false;
-            try
+            foreach (string link in links)
```
Metadorius: I don't think double clicking on a message should open multiple links

MahBoiDeveloper: > I don't think double clicking on a message should open multiple links

Feature ported from FF's client

Metadorius: well, FF isn't known for necessarily great UI/UX, so we should review features anyways
don't make changes for now, I am not sure how to handle that, accepting ideas

MahBoiDeveloper:
@SadPencil could you do a code review, please?

DXMainClient/DXGUI/Multiplayer/ChatListBox.cs
```diff
@@ -27,47 +27,50 @@ private void ChatListBox_DoubleLeftClick(object sender, EventArgs e)
             if (SelectedIndex < 0 || SelectedIndex >= Items.Count)
                 return;
 
-            // Get the clicked link
-            string link = Items[SelectedIndex].Text?.GetLink();
-            if (link == null)
+            // Get the clicked links
+            string[] links = Items[SelectedIndex].Text?.GetLinks();
+            if (links == null)
```
SadPencil, line 34: What if links.Length be 0?

DXMainClient/DXGUI/Multiplayer/CnCNet/GlobalContextMenu.cs
```diff
@@ -181,7 +223,7 @@ private void CopyLink(string link)
             }
             catch (Exception)
             {
-                XNAMessageBox.Show(WindowManager, "Error".L10N("Client:Main:Error"), "Unable to copy link".L10N("Client:Main:ClipboardCopyLinkFailed"));
+                XNAMessageBox.Show(WindowManager, "Error".L10N("Client:Main:Error"), "Unable to copy links".L10N("Client:Main:ClipboardCopyLinkFailed"));
```
SadPencil: L10N key should be updated or appended with something like V2 to remind translators to update the text

DXMainClient/DXGUI/Multiplayer/CnCNet/GlobalContextMenu.cs
```diff
@@ -223,6 +265,19 @@ private IRCUser GetIrcUser()
             return null;
         }
 
+        public void RemoveLinks()
+        {
+            for (int i = 0; i < Items.Count; i++)
+            {
+                if (!(Items[i].Text.Contains(COPY_LINK) || Items[i].Text.Contains(OPEN_LINK)))
```
SadPencil: Umm, not sure about this line. Can you avoid using string operations to determine different behaviors?

DXMainClient/DXGUI/Multiplayer/ChatListBox.cs
```diff
@@ -27,47 +27,50 @@ private void ChatListBox_DoubleLeftClick(object sender, EventArgs e)
...
+                    continue;
+                }
 
-            // Show the warning if the link is not trusted
-            var msgBox = new XNAMessageBox(WindowManager,
-                "Open Link Confirmation".L10N("Client:Main:OpenLinkConfirmationTitle"),
-                """
-                You're about to open a link shared in chat.
+                // Show the warning if the links is not trusted
+                var msgBox = new XNAMessageBox(WindowManager,
```
SadPencil: So if a message contains multiple links, there will be multiple messageboxes shown one by one? Not sure if this is a good behavior or not

MahBoiDeveloper: Yes, it will be cascade of message boxes with untrusted links. Trusted links will be open without asking.

DXMainClient/DXGUI/Multiplayer/ChatListBox.cs
MahBoiDeveloper: Match always 1 and more or null

DXMainClient/DXGUI/Multiplayer/ChatListBox.cs
SadPencil: I don't think XNAUI is robust against multiple message boxes. https://github.com/CnCNet/xna-cncnet-client/issues/687

DXMainClient/DXGUI/Multiplayer/ChatListBox.cs
SadPencil: > Match always 1 and more or null

One can't get such an assumption from the parameters and return type of function `GetLinks()`. This makes the codes heavily relies on the corresponding **implementation** of function `GetLinks()`, adding unnecessary burden to write and maintain codes. Imagine someone modifies function `GetLinks()` in the future -- there's no text prompting him/her **not** returning an empty list.

MahBoiDeveloper: I can parry our argument with `foreach` cycle next to the line, where you started this conversation. Even if the impelentation would be changed by someone, and it would be return string list with zero elements in it, code wouldn't break.

DXMainClient/DXGUI/Multiplayer/ChatListBox.cs
MahBoiDeveloper: Until chat messages would support markdown or highlight links, this implementation is good enough.

DXMainClient/DXGUI/Multiplayer/ChatListBox.cs
MahBoiDeveloper: Until chat messages support markdown or highlight links, this implementation is good enough.

SadPencil: > Until chat messages support markdown or highlight links, this implementation is good enough.

"robust" means reliable and bug-free, instead of powerful. multiple messageboxes cause the messagebox buttons sometimes unclickable. nothing to do with markdown or highlight things

SadPencil: Also please leave the "resolve conversation" button to the reviewer, unless it's from copilot or a consensus has been clearly reached from the dialog

SadPencil:
> @SadPencil I'll re-post the convo I had with @MahBoiDeveloper from Discord on the multiple link handling.

I‘ll prefer a window (allowing the user select one of the URLs to open) instead of a context menu since it's weird that double-clicking results in a context menu

Otherwise, for a simple implementation, if there are more than one URLs, maybe the client can refuse to open links in this case.

Metadorius:
> I‘ll prefer a window (allowing the user select one of the URLs to open) instead of a context menu since it's weird that double-clicking results in a context menu

I have seen such behavior somewhere. Besides, menu bars in software often also open a menu on left click, so I don't see what's wrong with that. Maybe unusual but not illogical, but then not being able to click on individual links is unusual by itself. And I would argue that making a dialog with dynamic amount of link buttons is going to be more meh than that.

DXMainClient/DXGUI/Multiplayer/CnCNet/GlobalContextMenu.cs
MahBoiDeveloper: It's VS refactoring generated issue. Change has been reverted.

DXMainClient/DXGUI/Multiplayer/CnCNet/GlobalContextMenu.cs
MahBoiDeveloper: This check cannot be replaced because string comparison is the only way to determine whether the current context option is a link operation.

DXMainClient/DXGUI/Multiplayer/CnCNet/GlobalContextMenu.cs
```diff
@@ -153,24 +142,77 @@ private void UpdatePlayerBasedButtons()
...
+            foreach (string link in links)
+            {
+                string linkToDisplay = LINK_LENGTH > link.Length ? link.Substring(0, link.Length) : link.Substring(0, LINK_LENGTH) + "...";
+
+                if (Items.Where(item => item.Text.Contains(linkToDisplay)).ToList().Count > 0)
+                    continue;
+
+                var copyLinkItem = new XNAContextMenuItem()
+                {
+                    Text = string.Format("{0} {1}", COPY_LINK, linkToDisplay),
```
Metadorius:
```suggestion
                    Text = $"{COPY_LINK} {linkToDisplay}",
```

DXMainClient/DXGUI/Multiplayer/CnCNet/GlobalContextMenu.cs
```diff
@@ -153,24 +142,77 @@ private void UpdatePlayerBasedButtons()
...
+                    SelectAction = () => CopyLink(link)
+                };
+
+                var openLinkItem = new XNAContextMenuItem()
+                {
+                    Text = string.Format("{0} {1}", OPEN_LINK, linkToDisplay),
```
Metadorius:
```suggestion
                    Text = $"{OPEN_LINK} {linkToDisplay}",
```

DXMainClient/DXGUI/Multiplayer/ChatListBox.cs
```diff
@@ -53,17 +66,17 @@ private void ChatListBox_DoubleLeftClick(object sender, EventArgs e)
...
 
-                Please note that this link hasn't been verified,
-                and CnCNet is not responsible for its content.
+                    Please note that this link hasn't been verified,
+                    and CnCNet is not responsible for its content.
 
-                Would you like to open the following link in your browser?
-                """.L10N("Client:Main:OpenLinkConfirmationText")
+                    Would you like to open the following link in your browser?
+                    """.L10N("Client:Main:OpenLinkConfirmationText")
```
Metadorius: indentation went off

DXMainClient/DXGUI/Multiplayer/CnCNet/GlobalContextMenu.cs
```diff
@@ -153,24 +142,77 @@ private void UpdatePlayerBasedButtons()
...
+
+                            Please note that this link hasn't been verified,
+                            and CnCNet is not responsible for its content.
+
+                            Would you like to open the following link in your browser?
+                            """.L10N("Client:Main:OpenLinkConfirmationText")
+                            + Environment.NewLine + Environment.NewLine + link,
+                            XNAMessageBoxButtons.YesNo);
+                        msgBox.YesClickedAction = (msgBox) => ProcessLauncher.StartShellProcess(link);
+                        msgBox.Show();
```
Metadorius: I think this should be made into a function or something because it repeats

DXMainClient/DXGUI/Multiplayer/CnCNet/GlobalContextMenu.cs
Metadorius: IIRC you have the `Tag` object or something like this. You could assign some special const value and check it. This would be cleaner.

Metadorius (changes requested):

DXMainClient/DXGUI/Multiplayer/ChatListBox.cs
MahBoiDeveloper: fixed

DXMainClient/DXGUI/Multiplayer/CnCNet/GlobalContextMenu.cs
MahBoiDeveloper: No, because this is the only second place where it used.

DXMainClient/DXGUI/Multiplayer/CnCNet/GlobalContextMenu.cs
MahBoiDeveloper: > IIRC you have the `Tag` object or something like this. You could assign some special const value and check it. This would be cleaner.

Ehh, no. `XNAContextMenuItem` is a XNAUI class and it doesn't have any `Tag` property, what can be used to define what type of information item contains.

DXMainClient/DXGUI/Multiplayer/CnCNet/GlobalContextMenu.cs
Metadorius: > No, because this is the only second place where it used.

I mean, that's exactly the point when such an amount of lines/code should be split out into a function. Imagine someone adjusting one and forgetting to adjust another.

MahBoiDeveloper: > I mean, that's exactly the point when such an amount of lines/code should be split out into a function. Imagine someone adjusting one and forgetting to adjust another.

Feel free to reduce another code review step with discussion about what class should contain this message box and extract it by yourself.

DXMainClient/DXGUI/Multiplayer/ChatListBox.cs
SadPencil: ~~hint: this comment is unsolved. don't press the resolve button, please~~

SadPencil: sorry. You have resolved the issue by moving the message box into a context menu

Metadorius: > Also please leave the "resolve conversation" button to the reviewer, unless it's from copilot or a consensus has been clearly reached from the dialog

I told him that he can use the button instead of writing "done" on every one. IMO it's not a problem to un-resolve the conversation?

SadPencil: > > Also please leave the "resolve conversation" button to the reviewer, unless it's from copilot or a consensus has been clearly reached from the dialog
> 
> I told him that he can use the button instead of writing "done" on every one. IMO it's not a problem to un-resolve the conversation?

But if there are 10 comments for example, allowing PR proposer to use this button requiring the reviewer memorize what issues are truly solved and what issues need a check

DXMainClient/DXGUI/Multiplayer/CnCNet/GlobalContextMenu.cs
SadPencil: 1aacf142

Metadorius (approved):

PR #768, opened 2025-06-28
Refactor ClientCore: Improved Code Quality & Bug Fixes Core Improvements

qyjoy:
Refactor ClientCore: Improved Code Quality & Bug Fixes Core Improvements:
ProfanityFilter.cs
Critical Bug Fix:
Fixed a major logic issue in the IsOffensive method. Previously, it only flagged exact matches (i.e., the entire string had to be a banned word). This has been corrected to properly detect any occurrence of offensive terms within the input text.

Pattern Matching Overhaul:
Rewrote the ToRegexPattern method, which converts wildcard strings into regex. The old logic was ambiguous and overly complex. The new implementation is cleaner, more robust, and accurately handles wildcard characters like * and ?.

LoadingScreenController.cs
Randomness Fix:
Addressed improper usage of the Random class. Instantiating new Random() on every call caused duplicate outputs under high frequency. Replaced with a static readonly instance to ensure proper random number generation.

Cleaner Control Flow:
Replaced outdated if-else-if chains with C# 8 switch expressions. The result is more concise, readable logic.

SavedGameManager.cs
Improved Maintainability:

Migrated all string.Format() calls to modern string interpolation ($"") for clarity.

Removed magic number 1000 and replaced it with a named private const. This makes the purpose of the constant obvious and future changes safer.

FileHelper.cs
Better Diagnostics (Linux/macOS):
Enhanced error messages for hard link creation failures. Instead of cryptic error codes, users now get human-readable descriptions, making debugging much easier on Unix-based systems.

### DIFF
--- a/ClientCore/FileHelper.cs
+++ b/ClientCore/FileHelper.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.IO;
 using System.Linq;
 using System.Runtime.InteropServices;
@@ -16,41 +17,16 @@ namespace ClientCore
 {
     public class FileHelper
     {
-
-        /// <summary>
-        /// Establishes a hard link between an existing file and a new file. This function is only supported on the NTFS file system, and only for files, not directories.
-        /// <br/>
-        /// https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createhardlinkw
-        /// </summary>
-        /// <param name="lpFileName">The name of the new file.</param>
-        /// <param name="lpExistingFileName">The name of the existing file.</param>
-        /// <param name="lpSecurityAttributes">Reserved; must be NULL.</param>
-        /// <returns>If the function succeeds, the return value is nonzero. If the function fails, the return value is zero (0).</returns>
         [DllImport("Kernel32.dll", SetLastError = true, CharSet = CharSet.Unicode, EntryPoint = "CreateHardLinkW")]
         [DefaultDllImportSearchPaths(DllImportSearchPath.System32)]
         [SupportedOSPlatform("windows")]
         private static extern bool CreateHardLink(string lpFileName, string lpExistingFileName, IntPtr lpSecurityAttributes);
 
-        /// <summary>
-        /// The link function makes a new link to the existing file named by oldname, under the new name newname.
-        /// <br/>
-        /// https://www.gnu.org/software/libc/manual/html_node/Hard-Links.html
-        /// <param name="oldname"></param>
-        /// <param name="newname"></param>
-        /// <returns>This function returns a value of 0 if it is successful and -1 on failure.</returns>
         [DllImport("libc", EntryPoint = "link", SetLastError = true)]
         [SupportedOSPlatform("linux")]
         [SupportedOSPlatform("osx")]
         private static extern int link([MarshalAs(UnmanagedType.LPUTF8Str)] string oldname, [MarshalAs(UnmanagedType.LPUTF8Str)] string newname);
 
-        /// <summary>
-        /// Creates hard link to the source file or copy that file, if got an error.
-        /// </summary>
-        /// <param name="source"></param>
-        /// <param name="destination"></param>
-        /// <param name="fallback"></param>
-        /// <exception cref="IOException"></exception>
-        /// <exception cref="PlatformNotSupportedException"></exception>
         public static void CreateHardLinkFromSource(string source, string destination, bool fallback = true)
         {
             if (fallback)
@@ -83,8 +59,7 @@ namespace ClientCore
             else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux) || RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
             {
                 if (link(source, destination) != 0)
-                    throw new IOException(string.Format("Unable to create hard link at {0} with the following error code: {1}"
-                        .L10N("Client:DTAConfig:CreateHardLinkFailed"), destination, Marshal.GetLastWin32Error()));
+                    throw new IOException($"{"Client:DTAConfig:CreateHardLinkFailed".L10N()}: {destination}. Error: {new Win32Exception(Marshal.GetLastWin32Error()).Message}");
             }
             else
             {
--- a/ClientCore/LoadingScreenController.cs
+++ b/ClientCore/LoadingScreenController.cs
@@ -5,18 +5,19 @@ namespace ClientCore
 {
     public static class LoadingScreenController
     {
+        private static readonly Random _random = new Random();
+        
         public static string GetLoadScreenName(string sideId)
         {
             int resHeight = UserINISettings.Instance.IngameScreenHeight;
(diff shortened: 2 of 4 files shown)


ClientCore/FileHelper.cs
```diff
@@ -16,41 +17,16 @@ namespace ClientCore
...
         private static extern int link([MarshalAs(UnmanagedType.LPUTF8Str)] string oldname, [MarshalAs(UnmanagedType.LPUTF8Str)] string newname);
 
-        /// <summary>
-        /// Creates hard link to the source file or copy that file, if got an error.
-        /// </summary>
-        /// <param name="source"></param>
-        /// <param name="destination"></param>
-        /// <param name="fallback"></param>
-        /// <exception cref="IOException"></exception>
-        /// <exception cref="PlatformNotSupportedException"></exception>
```
MahBoiDeveloper, line 53: Do not remove these comments

ClientCore/SavedGameManager.cs
```diff
@@ -54,20 +52,14 @@ public static List<string> GetSaveGameTimestamps()
 
         public static bool AreSavedGamesAvailable()
         {
-            if (Directory.Exists(GetSaveGameDirectoryPath()))
-                return true;
-
-            return false;
+            return Directory.Exists(GetSaveGameDirectoryPath());
         }
```
MahBoiDeveloper, line 56: If you want to improve `AreSavedGamesAvailable` use lambda expression then

ClientCore/FileHelper.cs
```diff
@@ -83,8 +59,7 @@ public static void CreateHardLinkFromSource(string source, string destination, b
             else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux) || RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
             {
                 if (link(source, destination) != 0)
-                    throw new IOException(string.Format("Unable to create hard link at {0} with the following error code: {1}"
-                        .L10N("Client:DTAConfig:CreateHardLinkFailed"), destination, Marshal.GetLastWin32Error()));
+                    throw new IOException($"{"Client:DTAConfig:CreateHardLinkFailed".L10N()}: {destination}. Error: {new Win32Exception(Marshal.GetLastWin32Error()).Message}");
```
MahBoiDeveloper, line 62: Explain this change more, please.

Why exception message has changed to something controversial with other client's code?

ClientCore/LoadingScreenController.cs
```diff
@@ -5,18 +5,19 @@ namespace ClientCore
 {
     public static class LoadingScreenController
     {
+        private static readonly Random _random = new Random();
```
SadPencil, line 8: See https://github.com/CnCNet/xna-cncnet-client/pull/728 for the most preferred way to get the random object (dependency injection, if possible). We don't create a new one anymore

ClientCore/ProfanityFilter.cs
```diff
@@ -8,9 +8,6 @@ public class ProfanityFilter
     {
         public IList<string> CensoredWords { get; private set; }
 
-        /// <summary>
-        /// Creates a new profanity filter with a default set of censored words.
-        /// </summary>
```
SadPencil, line 13: Don't remove XML comments. Same for other unmentioned deletions.

ClientCore/FileHelper.cs
SadPencil, line 62: (a) The client heavily uses `string.Format` for localization and your modification breaks it
(b) You can show a friendly message about the error but please also keep showing the error code as an integer

ClientCore/ProfanityFilter.cs
```diff
@@ -75,16 +68,21 @@ private static string StarCensoredMatch(Match m)
...
+            if (startsWithWildcard && endsWithWildcard)
+                return pattern; 
+            
+            if (startsWithWildcard)
+                return pattern + @"\b";
+
+            if (endsWithWildcard)
+                return @"\b" + pattern;
+
+            return @"\b" + pattern + @"\b";
```
SadPencil, line 85: Can you explain the changes here and prove this method returns the same result as if this PR were not merged? Or, if this is a bug fix instead of a refactor, please demonstrate in which situation the old codes led to a bug with an example

ClientCore/SavedGameManager.cs
```diff
@@ -42,7 +40,7 @@ public static List<string> GetSaveGameTimestamps()
 
             for (int i = 0; i < saveGameCount; i++)
             {
-                FileInfo sgFile = SafePath.GetFile(saveGameDirectory, string.Format("SVGM_{0}.NET", i.ToString("D3")));
+                FileInfo sgFile = SafePath.GetFile(saveGameDirectory, $"SVGM_{i:D3}.NET");
```
SadPencil, line 43: Honestly we are tired of such a change. If you have modified surrounding codes, you can change it but otherwise, I suggest not doing so to reduce possible pains in git rebase/cherrypicks etc. But I guess we can torelate this one for consistency.

ClientCore/LoadingScreenController.cs
```diff
@@ -5,18 +5,19 @@ namespace ClientCore
...
-            else
-                resolutionText = "600";
+            int randomInt = _random.Next(1, 1 + ClientConfiguration.Instance.LoadingScreenCount);
+            
+            string resolutionText = resHeight switch
+            {
+                < 480 => "400",
+                < 600 => "480",
+                _ => "600"
+            };
```
SadPencil, line 20: Honestly we are tired of such a grammatical-only change. If you have modified surrounding codes, you can change it but otherwise, I suggest not doing so to reduce possible pains in git rebase/cherrypicks etc.

SadPencil (changes requested):

SadPencil:
Besides, it's advise not using the default branch (`develop` for this repo. Other possible name for other repos: `main`, `master`) to propose a PR in case you have pushed more commits in accident. Create a new branch instead.

ClientCore/ProfanityFilter.cs
```diff
@@ -34,20 +31,16 @@ public ProfanityFilter()
...
         {
-            string censoredText = text;
             foreach (string censoredWord in CensoredWords)
             {
                 string regularExpression = ToRegexPattern(censoredWord);
-                censoredText = Regex.Replace(censoredText, regularExpression, "",
-                    RegexOptions.IgnoreCase | RegexOptions.CultureInvariant);
-
-                if(string.IsNullOrEmpty(censoredText))
+                if (Regex.IsMatch(text, regularExpression, RegexOptions.IgnoreCase | RegexOptions.CultureInvariant))
```
SadPencil, line 43: Nice catch. I appreciate this bug fix

ClientCore/SavedGameManager.cs
qyjoy: gotcha but why, already simplified, we dont need lambda totally here

MahBoiDeveloper: Because it's the code style that we use in client for such cases. See ClientCore/Extensions/StringExtension/ToEnum.

Metadorius: It's not "lambda expression", it's "expression body. You have a one-line method, use expression body.

PR #774, opened 2025-07-05
Implement Migration Tool

### LINKED ISSUE #750: Migration tool
### Description

With the release of version 2.11.0.0, me and other modders were faced with the problem of having to manually change the configurations to make the client compatible with the changes in newer versions. The situation is grim because it is difficult to remember each change, and it is also difficult to complete the migration from the 2.6.x.x versions without spending tenths hours on the process.

As you can see, we need to create a tool to automate migration process for configs from old versions to the latest. I suggest that this tool should be a separate executable file that updates config files, adds new ones and makes changes to old ones, as well as adding new placeholder pictures.

MahBoiDeveloper:
Closes #750 

## To Do List

- [x] Write patches for each release with major config changes
- [x] Add embedded missing images and copy them if they are absent in `Resource` folder
- [x] Fix `Migration-INI.md` wrong `Edit GlobalThemeSettings.ini` section
- [x] Add documentation with migration tool description and detailed steps for migration workflow
- [x] Add .NET 4.8 version of the tool
- [ ] Add negotiation icons
- [ ] Test MT with old client configs samples
  - [ ] Mental Omega 3.3.6
  - [ ] Twisted Insurrection 0.9.0.x
  - [ ] Tiberian Sun Client
  - [ ] YR client package
  - [ ] YR Mod Base
  - [ ] Rise Of The East
  - [ ] A mysterious client config from SadPencil (not provided yet)


MigrationTool/Program.cs
```diff
@@ -0,0 +1,277 @@
...
+                    if (clientDefsIni.KeyExists("Settings", "TrustedDomains"))
+                    {
+                        Log("Update ClientDefinitions.ini: Skip add [Settings]->TrustedDomains, reason: already exist");
+                        continue;
+                    }
+
+                    var td = "moddb.com";
+                    Log($"Update ClientDefinitions.ini: Add [Settings]->TrustedDomains={td}");
+                    clientDefsIni.GetSection("Settings").AddKey("TrustedDomains", td);
+                    clientDefsIni.WriteIniFile();
```
Metadorius: don't add keys that are not *needed*

MigrationTool/Program.cs
```diff
@@ -0,0 +1,277 @@
...
+        Begin,
+        V_2_11_0,
+        V_2_11_1,
+        V_2_11_2,
+        V_2_12_0,
+        V_2_12_1,
+        V_2_12_5,
+        Latest,
+        End
+    }
```
Metadorius: using enum for this looks like an overcomplication without actual use to me

MigrationTool/Program.cs
MahBoiDeveloper: Deleted

MigrationTool/Program.cs
MahBoiDeveloper: I don't think so because it allows step-by-step migration from one version to another and to the latest in the end.

Metadorius: this is not inherent to enums only, it's one of the ways to do that. instead just make each version migration a function and call it

MigrationTool/MigrationTool.csproj
```diff
@@ -0,0 +1,14 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <UseAppHost>false</UseAppHost>
+  </PropertyGroup>
+  <PropertyGroup>
+    <Title>CnCNet.MigratioTool</Title>
+    <Description>CnCNet Client Migration Tool</Description>
+    <Product>CnCNet.MigratioTool</Product>
```
SadPencil: typo

MigrationTool/Program.cs
```diff
@@ -0,0 +1,279 @@
...
+                    AddKeyWithLog(playerExtraOptionsPanelIni, "lblPreset",                  "Location", "12,156");
+                    AddKeyWithLog(playerExtraOptionsPanelIni, "ddTeamStartMappingPreset",   "Location", "65,154");
+                    AddKeyWithLog(playerExtraOptionsPanelIni, "ddTeamStartMappingPreset",   "Size",     "157,21");
+                    AddKeyWithLog(playerExtraOptionsPanelIni, "teamStartMappingsPanel",     "Location", "12,189");
+
+                    continue;
+
+                case (Version.V_2_11_1): // https://github.com/CnCNet/xna-cncnet-client/releases/tag/2.11.1.0
+                    // Add ClientDefinitions.ini->[Settings]->RecommendedResolutions
+                    AddKeyWithLog(clientDefsIni, "Settings", "RecommendedResolutions", "1280x720");
```
SadPencil: Determine if 1280x720 can fit in MaximumRenderWidth x MaximumRenderHeight

MigrationTool/Program.cs
```diff
@@ -0,0 +1,279 @@
...
+                    }
+
+                    // Add ClientDefinitions.ini->[Settings]->ShowDevelopmentBuildWarnings
+                    AddKeyWithLog(clientDefsIni, "Settings", "ShowDevelopmentBuildWarnings", "true");
+                    clientDefsIni.WriteIniFile();
+                    continue;
+
+                case (Version.V_2_12_0): // https://github.com/CnCNet/xna-cncnet-client/releases/tag/2.12.0
+                    // Remove Rampastring.Tools from Resources directory (not recursive)
+                    Log("Remove Resources/Rampastring.Tools.* (* -- dll, pdb, xml)");
```
SadPencil: Resources/Rampastring.Tools.dll prevents the client from launching since the ~2.8.0.0 chaos. Not from v2.12.0 only

MigrationTool/Program.cs
```diff
@@ -0,0 +1,279 @@
...
+
+                    AddKeyWithLog(clientDefsIni, "Settings", "ClientGameType", cgt);
+                    clientDefsIni.WriteIniFile();
+                    continue;
+
+                case (Version.Latest):
+                    // Add GameLobbyBase.ini->[ddPlayerColorX]->ItemsDrawMode
+                    gmLobbyBaseIni ??= new IniFile(SafePath.CombineFilePath(resouresDir.FullName, "GameLobbyBase.ini"));
+                    string ddPlayerColor = nameof(ddPlayerColor);
+                    foreach (var n in new int[] { 0, 1, 2, 3, 4, 5, 6, 7 })
```
SadPencil: Enumerate.Range()

SadPencil (changes requested):

MigrationTool/Program.cs
```diff
@@ -0,0 +1,279 @@
...
+            """;
+
+        Console.WriteLine(text);
+    }
+
+    private static void AddKeyWithLog(IniFile src, string section, string key, string value)
+    {
+        if (src.KeyExists(section, key))
+        {
+            Log($"Update {src.FileName}.ini: Skip add [{section}]->{key}, reason: already exist");
```
SadPencil: Skip add -> Skip adding

MigrationTool/Program.cs
```diff
@@ -0,0 +1,279 @@
...
+                    SafePath.DeleteFileIfExists(resouresDir.FullName, "Rampastring.Tools.xml");
+                    continue;
+
+                case (Version.V_2_12_1): // https://github.com/CnCNet/xna-cncnet-client/releases/tag/2.12.1
+                    // And add ClientDefinitions.ini->[Settings]->ClientGameType
+                    string cgt = clientGameType switch
+                    {
+                        ClientGameType.Ares => "Ares",
+                        ClientGameType.YR => "YR",
+                        _ => "TS"
```
SadPencil: ClientGameType.TS => "TS",
_ => new Exception("Unknown client game type")

MigrationTool/MigrationTool.csproj
MahBoiDeveloper: Done by ae8ed1618fbd57948a230173ca7fcceac3cd113c.

MigrationTool/Program.cs
MahBoiDeveloper: Done by b7ddb2f16a7d169b06390d01bf52425aefc4e2fa.

MigrationTool/Program.cs
MahBoiDeveloper: Done by 96d9225521de5f3b0296c177551f0925f6757530.

MigrationTool/Program.cs
MahBoiDeveloper: Done by 96d9225521de5f3b0296c177551f0925f6757530.

MigrationTool/Program.cs
MahBoiDeveloper: > ClientGameType.TS => "TS", _ => new Exception("Unknown client game type")

Default client type should be TS and do not crash the whole migration process. Also this isn't valid syntax for switch expression.

MigrationTool/Program.cs
MahBoiDeveloper: > this is not inherent to enums only, it's one of the ways to do that. instead just make each version migration a function and call it

Then better to refactor patch apply mechanism in the end of the PR before merge.

MigrationTool/Program.cs
MahBoiDeveloper: Check b5f5f574d67ac01fe1de2d9686a2403679f790b0.

SadPencil: Specify MaximumRenderWidth and MaximumRenderHeight first. (read it; create it only if it does not exist)

Then, if width and height are respectively greater or equal than 1280 720, write RecommendedResolutions as 1280x720, "{MaximumRenderWidth}x{MaximumRenderHeight}" otherwise

MigrationTool/Program.cs
SadPencil:
```cs
MyType t = MyType.A;
string result = t switch
{
    MyType.A => "A",
    MyType.B => "B",
    _ => throw new Exception("")
};
Console.Write(result);

enum MyType {A,B}
```

SadPencil: 图片

SadPencil: Why specifying TS as the default type? Also, "do not crash" is not a correct behavior if something unexpected happen, even though it could never happens.

ZivDero: Uhm... How about just calling `.ToString()` on the enum? Why is this even necessary. And there shouldn't be any other unhandled enum values in the first place, that's what the enum is for...

MahBoiDeveloper: > Uhm... How about just calling `.ToString()` on the enum? Why is this even necessary. And there shouldn't be any other unhandled enum values in the first place, that's what the enum is for...

image

Seems reasonable.

SadPencil: > And there shouldn't be any other unhandled enum values in the first place, that's what the enum is for...

e.g., let's suppose we updated the enum to 4 values, Ares, TS, YR, and Dune. If we have write an exception on unrecognizable enum, we will be notified by a crash if one of the code doesn't get updated. Otherwise, we'll just silently get an incorrect value without knowing it.

MigrationTool/Program.cs
SadPencil: I cancelled the "resolve conversation" button since this change has not been made.

@Metadorius This is why I insist on leaving this button to reviewers. We can have more than 10 conversations, and if we don't use this feature to label what have done and what remained to be done, we'll forget them soon.

Metadorius: My intention when telling it to mah_boi was to mark the actually resolved feedback, i.e. when all participants are obviously agreeing. Like for small corrections that don't require any further input and were addressed, not for actually debatable changes and resolutions. In those cases obviously one shouldn't resolve the conversation and better leave that to maintainers.

MigrationTool/Program.cs
ZivDero: > > And there shouldn't be any other unhandled enum values in the first place, that's what the enum is for...
> 
> e.g., let's suppose we updated the enum to 4 values, Ares, TS, YR, and Dune. If we have write an exception on unrecognizable enum, we will be notified by a crash if one of the code doesn't get updated. Otherwise, we'll just silently get an incorrect value without knowing it.

The crash doesn't have to be in this specific part of the code, though, it has nothing to do with converting the enum to the string. I support the idea that checking to make sure the code supports all the values of the enum, but this isn't the place to do it.

ZivDero: IOW I think just placing something along the lines of

```cs
switch (clientGameType)
{
    case ClientGameType.TS:
    case ClientGameType.YR:
    case ClientGameType.Ares:
        break;
		
    default:
        throw new InvalidArgumentException("An unsupported game type has been specified!");
        break;
}
```
elsewhere would be a better idea.

MigrationTool/Patch_v2_11_0.cs
```diff
@@ -0,0 +1,423 @@
...
+
+            // Old configs
+            IniFile skirmishLobbyIni_old        = new IniFile(SafePath.CombineFilePath(ResouresDir.FullName, $"{SkirmishLobby}.ini"));
+            IniFile multiplayerGameLobbyIni_old = new IniFile(SafePath.CombineFilePath(ResouresDir.FullName, $"{MultiplayerGameLobby}.ini"));
+            IniFile gameOptionsIni              = new IniFile(SafePath.CombineFilePath(ResouresDir.FullName, "GameOptions.ini"));
+
+            // New configs
+            IniFile gameLobbyBaseIni         = new IniFile(SafePath.CombineFilePath(ResouresDir.FullName, $"{GameLobbyBase}.ini"));
+            IniFile skirmishLobbyIni         = new IniFile(SafePath.CombineFilePath(ResouresDir.FullName, $"{SkirmishLobby}_New.ini"));
+            IniFile multiplayerGameLobbyIni  = new IniFile(SafePath.CombineFilePath(ResouresDir.FullName, $"{MultiplayerGameLobby}_New.ini"));
```
SadPencil: ~~New? Why not moving the old lobby files as "_Old" and using the new ini files by default? There're no prompting on manual replacement or something~~

SadPencil: It appears that the client automatically override the old files. never mind

MigrationTool/Program.cs
```diff
@@ -0,0 +1,99 @@
...
+                    patch = new Patch_v2_11_0(arg).Apply(); Console.WriteLine("");
+                    patch = new Patch_v2_11_1(arg).Apply(); Console.WriteLine("");
+                    patch = new Patch_v2_11_2(arg).Apply(); Console.WriteLine("");
+                    patch = new Patch_v2_12_1(arg).Apply(); Console.WriteLine("");
+                    patch = new Patch_Latest(arg).Apply();
+                }
+                catch (Exception ex)
+                {
+                    Logger.Log("");
+                    Logger.Log($"Unable to apply migration patch for client version {patch.ClientVersion.ToString().Replace('_', '.')} due to internal error. Message: {ex.Message}");
```
SadPencil: I got a System.NullReferenceException on this line.

patch is null

write this instead

```cs
patch = new Patch_v2_11_0(arg); patch.Apply();
```

SadPencil:
Can you provide a way to launch this tool in a Debug mode in Visual Studio?

MigrationTool/Program.cs
SadPencil: [bdbe183](https://github.com/CnCNet/xna-cncnet-client/pull/774/commits/bdbe183dc0c41ab2ae9968a0f0193103f300a1d5)

SadPencil:
I have started thinking maybe we should make this tool a standalone project?

MahBoiDeveloper:
> Can you provide a way to launch this tool in a Debug mode in Visual Studio?

Add in args path to the folder with the client. For example:

```csharp
args = new string[] { @"C:\MO" };
```

MahBoiDeveloper:
> Seems like something are also broken
> 
> before migration
> 
> after migration

Fixed by fe4b62edc80c4ae9b757cad400feb08ce7f00a25

SadPencil:
> > Seems like something are also broken
> 
> > 
> 
> > before migration
> 
> > 
> 
> > after migration
> 
> 
> 
> Fixed by fe4b62edc80c4ae9b757cad400feb08ce7f00a25

actually not this border... I can't share the assets. Can you verify that on your machine the debug features including breakpoints all run as expected in Visual Studio?

MahBoiDeveloper:
> > > Seems like something are also broken
> > > before migration
> > > after migration
> > 
> > Fixed by [fe4b62e](https://github.com/CnCNet/xna-cncnet-client/commit/fe4b62edc80c4ae9b757cad400feb08ce7f00a25)
> 
> actually not this border... I can't share the assets. Can you verify that on your machine the debug features including breakpoints all run as expected in Visual Studio?

Sure

image

image

Metadorius:
The solution to this is pre-releases, like x.y.z.a-rc1 for example.

SadPencil:
> The solution to this is pre-releases, like x.y.z.a-rc1 for example.

Then at least I will release 2.12.6 before this and other major PR being merged.

SadPencil:
图片

MigrationTool/Patch_v2_11_1.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        if (presets == null)
+            return this;
+
+        presets.SectionName = "btnSaveLoadGameOptions";
+
+        foreach (var pair in glb.GetSection("SkirmishLobby").Keys)
+        {
+            if (pair.Value.Contains("BtnSaveLoadGameOptions"))
+                pair.Value.Replace("BtnSaveLoadGameOptions", "btnSaveLoadGameOptions");
```
SadPencil: CA1806 Apply calls Replace but does not use the new string instance that the method returns. Pass the instance as an argument to another method, assign the instance to a variable, or remove the call if it is unnecessary.

SadPencil: I prompted CA1806 as a warning (by default its level is info) and maybe we should even prompt this check as an error

SadPencil:
图片

exeFile?

SadPencil:
> > Seems like something are also broken
> > before migration
> > after migration
> 
> Fixed by [fe4b62e](https://github.com/CnCNet/xna-cncnet-client/commit/fe4b62edc80c4ae9b757cad400feb08ce7f00a25)

I manually reverted your change about `DrawBorders`. This border is not the same border in my screenshot, and the problem still exists.

SadPencil:
图片

I'll replace your `nameof()` with a regular string based on the null reference exception

SadPencil:
> ```
> Unable to apply migration patch for client version v2.11.0 due to internal error. Message: System.NullReferenceException: 未将对象引用设置到对象的实例。
>    在 MigrationTool.Patch.TransferKeys(IniFile srcIni, String srcSection, IniFile desIni, String desSection) 位置 D:\SadPencil\Documents\GitHub\xna-cncnet-client\MigrationTool\Patch.cs:行号 121
>    在 MigrationTool.Patch_v2_11_0.Apply() 位置 D:\SadPencil\Documents\GitHub\xna-cncnet-client\MigrationTool\Patch_v2_11_0.cs:行号 220
>    在 MigrationTool.Program.Main(String[] args) 位置 D:\SadPencil\Documents\GitHub\xna-cncnet-client\MigrationTool\Program.cs:行号 79
> Migration to the latest client version has been failed
> ```
> 
> Needs a way to run in a debugger. I can't share the assets

Note that this issue is fixed via changing `patch = new Patch_v2_11_0(arg).Apply();` to `patch = new Patch_v2_11_0(arg); patch.Apply();`. And therefore, the return type of `Apply()` method has been changed to void in de40f805

SadPencil (changes requested):
Remaining tasks:

- Assist Rampastring.Tools to allow editing the ini files without writing the parent ini to the child https://github.com/Rampastring/Rampastring.Tools/issues/15
- Don't remove the comments in an ini file
- Investigate why the message box border (not the border in DrawBorders!) in Mental Omega disappears with @SadPencil, perhaps in Auguest
- Cliam this tool is experimental and might have potential bugs even if it gets included in a stable release in xna-cncnet-client. Show a prompt that modders must backup their files and test in through after applying this migration tool and we have absolutely no warrently.
- Address the wrong usage of `string.Replace()` method that triggers a CA1806 warning

Metadorius:
> * Don't remove the comments in an ini file

that one would be hard

MahBoiDeveloper:
> Note that this null reference issue was fixed via changing `patch = new Patch_v2_11_0(arg).Apply();` to `patch = new Patch_v2_11_0(arg); patch.Apply();`. And therefore, the return type of `Apply()` method has been changed to void in [de40f80](https://github.com/CnCNet/xna-cncnet-client/commit/de40f805e49e08056097ed948e5fc1c3c679e111)

Check new implementation that removes needs to manually add every new patch in apply list.

SadPencil:
> > What if we released 2.12.10?
> 
> It would be applied after 2.12.9 and before 2.12.11 and before latest patch. Latest patch closes the list.



> > What if we released 2.12.10?
> 
> It would be applied after 2.12.9 and before 2.12.11 and before latest patch. Latest patch closes the list.

It wouldn't be applied after 2.12.9 and before 2.12.11.

```cs
List<string> names = ["2.12.10", "2.12.9", "2.12.9-rc1"];
names=names.OrderBy(name=>name).ToList();
Console.Write(string.Join(", ", names));
```

Output:
```
2.12.10, 2.12.9, 2.12.9-rc1
```

That's why I suggest using a list of strings

MigrationTool/Program.cs
```diff
@@ -0,0 +1,108 @@
...
+                try
+                {
+                    // https://stackoverflow.com/questions/16038819/how-to-find-all-direct-subclasses-of-a-class-with-net-reflection
+                    assembly.GetTypes()
+                        .Where(type => type.BaseType == typeof(Patch))
+                        .OrderBy(type => type.FullName) // Used to order patches by their names
+                        .ToList()
+                        .ForEach(type => 
+                        {
+                            var patch = (Patch)Activator.CreateInstance(type, arg);
```
SadPencil: Note that you need to save the `patch` variable outside of the try-catch scope. `patch` is always null in the catch block now.

```cs
var newPatch = CreateInstance();
patch = newPatch;
patch.Apply();
```

MahBoiDeveloper:
> That's why I suggest using a list of strings

Set this list manually is a mistake. But I have found another way to organize this list. We can just use enum extension to get all values of `Version` enum.

image

SadPencil:
> > That's why I suggest using a list of strings
> 
> 
> 
> Set this list manually is a mistake. But I have found another way to organize this list. We can just use enum extension to get all values of `Version` enum.
> 
> 
> 
> image
> 
> 
> 
> 

seems good

SadPencil:
Including clientcore? What if polyfill requires a lot of dll files (like the updater in 2.12.0)

SadPencil:
> > Including clientcore? What if polyfill requires a lot of dll files (like the updater in 2.12.0)
> 
> 
> 
> It is a good question how to include already existed enums into migration tool.

Just repeat the definition for now? I don't have a better way

MahBoiDeveloper:
image

This doesn't feel right

SadPencil:
> image
> 
> 
> 
> This doesn't feel right

yeah. Let's just repeat the definition.

Metadorius:
Why can't you reuse the same libs the client uses?

BTW, this is one of the reasons I proposed for it to be a part of the client invoked as a command line parameter.

Metadorius:
> Client assembly differences with libraries between different versions (like `Rampastring.Tools.dll` near to `clientdx.exe` that crash client on start) should not affect the migration tool. That's why I decided to make it as cli-tool in executable file.

Do we really need to inconvenience ourselves that much because of issues of wrong binaries? IMO we would save a lot of hassle by not handling the problems with binaries. The migration tool doesn't apply 100% of the needed changes anyway, so it's not like we will lose a lot by not handling this.

MigrationTool/Program.cs
```diff
@@ -0,0 +1,91 @@
...
+                var assembly = Assembly.GetExecutingAssembly();
+                Patch? patch = null;
+                try
+                {
+                    // https://stackoverflow.com/questions/16038819/how-to-find-all-direct-subclasses-of-a-class-with-net-reflection
+                    var patches = assembly.GetTypes().Where(type => type.BaseType == typeof(Patch)).ToList();
+                    var patchNames = Enum.GetValues(typeof(Version));
+                    foreach (var patchName in patchNames)
+                    {
+                        Type type = patches.Where(t => t.FullName.Contains(patchName.ToString())).First();
```
SadPencil: Contains? Let's say `Patch_v2_12_9` and `Patch_v2_12_9-rc_1`

MahBoiDeveloper: Reworked, check new implementation

MahBoiDeveloper:
@SadPencil Would you like to help with WindowsXNA Release NET4.8 target build? I don't know how to fix compilation for this target.

SadPencil:
> @SadPencil Would you like to help with WindowsXNA Release NET4.8 target build? I don't know how to fix compilation for this target.

I guess you'll have to consult @frg2089 for this error. Maybe you can also try moving the tool inside the DXMainClient project to avoid touching the build files

MigrationTool/Patch_v2_12_6.cs
```diff
@@ -0,0 +1,32 @@
...
+        base.Apply();
+
+        // Add GameLobbyBase.ini->[ddPlayerColorX]->ItemsDrawMode
+        IniFile gmLobbyBaseIni = new IniFile(SafePath.CombineFilePath(ResouresDir.FullName, "GameLobbyBase.ini"));
+        string ddPlayerColor = nameof(ddPlayerColor);
+        Enumerable.Range(0, 8).ToList().ForEach(i => AddKeyWithLog(gmLobbyBaseIni, ddPlayerColor + i, "ItemsDrawMode", "Text"));
+        gmLobbyBaseIni.WriteIniFile();
+
+        // Remove GenericWindow.ini->{ GameCreationWindow; GameCreationWindow_Advanced }->Size
+        IniFile genericWindowIni = new IniFile(SafePath.CombineFilePath(ResouresDir.FullName, "GenericWindow.ini"));
```
MortonPL, line 25: This change is missing a log message.

MigrationTool/Patch_v2_11_0.cs
```diff
@@ -0,0 +1,424 @@
...
+            addKey("teamStartMappingsPanel", "Location", "12,189");
+            playerExtraOptionsPanelIni.WriteIniFile();
+        }
+
+        string GameLobbyBase = nameof(GameLobbyBase);
+
+        // Rework skirmish/lan/cncnet lobbies ini's
+        if (File.Exists(SafePath.CombineFilePath(ResouresDir.FullName, $"{GameLobbyBase}.ini")))
+        {
+            Logger.Log($"Update lobbies has been aborted, {GameLobbyBase}.ini already exists");
```
MortonPL: I think that "skipped" would be a better word than "aborted" here, as we continue patching.

MigrationTool/Patch_v2_11_0.cs
```diff
@@ -0,0 +1,424 @@
...
+                    }
+
+                    if (value.Contains(":CustomSettingFileDropDown"))
+                    {
+                        optionsWindowIni.SetStringValue(section, key, value.Replace(":CustomSettingFileDropDown", ":FileSettingDropDown"));
+                        continue;
+                    }
+                }
+            }
+
```
MortonPL, line 70: This change is missing a log message.

Docs/Migration.md
```diff
@@ -1,6 +1,11 @@
 Migrating from older versions
 -----------------------------
 
+> [!NOTE]
+> CnCNet has a program called `Migration Tool` that implements all the instructions described down below notices to automate the migration process.
+> To use this tool, download the latest nightly build and run `Resources/BinariesNET8/MigrationTool/MigrationTool.exe` with older client path as the first argument.
+> Example: `MigrationTool.exe C:\TiberianSunClient`
+
```
MortonPL, line 10: I believe it's a good idea to remind users to backup their client before running the tool.

SadPencil:
This PR still breaks some settings but I can't share the corresponding client assets. Needs to wait for a certain period of time.

MigrationTool/Patch_v2_11_0.cs
MahBoiDeveloper: Good idea. Fixed.

MigrationTool/Patch_v2_11_0.cs
MahBoiDeveloper: Fixed in 31844b493684e115cea5d7ea47b2080ab3a97929

MigrationTool/Patch_v2_12_6.cs
MahBoiDeveloper: Add `RemoveKeyWithLog` method to log this.

Docs/Migration.md
MahBoiDeveloper: Added in 98ad9ebbf1897ea5e3fa6f07a3667a321e86875e

PR #785, opened 2025-07-09
​Add docs for `MapEncoding` translation key

MahBoiDeveloper:


Docs/Translation.md
```diff
@@ -56,9 +56,10 @@ The folder name doesn't explicitly need to match the existing locale code. Howev
 ### Translation INI format
 
 ```ini
-[General]            ; translation metadata
-Name=Some Language   ; string, used instead of a system-provided name if set
-Author=Someone       ; string
+[General]          ; translation metadata
+Name=Some Language ; string, used instead of a system-provided name if set
+Author=Someone     ; string
+MapEncoding=Auto   ; string, defines the name of the map encoding to be used to load the map files to the spawnmap.ini file. The 'Auto' option means that the client will try to guess the encoding.
```
SadPencil: By default it could be omitted or specified as UTF-8

Only specify `Auto` if you know what you are doing

SadPencil:
```suggestion
MapEncoding=UTF-8 ; string, defines the name of the map encoding to be used to load the map files to the spawnmap.ini file. The 'Auto' option means that the client will try to guess the encoding. Please either completely omit this line or use `UTF-8`. Only specify `Auto` or an encoding different from `UTF-8` if you really know what you are doing.
```

PR #805, opened 2025-07-26
Add ArabicSupport in XNA Client

G-LimeJuice:
an issue of reversed Arabic letters has been fixed but the fonts are still unclear and need improvement


MahBoiDeveloper:
Why PR has been closed?

PR #807, opened 2025-07-27
Arabic Supperts

G-LimeJuice:
an issue of reversed Arabic letters has been fixed but the fonts are still unclear and need improvement

### DIFF
--- a/ClientGUI/IME/IMEHandler.cs
+++ b/ClientGUI/IME/IMEHandler.cs
@@ -83,20 +83,53 @@ public abstract class IMEHandler : IIMEHandler
 
     public abstract void StopTextComposition();
 
+    private string imeText = "";
+    private string Storedtext = string.Empty;
     protected virtual void OnIMETextInput(char character)
     {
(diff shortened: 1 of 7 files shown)


ClientGUI/IME/IMEHandler.cs
```diff
@@ -83,20 +83,53 @@ public static IMEHandler Create(Game game)
...
     public abstract void StopTextComposition();
 
+    private string imeText = "";
+    private string Storedtext = string.Empty;
     protected virtual void OnIMETextInput(char character)
     {
         //Debug.WriteLine($"IME: OnIMETextInput: {character} {(short)character}; IMEFocus is null? {IMEFocus == null}");
 
-        IMEEventReceived = true;
-        LastActionIMEChatInput = true;
```
SadPencil, line 91: What are you doing

SadPencil:
I don't think this change can be merged. Please just clearly wrote what's the current issue as a bug report, providing necessary texts or even keyboard to reproduce, and let us to fix them.

PR #817, opened 2025-08-13
Make CnCNetGameCheck a singleton and improve Cheat Engine check

11EJDE11:
Three main changes here:

- Change to singleton. This is needed as I had an instance where a couple copies of the check were running at once. Not quite sure what caused that (maybe pausing debugger for too long, irc logged out).

- Stops the check from taking place if the game isn't open. This is safe as we only kill the game if a process is detected, so no point checking the processes if the game isn't running.

- Remove useless handle check: handle is an int so checking if it contains string is pointless.

----

The call to process.MainWindowTitle spikes the CPU and makes the check take a few seconds to run. We really should consider changes to that in future.

### DIFF
--- a/DXMainClient/Online/CnCNetGameCheck.cs
+++ b/DXMainClient/Online/CnCNetGameCheck.cs
@@ -83,8 +83,8 @@ namespace DTAClient.Online
                     process.Dispose();
                 }
             }
-            catch 
-            { 
+            catch
+            {
             }
         }
     }


DXMainClient/Online/CnCNetGameCheck.cs
```diff
@@ -75,8 +83,8 @@ private void KillGameInstance()
                     process.Dispose();
                 }
             }
-            catch
-            {
+            catch 
+            {
```
SadPencil: unnecessary white spaces introduced

SadPencil (approved):
There are two unnecessary white spaces. Please delete them before merging this PR

DXMainClient/Online/CnCNetGameCheck.cs
11EJDE11: Thanks, SadPencil. Fixed now.

PR #834, opened 2025-09-23
Remove XNAUI dependency from `ClientCore` project and remove `DTAConfig`

MahBoiDeveloper:
PR opens #774 that previously was blocked by ClientCore's dependency on XNAUI.


MahBoiDeveloper:
> `ClientGUI.Settings`? In my opinion the setting stuffs should be merged in DXMainClient project. CC @Metadorius 

They are generic UI elements extending functionality of XNAUI, so they moved to ClientGUI, but currently used by OptionsWindow only.

Metadorius:
@SadPencil Per my understanding `ClientGUI` should host separate client-related controls and components, and `DXMainClient` is a project where the final client is assembled together, so I instructed @MahBoiDeveloper to do the same in his PR.

DXMainClient/Domain/Multiplayer/CnCNet/GameCollection.cs
```diff
@@ -27,20 +28,20 @@ public void Initialize()
...
-            using Stream rrIconStream = assembly.GetManifestResourceStream("ClientCore.Resources.rricon.png");
-            using Stream reIconStream = assembly.GetManifestResourceStream("ClientCore.Resources.reicon.png");
-            using Stream cncrIconStream = assembly.GetManifestResourceStream("ClientCore.Resources.cncricon.png");
-            using Stream cncnetIconStream = assembly.GetManifestResourceStream("ClientCore.Resources.cncneticon.png");
-            using Stream tdIconStream = assembly.GetManifestResourceStream("ClientCore.Resources.tdicon.png");
-            using Stream raIconStream = assembly.GetManifestResourceStream("ClientCore.Resources.raicon.png");
-            using Stream d2kIconStream = assembly.GetManifestResourceStream("ClientCore.Resources.d2kicon.png");
-            using Stream ssIconStream = assembly.GetManifestResourceStream("ClientCore.Resources.ssicon.png");
-            using Stream unknownIconStream = assembly.GetManifestResourceStream("ClientCore.Resources.unknownicon.png");
+            using Stream dtaIconStream = assembly.GetManifestResourceStream("DTAClient.Icons.dtaicon.png");
```
SadPencil, line 31: Can we tolerate the old png names? e.g., make a private method like `LoadGameIcon("dtaicon")`, where the method loads the old name as a fallback. Otherwise it requires modders to rename these png files in order to upgrade the client

MahBoiDeveloper: > Otherwise it requires modders to rename these png files in order to upgrade the client

WDYM? All `DTAClient.Icons.*.png` are embedded files that stores in exe.

SadPencil: Oops. never mind then

Metadorius (approved):

PR #835, opened 2025-09-25
Account Login and Verification Support

CnCRAZER:
### Establishes 2 new flags within `ClientDefinitions.ini`:
- `UseCnCNetAPI`
- `CnCNetAPIUrl`

Setting `UseCnCNetAPI` to `Yes` or `True` allows online login window to connect to cncnet ladder account and retrieve active nicknames. Setting this flag to `No` or `False` hides this field entirely.

image

image

image

### Verified Account Only Option

image


DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetLobby.cs
```diff
@@ -1506,7 +1691,7 @@ private void GameBroadcastChannel_CTCPReceived(object sender, ChannelCTCPEventAr
             string msg = e.Message.Substring(5); // Cut out GAME part
             string[] splitMessage = msg.Split(new char[] { ';' });
 
-            if (splitMessage.Length != 12)
+            if (splitMessage.Length < 12)
```
SadPencil: huh? Why changing the `!=` operator instead of adding the number here?

ClientCore/ClientConfiguration.cs
```diff
@@ -389,6 +389,13 @@ public string GetGameExecutableName()
 
         public bool DisplayPlayerCountInTopBar => clientDefinitionsIni.GetBooleanValue(SETTINGS, "DisplayPlayerCountInTopBar", false);
 
+        public bool UseCnCNetAPI => clientDefinitionsIni.GetBooleanValue(SETTINGS, "UseCnCNetAPI", true);
+
+        /// <summary>
+        /// Base URL for the CnCNet API, typically ends with /api/v1/.
+        /// </summary>
+        public string CnCNetApiUrl => clientDefinitionsIni.GetStringValue(SETTINGS, "CnCNetApiUrl", "http://cncnet-api/api/v1/");
```
SadPencil: Inconsistent upper-case and lower-case spelling for `API`

UseCnCNetAPI
CnCNetApiUrl

CnCRAZER: Good catch!

DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetAccountLoginPrompt.cs
```diff
@@ -0,0 +1,90 @@
...
+        {
+            Name = nameof(CnCNetAccountLoginPrompt);
+            BackgroundTexture = AssetLoader.LoadTextureUncached("logindialogbg.png");
+            ClientRectangle = new Rectangle(0, 0, 350, 150);
+
+            lblWindowTitle = new XNALabel(WindowManager)
+            {
+                Name = "lblWindowTitle",
+                FontIndex = 1,
+                Text = "CONNECT TO CNCNET"
```
11EJDE11, line 32: L10N all the visible-to-user strings for translations.

CnCRAZER: I've pr'd this as draft with some things still needing to be polished. Can work on translations once everything seems good to go

DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetLobby.cs
```diff
@@ -1372,6 +1492,11 @@ private void DdCurrentChannel_SelectedIndexChanged(object sender, EventArgs e)
             currentChatChannel.UserQuitIRC += RefreshPlayerList;
             currentChatChannel.UserKicked += RefreshPlayerList;
             currentChatChannel.UserListReceived += RefreshPlayerList;
+            // Trigger verification after user list is received, if API is enabled
+            if (ClientConfiguration.Instance.UseCnCNetAPI)
+            {
+                currentChatChannel.UserListReceived += VerifyUserList;
```
11EJDE11, line 1498: UserListReceived is a list of users on the IRC channel with no ident info which would make verification impossible here if I'm not mistaken?

CnCRAZER: I believe so yeah. I used grants pr from 6 years ago for this and tried to update it. Having to go back and update things.

DXMainClient/Domain/Multiplayer/CnCNet/CnCNetAPI.cs
```diff
@@ -0,0 +1,364 @@
...
+        public event Action<bool> Initialized;
+        public event Action<object> AccountUpdated;
+        public event Action<List<VerifiedAccounts>> VerifyAccountsComplete;
+        // Registration URL is separate (web)
+        public static string API_REGISTER_URL = "https://ladder.cncnet.org/auth/register";
+
+    private const string API_AUTH_LOGIN = "auth/login";
+    // Official API: GET /api/v1/user/account (auth: Bearer <JWT>)
+    private const string API_USER_ACCOUNT = "user/account";
+    private const string API_IDENTS_VERIFY = "client/accounts/verify";
```
MahBoiDeveloper: Fix spacing pls

DXMainClient/Domain/Multiplayer/CnCNet/CnCNetAPI.cs
```diff
@@ -0,0 +1,364 @@
...
+        /// </summary>
+        public bool Login(string email, string password)
+        {
+            try
+            {
+                using (ExtendedWebClient client = new ExtendedWebClient(REQUEST_TIMEOUT))
+                {
+                    var request = new NameValueCollection();
+                    request.Add("email", email);
+                    request.Add("password", password);
```
MahBoiDeveloper, line 126: Hash password with salt pls.

MahBoiDeveloper (changes requested):

DXMainClient/Domain/Multiplayer/CnCNet/CnCNetAPI.cs
SadPencil: no need for an HTTPS connection. Only the database needs to hash the password with salt and it is out of the client's concern

MahBoiDeveloper: > no need for an HTTPS connection. Only the database needs to hash the password with salt and it is out of the client's concern

No real password should be transferred through the Internet, so client should implement hashing with salt too IMO.

HTTPS needs to remove MitM attack.

SadPencil: dude, HTTPS is considered safe, without mitm attacks unless the system ca is compromised

MahBoiDeveloper: > dude, HTTPS is considered safe, without mitm attacks unless the system ca is compromised

In Russia we have installed certificates from our local CA center that allows to get access to the banking sites and potentially get us under MittM HTTPS attack. So that is why I suggest to use HTTPS and hashing with salt to protect user's data.

MahBoiDeveloper:
@CnCRAZER @SadPencil @SadPencil @Metadorius I find that FF's client ladder as lobbies with ladder icon in the regular lobby list (that neogrant's implement in old days) is a bit hacky instead of what we have right now in YR package -- separated button for quick matches in main menu. I suggest the future client's implementation of ladder API should follow the same idea -- separate window for searching quick matches in the ladder.

DXMainClient/Domain/Multiplayer/CnCNet/CnCNetAPI.cs
SadPencil: all right...

CnCRAZER:
> Btw there should be button `Terms of Service` or something that redirects to CnCNet's ToS of ladder usage.

The register button takes you to the ladder account creation page. Not the exact same thing but meh

DXMainClient/Domain/Multiplayer/CnCNet/CnCNetAPI.cs
Metadorius: > If you hash on the client side, the hashed password becomes the actual password (with the hashing algorithm being nothing more than a means to convert a user-held mnemonic to the actual password).

> This means that you will be storing the full "plain-text" password (the hash) in the database, and you will have lost all benefit of hashing in the first place.

> If you decide to go this route, you might as well forgo any hashing and simply transmit and store the user's raw password (which, incidentally, I wouldn't particularly recommend).

https://security.stackexchange.com/a/8600

SadPencil: not really. We can have two layers of hashes. The database stores the 1st hash in plaintext; the client generates the 2nd hash from the 1st hash. 1st hash can be calculated using raw password and the salt

Metadorius: why even have the first hash?

Metadorius: This is not what mah_boi proposes though.

> the client generates the 2nd hash from the 1st hash.

doesn't that require transferring the hashed password/salt from server to client? isn't that a big no-no as the client could be malicious and use the obtained correct password hash to calculate locally what the password is?

Metadorius:
> I find that FF's client ladder as lobbies with ladder icon in the regular lobby list (that neogrant's implement in old days) is a bit hacky instead of what we have right now in YR package -- separated button for quick matches in main menu. I suggest the future client's implementation of ladder API should follow the same idea -- separate window for searching quick matches in the ladder.

While I don't have much experience with FF's client ladder implementation, I disagree with the notion that the concept itself is bad.
1. There is *a ton* of ways to implement ranking and/or ladder, including the one using regular lobbies. Age of Empires II uses ranked custom lobbies which only get counted towards the ELO if certain conditions are met, for example, and it works fine.
2. Having to upkeep two separate clients using separate tech and implementations is a big maintenance headache. From my knowledge, only @Sneer-ra2 upkeeps the ladder client. What if he's gone? None of the XNA client devs could replace him properly.

CnCRAZER:
I wasn't expecting so much feedback so quickly lol.

So, currently: user has to go into an active ladder table and be sure they have an active username for the ladder to have the username show up in the login window. Am in talks with ladder api team to make a new _non ladder_ api for generic lobbies.


Will wait for that before I do any more changes I think

DXMainClient/Domain/Multiplayer/CnCNet/CnCNetAPI.cs
SadPencil: no. the 1st hash is calculated from the raw password -- it won't be transmitted from the server to the client. The server stores the 1st only to enable an HTTP-digest-like challenge, removing the need of transferring plaintext password as well as removing the need to store the plaintext password in the database

MahBoiDeveloper:
> Lets say a user decided to not use a cncnet account and use a nick of "xyz", but then a verified account logins with the same nick. Should the unverified account be kicked or told to change their name? Is this even possible? lol

I believe that provide users logging only with email+password would solve issue.

Currently if unregistered user would connect with the same name that already exist in the IRC chat, it would be renamed with `_` in the end, i.e. if I would make a fast reconnect to the Gamesurge IRC, my `mah_boi` nickname would stay in the net and new name would be `mah_boi_`.

If we talk about guests nicknames vs authorized nicknames, I prefer to kick guests or add them postfix like `mah_boi (Guest)` on logging and keep registered users without it.

PR #841, opened 2025-10-03
Rename and move `FileHelper` class to extensions folder with name `FileExtensions`

MahBoiDeveloper:


SadPencil:
This class contains no `extension` method. So why moving it to the `Extensions` namespace?

MahBoiDeveloper:
> This class contains no `extension` method. So why moving it to the `Extensions` namespace?

Because it extends the functionality of the `File` class even if have no extension method like `void DoSomething(this File file) {}`. Previous naming was a bit incorrect.

MahBoiDeveloper:
> We have 4 classes whose name ends up with `Helper` now. Can you further explain why the helper name is incorrect?

`Helper` classes are mostly incorrectly named because they are extensions, managers, inbetween entities.

ClientCore/Extensions/FileExtensions.cs
```diff
@@ -0,0 +1,115 @@
...
+        }
+    }
+
+    /// <summary>
+    /// Predicts text file encoding by its content.
+    /// </summary>
+    /// <param name="filename"></param>
+    /// <param name="minimalConfidence"></param>
+    /// <returns></returns>
+    public static Encoding GetEncoding(string filename, float minimalConfidence = 0.5f)
```
Metadorius: while you're at it, adjust the name of the method to `GetDetectedEncoding` or something like that

MahBoiDeveloper: Done

Metadorius (approved):

PR #847, opened 2025-10-11
Add world flags per ISO-3166-1 to TunnelListBox

### LINKED ISSUE #846: Add country flag before server name in the tunnel list
### Description

Feature should helps to define where server is placed to solve connection issues.

11EJDE11:
Closes #846 

Uses the flags from this project:
https://github.com/lafeber/world-flags-sprite

16px:
https://github.com/lafeber/world-flags-sprite/blob/master/images/flags16.png

CSS contains the offsets:
https://github.com/lafeber/world-flags-sprite/blob/master/stylesheets/flags16-both.css

Place the png in the resources folder named flags16.png

Should the offsets be parsed from a file in the resources folder instead?

---

flags


DXMainClient/DXGUI/Multiplayer/CnCNet/TunnelListBox.cs
```diff
@@ -172,5 +192,137 @@ private void TunnelListBox_SelectedIndexChanged(object sender, EventArgs e)
...
+            offsets["so"] = 3312; offsets["sr"] = 3328; offsets["ss"] = 3936; offsets["st"] = 3344;
+            offsets["sv"] = 3360; offsets["sx"] = 3904; offsets["sy"] = 3376; offsets["sz"] = 3392;
+            offsets["tc"] = 3408; offsets["td"] = 3424; offsets["tg"] = 3440; offsets["th"] = 3456;
+            offsets["tj"] = 3472; offsets["tl"] = 3488; offsets["tm"] = 3504; offsets["tn"] = 3520;
+            offsets["to"] = 3536; offsets["tr"] = 3552; offsets["tt"] = 3568; offsets["tv"] = 3584;
+            offsets["tw"] = 3600; offsets["tz"] = 3616; offsets["ua"] = 3632; offsets["ug"] = 3648;
+            offsets["us"] = 3664; offsets["uy"] = 3680; offsets["uz"] = 3696; offsets["va"] = 3712;
+            offsets["vc"] = 3728; offsets["ve"] = 3744; offsets["vg"] = 3760; offsets["vi"] = 3776;
+            offsets["vn"] = 3792; offsets["vu"] = 3808; offsets["ws"] = 3824; offsets["ye"] = 3840;
+            offsets["yt"] = 1424; offsets["za"] = 3856; offsets["zm"] = 3872; offsets["zw"] = 3888;
```
MahBoiDeveloper: Could we use country code and some offset into image to define country flag instead such bulk definition?

MahBoiDeveloper: Btw old Funky's client has country flags in the separate folder, so we can use them, I suggest.

[Icons.zip](https://github.com/user-attachments/files/22870328/Icons.zip)

DXMainClient/DXGUI/Multiplayer/CnCNet/TunnelListBox.cs
```diff
@@ -29,7 +34,20 @@ public TunnelListBox(WindowManager windowManager, TunnelHandler tunnelHandler) :
             Height = LineHeight * 12 + headerHeight + 3;
             PanelBackgroundDrawMode = PanelBackgroundImageDrawMode.STRETCHED;
             BackgroundTexture = AssetLoader.CreateTexture(new Color(0, 0, 0, 128), 1, 1);
-            AddColumn("Name".L10N("Client:Main:NameHeader"), 230);
+
+            flagsSpriteSheet = AssetLoader.LoadTexture("flags16.png");
```
MahBoiDeveloper: I suggest we could make `flags16.png` embedded and ship it inside client binaries.

DXMainClient/DXGUI/Multiplayer/CnCNet/TunnelListBox.cs
Metadorius: I think you can use the `{ key = value, }` initializer instead

DXMainClient/DXGUI/Multiplayer/CnCNet/TunnelListBox.cs
Metadorius: I don't think customisationlike you suggested in another comment is needed then

DXMainClient/DXGUI/Multiplayer/CnCNet/TunnelListBox.cs
11EJDE11: Unfortunately some are out of order like "bl" for instance - makes it a bit ugly.
Funky's icons look good - I prefer the sprite sheet approach - less file reads and feels leaner. Definitely makes it uglier with the dictionary lookup and more difficult to swap out flags though. Anyone else got any thoughts?

DXMainClient/DXGUI/Multiplayer/CnCNet/TunnelListBox.cs
11EJDE11: Can do. What's the best way to do that?

DXMainClient/DXGUI/Multiplayer/CnCNet/TunnelListBox.cs
11EJDE11: Indeed. b3bff80 covers it, thanks.

DXMainClient/DXGUI/Multiplayer/CnCNet/TunnelListBox.cs
MahBoiDeveloper: > Can do. What's the best way to do that?

Embedded file sounds better because it will be always available by client like as known irc game chat icons (RA, D2K, YR and others)

11EJDE11: Cheers. Sorted.

11EJDE11:
Requesting some feedback: 
Are we happy with the single png containing all flags, or do we want these split out - one png for each country like Funky's client did?

Single PNG is nice - just one file read. But it does make it trickier to edit and means we need that dictionary lookup.

MahBoiDeveloper:
> Requesting some feedback: Are we happy with the single png containing all flags, or do we want these split out - one png for each country like Funky's client did?
> 
> Single PNG is nice - just one file read. But it does make it trickier to edit and means we need that dictionary lookup.

I prefer FF's client method because, if you know the two-letter country code, you can load it by name without needing to edit the dictionary.

```csharp
flagsSpriteSheet = AssetLoader.TextureFromImage(Image.Load("path_sample" + countryTwoLettersCode + ".png" ));
```

However, you have already created a dictionary containing all countries, so I'm OK with the current implementation.

PR #872, opened 2025-11-01
Fix game loading lobbies not being shown because of missing map hash arg

Metadorius:


SadPencil (approved):
Has re-read #825. LGTM

Issue #816, opened 2025-08-11
Suggestion for the game filter menu


Mart1n0vic:
### Description

I would suggest, if technically feasible, adding the following filter options in a later patch into the existing Game Filter Menu, since its very hard to find a game in the lobby WITHOUT crates or stuff (u need join 20 games till u find a competituve game):

- Ra2 Mode
- Yr Mode
- Original (Ladder) Settings ( ! ! ! )
- Blitz
- official games (so no crates, no increased money, no mod maps or other stuff, may to combine with "Yr / Ra2 Ladder Map Pool", ...),

may game with crates or some other options ...,

I could imagine that this would appeal to many players (like me) that looking for their matching games..

https://discord.com/channels/188156159620939776/1331327709464363028

For other may also good find games

- Crates activated
- Rebalance
....

Atm u can´t filter any of this modes, u need join every game in the lobby to find any competitive game.

You can search for player names, map names and the main mode (Battle, Cooperative, ...) but u need write the full word.
If you serach for "Battl" (without "e") u have 0 results which is broken as well i guess.
So u can use this only if the Host change map pool from "Battle" to "Ra2 Ladder" (or "Yr Ladder") but 99% use "Battle" + u need use the search everytime manual and type the full word every time.

In the Filter menu i have activated usually:

Hide locked Games
Hide Passworded Games
Hide incompatible Games
But here would be the top points very useful to add.

Like i said atm i need join every game, wait some seconds till the settings loaded, "Damn crates and 100k" ... leave, join next game, wait,  ...

Thank you in Advance !!!!

PR #874, opened 2025-11-08
Broadcast and filter game options

11EJDE11:
Closes or helps close

- #816 
- https://github.com/CnCNet/cncnet-yr-client-package/issues/611 
- https://github.com/CnCNet/cncnet-yr-client-package/issues/278
  
--------

- Adds game options to the GAME broadcast
- Adds broadcasted game options to the filters in CnCNetLobby
- Adds broadcasted game options to the GameListBox and GameInformationPanel
- Adds setting to disable the game icon from showing in the GameListBox

**New INI settings for GameSessionCheckBox (located in `GameLobbyBase.ini`):**
```ini
BroadcastToLobby=Yes|No
ShowInGameList=Yes|No
ShowInGameListOnRight=Yes|No
ShowInGameInformationPanel=Yes|No
ShowInGameInformationPanelAsIconOnly=Yes|No
ShowIconInGameLobby=Yes|No
ShowInFilters=Yes|No
EnabledIcon=<filename.png>
DisabledIcon=<filename.png>
SortOrder=<integer>
```
New settings for GameSessionDropDown (located in GameLobbyBase.ini):
```ini
BroadcastToLobby=Yes|No
ShowInGameList=Yes|No
ShowInGameListOnRight=Yes|No
ShowInGameInformationPanel=Yes|No
ShowInGameInformationPanelAsIconOnly=Yes|No
ShowIconInGameLobby=Yes|No
ShowInFilters=Yes|No
Icons=<icon1.png>,<icon2.png>,...
SortOrder=<integer>
```

**New setting in `ClientDefinitions.ini`:**
```ini
[Settings]
ShowGameIconInGameList=Yes|No
```

**GameListBox and GameInformationPanel changes:**
image

**Filters menu:**
image


ClientCore/Settings/UserINISettings.cs
```diff
@@ -267,6 +268,34 @@ protected UserINISettings(IniFile iniFile)
...
+        /// Gets the filter value for a dropdown game option.
+        /// Returns -1 for "All", or the selected index.
+        /// </summary>
+        public int GetDropdownFilterValue(string optionName) => SettingsIni.GetIntValue(GAME_OPTION_FILTERS, optionName, -1);
+
+        /// <summary>
+        /// Sets the filter value for a dropdown game option.
+        /// -1 = "All", or the selected index.
+        /// </summary>
+        public void SetDropdownFilterValue(string optionName, int value) => SettingsIni.SetIntValue(GAME_OPTION_FILTERS, optionName, value);
```
Metadorius: 1. This is a bit inconsistent, I think -1 should mean "irrelevant" in both the dropdown and the checkbox.
2. I am not sure yet, but to me `int?` and `bool?` sounds like a more obvious choice here. Curious to hear what you think about that.

DXMainClient/DXGUI/Generic/GameSessionCheckBox.cs
```diff
@@ -57,7 +57,43 @@ public bool AllowScoring
...
+    /// <summary>
+    /// Whether the icon should be shown in the game list.
+    /// </summary>
+    public bool IconShownInGameList { get; private set; }
+
+    /// <summary>
+    /// Whether the icon should be shown on the right side of the game list.
+    /// Only applies if IconShownInGameList is true.
+    /// </summary>
+    public bool IconShownInGameListOnRight { get; private set; }
```
Metadorius: Clever customisation!

DXMainClient/DXGUI/Generic/GameSessionDropDown.cs
```diff
@@ -33,6 +33,26 @@ public GameSessionDropDown(WindowManager windowManager) : base(windowManager) {
...
+
+    /// <summary>
+    /// Whether the icon should be shown in the game filters panel.
+    /// </summary>
+    public bool IconShownInFilters { get; private set; }
+
+    /// <summary>
+    /// The texture name for the icon.
+    /// </summary>
+    public string Icon { get; private set; }
```
Metadorius: I think for this to be more useful, we have to have not one icon per dropdown but actually an icon per option in the dropdown. If the icon is missing -- then no icon would be shown.

This would also streamline the behavior between the checkbox and the dropdown. The icon could be then drawn as ItemTexture or however the option was called.

DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetGameLoadingLobby.cs
```diff
@@ -717,8 +717,9 @@ private void BroadcastGame()
             sb.Append(0); // LoadedGameId
             sb.Append(";");
             sb.Append(ClientConfiguration.Instance.DefaultSkillLevelIndex); // we don't know the original skill level
-            sb.Append(";");
-            sb.Append(string.Empty); // MapHash
+            sb.Append(";"); // Map SHA1
+            sb.Append(";"); // Packed checkbox values (empty for loaded lobbies)
+            sb.Append(";"); // Dropdown indices (empty for loaded lobbies)
```
Metadorius: Is it necessary to separate dropdowns and indices like this? They are essentially both the same, checkbox is an option with 0, 1 states, and dropdown has 0, 1, 2, ... etc. I've  created an interface for game settings specifically to generalize this kind of behavior. What do you think?

DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetLobby.cs
```diff
@@ -460,6 +469,62 @@ private bool HostedGameMatches(GenericHostedGame hg)
...
+                    // -1 = All, otherwise check for exact match
+                    if (filterValue == -1)
+                        continue;
+
+                    int gameSelectedIndex = game.BroadcastedDropdownIndices[i];
+
+                    if (gameSelectedIndex != filterValue)
+                        return false;
+                }
+            }
```
Metadorius: this could be deduplicated I think?

DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetLobby.cs
```diff
@@ -1551,6 +1616,43 @@ private void GameBroadcastChannel_CTCPReceived(object sender, ChannelCTCPEventAr
                 int skillLevel = int.Parse(splitMessage[11]);
                 string mapHash = splitMessage[12];
 
+                bool[] checkboxValues = null;
+                int[] dropdownIndices = null;
+
+                // Games with different versions may have different option counts, so ignore
```
Metadorius, line 1593: not only with different versions, but the LocalGame can be different too, as the client has the ability to monitor other LocalGames game listings

DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetLobby.cs
```diff
@@ -1551,6 +1616,43 @@ private void GameBroadcastChannel_CTCPReceived(object sender, ChannelCTCPEventAr
...
+
+                        int localCheckboxCount = gameLobby.CheckBoxes.Count(cb => cb.BroadcastToLobby);
+                        checkboxValues = new bool[localCheckboxCount];
+                        for (int i = 0; i < localCheckboxCount; i++)
+                        {
+                            int packedIndex = i / 32;
+                            int bitIndex = i % 32;
+                            if (packedIndex < packedCheckboxes.Length)
+                                checkboxValues[i] = (packedCheckboxes[packedIndex] & (1 << bitIndex)) != 0;
+                        }
```
Metadorius: could you explain what is happening here? I see bitpacking which is a clever optimisation, but apart from that I have no idea

DXMainClient/DXGUI/Multiplayer/GameFiltersPanel.cs
```diff
@@ -129,6 +151,213 @@ public override void Initialize()
...
+                };
+
+                int topRowHeight = Math.Max(iconHeight, minLabelRowHeight);
+
+                var dropdown = new XNAClientDropDown(WindowManager)
+                {
+                    Name = $"dd{checkbox.Name}Filter",
+                    ClientRectangle = new Rectangle(columnX, rowY + topRowHeight + itemVerticalSpacing, dropdownWidth, UIDesignConstants.BUTTON_HEIGHT)
+                };
+                dropdown.AddItem("All".L10N("Client:Main:FilterAll"));
```
Metadorius: Our source analyzer suggests something is wrong with the value

DXMainClient/DXGUI/Multiplayer/GameInformationPanel.cs
```diff
@@ -221,6 +261,161 @@ public void SetInfo(GenericHostedGame game)
...
+
+                    if (checkbox.IconShownInGameInfo)
+                    {
+                        string iconName = isChecked ? checkbox.EnabledIcon : checkbox.DisabledIcon;
+                        if (!string.IsNullOrEmpty(iconName))
+                        {
+                            Texture2D icon = AssetLoader.LoadTexture(iconName);
+                            if (icon != null)
+                            {
+                                string text = $"{checkbox.Text}: {(isChecked ? "On".L10N("Client:Main:On") : "Off".L10N("Client:Main:Off"))}";
```
Metadorius, line 308: I think we use Yes/No already in other places? not sure, need to check

DXMainClient/DXGUI/Multiplayer/GameFiltersPanel.cs
```diff
@@ -153,7 +382,29 @@ private void Save()
...
+                if (filterControl.IsCheckbox)
+                {
+                    // For checkboxes: 0 = All (default), 1 = On, 2 = Off
+                    int filterValue = filterControl.DropDown.SelectedIndex;
+                    if (filterValue != 0) // Only save if not "All"
+                        userIniSettings.SetCheckboxFilterValue(filterControl.OptionName, filterValue);
+                }
+                else
+                {
+                    // For dropdowns: -1 = All (default), actual indices are offset by 1
```
Metadorius: Not sure what is the reason for that math here?

DXMainClient/DXGUI/Multiplayer/GameFiltersPanel.cs
Metadorius, line 1: If there are many options you could use the ScrollablePanel I implemented. How does it fare with a lot of options? Say, every option added to broadcast.

Wonder if there's a point in making this all INItializable...

DXMainClient/DXGUI/Generic/GameSessionCheckBox.cs
```diff
@@ -57,7 +57,43 @@ public bool AllowScoring
...
+
+    /// <summary>
+    /// Whether the icon should be shown in the game information panel.
+    /// </summary>
+    public bool IconShownInGameInfo { get; private set; }
+
+    /// <summary>
+    /// Whether the icon should be shown in the game filters panel.
+    /// </summary>
+    public bool IconShownInFilters { get; private set; }
```
Metadorius: Hmm, I know I said previously that it would be good to have it customisable, but I am not sure now. Do you think there would be any case when the modder or game maintainer would want to have an icon in game list but *not* have the icon anywhere else? ATM I am not sure if this makes sense, because the game info/filters icon work as legend, so probably could omit that.

Also I think we could draw the icons near game options in the lobby themselves the same way?

DXMainClient/DXGUI/Multiplayer/GameListBox.cs
```diff
@@ -244,12 +248,64 @@ private void GameListBox_HoveredIndexChanged(object sender, EventArgs e)
...
             int lgTextWidth = hg.IsLoadedGame ? loadedGameTextWidth : 0;
-            int maxTextWidth = Width - hg.Game.Texture.Width - 
+
+            var (leftIcons, rightIcons) = GetGameOptionIcons(hg);
+            int leftIconsWidth = leftIcons.Count > 0 ?
+                (leftIcons.Sum(icon => icon.Width) + (leftIcons.Count * ICON_MARGIN)) : 0;
+            int rightIconsWidth = rightIcons.Count > 0 ?
+                (rightIcons.Sum(icon => icon.Width) + (rightIcons.Count * ICON_MARGIN)) : 0;
+
+            int gameTextureWidth = ClientConfiguration.Instance.ShowGameIconInGameList ? hg.Game.Texture.Width : 0;
```
Metadorius: I do think that if LocalGame is different from the one the game has, the texture should be always on, as we would have no way of easily telling what game is that game room for.

Metadorius (changes requested):
Good work so far mate!

image

image

I don't see if credits are shown in the game info panel, is this intended? I thought if you're able to filter by it - it should also be visible there.

Also maybe the panel should be made two-column and use the same font as the main game info?

Also I think the icons here should be *item* icons, because so far per the idea icons convey the *value*, not *what option* it is. See my review comments on how to make this work.

image

Some of the comments that I left are more of open questions, so curious to hear your thoughts on them.

ClientCore/Settings/UserINISettings.cs
11EJDE11: True - nullable is better. Fixed in [dda5b0c](https://github.com/CnCNet/xna-cncnet-client/pull/874/commits/dda5b0c224148c1604f3c53596f97577f04aad80).

DXMainClient/DXGUI/Generic/GameSessionCheckBox.cs
11EJDE11: Thanks! Perhaps it would be better to have both "ShownInGameListLHS" and "ShownInGameListRHS". What do you think?

Metadorius: Not sure, is there a use for having both?

DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetGameLoadingLobby.cs
11EJDE11: Not needed. That stuff was mostly copied from the existing game options broadcast which had it split. See: https://github.com/CnCNet/xna-cncnet-client/blob/3cf078ba89fd59eb51ff661a7cf246f3696fa995/DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs#L1036

RE the interface - makes sense. I was working on an outdated copy and didn't see those changes. Hopefully [60b23c5](https://github.com/CnCNet/xna-cncnet-client/pull/874/commits/60b23c56f4fe4fab0d3dd81ecc501b26701141ce) sorts it as you imagined.

DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetLobby.cs
11EJDE11: [60b23c5](https://github.com/CnCNet/xna-cncnet-client/pull/874/commits/60b23c56f4fe4fab0d3dd81ecc501b26701141ce) sorts it.

DXMainClient/DXGUI/Multiplayer/CnCNet/CnCNetLobby.cs
11EJDE11: Based off https://github.com/Rampastring/Rampastring.Tools/blob/master/Conversions.cs
https://stackoverflow.com/questions/713057/convert-bool-to-byte

DXMainClient/DXGUI/Multiplayer/GameFiltersPanel.cs
11EJDE11: Good spotting. Sorted  with [fec6165](https://github.com/CnCNet/xna-cncnet-client/pull/874/commits/fec61657efcc9c52ad85c1dfc5006de70f2bbab2).

DXMainClient/DXGUI/Multiplayer/GameFiltersPanel.cs
11EJDE11: Sorted with [60b23c5](https://github.com/CnCNet/xna-cncnet-client/pull/874/commits/60b23c56f4fe4fab0d3dd81ecc501b26701141ce).

DXMainClient/DXGUI/Generic/GameSessionCheckBox.cs
11EJDE11: Yeah, I think we keep the settings but make it apply to the whole thing not just the icon. So "ShownInGameInfo" instead of "IconShownInGameInfo", etc. It's programmed a bit like that already - hence the missing Credits in that screenshot.

I was having hassles with icons in the game lobby. I'll take another look.

Metadorius: I think the filter and the game info should be switched on by the same option, otherwise it would be pretty user unfriendly to have something in filter but not in game info, and vice versa. Same for icons in the list - if the icon is shown in the game list - it must have the same icon (working as legend) in filter and in game info, so the users don't get confused on the meaning.

DXMainClient/DXGUI/Generic/GameSessionDropDown.cs
```diff
@@ -39,19 +39,48 @@ public GameSessionDropDown(WindowManager windowManager) : base(windowManager) {
...
+    /// The texture names for the icons for each dropdown option.
+    /// If specified, this array should match the number of items.
+    /// </summary>
+    public string[] Icons { get; private set; }
+
+    /// <summary>
+    /// Sort order for displaying icons in the GameInformationPanel and GameListBox.
+    /// Lower values appear first.
+    /// </summary>
+    public int SortOrder { get; private set; } = 100;
```
Metadorius: Not sure why this is 100 😅

DXMainClient/DXGUI/Multiplayer/GameInformationPanel.cs
```diff
@@ -221,6 +261,213 @@ public void SetInfo(GenericHostedGame game)
...
+
+            var divider = CreateDivider(0);
+            pnlGameOptions.AddChild(divider);
+
+            int currentY = divider.Bottom + legendPadding;
+
+            // First show icons in a row
+            if (optionIconsOnly.Count > 0)
+            {
+                var sortedIconsOnly = optionIconsOnly.OrderBy(x => x.sortOrder).ToList();
```
SadPencil, line 359: Nitpick. The `List<T>.Sort` method sorts the list in-place and has a higher efficiency than an `.OrderBy.ToList()`.

(It's a pity that PriorityQueue is only provided in .NET 6+ but sorting a list here is totally acceptable)

DXMainClient/DXGUI/Multiplayer/GameInformationPanel.cs
```diff
@@ -221,6 +261,213 @@ public void SetInfo(GenericHostedGame game)
...
+                    var iconPanel = new GameInformationIconOnlyPanel(WindowManager, icon);
+                    iconPanel.ClientRectangle = new Rectangle(iconX, currentY, icon.Width, icon.Height);
+                    pnlGameOptions.AddChild(iconPanel);
+                    iconX += icon.Width + iconSpacing;
+                }
+
+                currentY += maxIconHeight + legendPadding;
+            }
+
+            // Then show icons with text ( two columns)
```
SadPencil: `( two columns)` -> `(two columns)`

DXMainClient/DXGUI/Multiplayer/GameInformationPanel.cs
```diff
@@ -221,6 +261,213 @@ public void SetInfo(GenericHostedGame game)
...
+                    bool isLeftColumn = (itemIndex % 2 == 0);
+                    int xPosition = isLeftColumn ? leftColumnPositionX : rightColumnPositionX;
+                    int yPosition = isLeftColumn ? leftY : rightY;
+
+                    var iconPanel = new GameInformationIconPanel(WindowManager, icon, label, maxIconWidth);
+                    iconPanel.ClientRectangle = new Rectangle(xPosition, yPosition, columnWidth, legendIconHeight);
+                    pnlGameOptions.AddChild(iconPanel);
+
+                    if (isLeftColumn)
+                        leftY += legendIconHeight + 2;
```
SadPencil: You have defined `const int iconSpacing = 4; private const int legendPadding = 5;` but here we got another magic number `2`. You might want to define another constant

ClientCore/Settings/UserINISettings.cs
```diff
@@ -433,6 +469,26 @@ public void ResetGameFilters()
...
+        }
+
+        /// <summary>
+        /// Clears all game option filters.
+        /// </summary>
+        private void ResetGameOptionFilters()
+        {
+            var section = SettingsIni.GetSection(GAME_OPTION_FILTERS);
+            if (section != null)
+                section.RemoveAllKeys();
```
SadPencil: section?.RemoveAllKeys();

DXMainClient/DXGUI/Generic/GameSessionDropDown.cs
SadPencil: So people can easily specify a lower or higher control? Seems like a common way, like the default metric for a network interface.

Mabybe we can make this `100` a constant but `100` is definitely better than a default 0 (unless we want to allow negative numbers, then `0` will be a perfect default value) or `int.Max`

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
```diff
@@ -2087,6 +2087,39 @@ private void BroadcastGame()
...
+
+                // Pad to multiple of 4 bytes
+                while (byteList.Count % 4 != 0)
+                    byteList.Add(0);
+
+                byte[] byteArray = byteList.ToArray();
+
+                // Convert bytes to integers
+                for (int i = 0; i < byteArray.Length / 4; i++)
+                    gameOptionValues.Add(BitConverter.ToInt32(byteArray, i * 4));
```
SadPencil, line 2111: Same as the V3 PR: be careful using BitConverter. You can throw if `BitConverter.IsLittleEndian` does not match our expect, or use `BinaryPrimitives` instead

DXMainClient/DXGUI/Multiplayer/GameInformationIconPanel.cs
```diff
@@ -0,0 +1,54 @@
...
+            base.Draw(gameTime);
+
+            if (icon == null)
+                return;
+
+            var textSize = Renderer.GetTextDimensions(label, FontIndex);
+            int textHeight = (int)textSize.Y;
+
+            int iconY = (textHeight - icon.Height) / 2;
+            if (iconY < 0) iconY = 0;
```
SadPencil: Our style requires a line break after `if (condition)`

SadPencil (commented):
LGTM. Only some minor comments

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
SadPencil: Update: looks like we already have ~20 `BitConverter` usages. Maybe we should change them to `BinaryPrimitives` or globally throw on unexpected `BitConverter.IsLittleEndian` in another PR

DXMainClient/DXGUI/Generic/GameSessionDropDown.cs
Metadorius: I don't see why would we disallow negative values

DXMainClient/DXGUI/Generic/GameSessionDropDown.cs
11EJDE11: Thanks and fixed.

DXMainClient/DXGUI/Multiplayer/GameInformationPanel.cs
11EJDE11: I think on/off is more appropriate as we're displaying the status of an option not answering a yes/no question. We do use yes/no throughout the ini files though.
Happy to change it if you want, I have no strong preference either way.

-------------

SW: On
Crates: Off

-------------

SW: Yes
Crates: No

-------------

DXMainClient/DXGUI/Multiplayer/GameFiltersPanel.cs
11EJDE11: Thanks, it is using the scroll panel now. There is an annoying issue where scrolling with the mousewheel, if the cursor is over a dropdown, the dropdown scrolls too.

DXMainClient/DXGUI/Generic/GameSessionDropDown.cs
11EJDE11: I've changed it to 0 as I'm not blocking negatives so I guess it makes more sense. Setting it to 100 just read better to me.

DXMainClient/DXGUI/Multiplayer/GameInformationPanel.cs
11EJDE11: Thanks SP. Let's leave it as is for now. It's not sorting much and would require a bit more code otherwise.

DXMainClient/DXGUI/Multiplayer/GameInformationPanel.cs
11EJDE11: Fixed thanks.

DXMainClient/DXGUI/Multiplayer/GameInformationPanel.cs
11EJDE11: Thanks, have added a constant.

ClientCore/Settings/UserINISettings.cs
11EJDE11: Fixed, thanks.

DXMainClient/DXGUI/Multiplayer/GameLobby/CnCNetGameLobby.cs
11EJDE11: OK - let's do this in another PR.

DXMainClient/DXGUI/Multiplayer/GameInformationIconPanel.cs
11EJDE11: Sorted, thanks.

PR #880, opened 2025-11-14
Remove enforcement of all players in game loading lobby

Metadorius:


SadPencil (dismissed):

SadPencil:
Ah wait,

```cs
    public class CnCNetGameLoadingLobby : GameLoadingLobbyBase
    {
        protected override void HostStartGame()
        {
            AddNotice("Contacting tunnel server...".L10N("Client:Main:ConnectingTunnel"));
            List<int> playerPorts = tunnelHandler.CurrentTunnel.GetPlayerPortInfo(SGPlayers.Count);

            if (playerPorts.Count < Players.Count)
            {
                ShowTunnelSelectionWindow(("An error occured while contacting the CnCNet tunnel server.\nTry picking a different tunnel server:").L10N("Client:Main:ConnectTunnelError1"));
                AddNotice(("An error occured while contacting the specified CnCNet " +
                    "tunnel server. Please try using a different tunnel server").L10N("Client:Main:ConnectTunnelError2") + " ", Color.Yellow);
                return;
            }
```

For example, here we get a list using `SGPlayers.Count` and then determining if it is smaller than `Players.Count`

11EJDE11:
As SadPencil said, the check there reads wrong - Players.Count would be better (generate a port for each player in the lobby, not each player in the saved game). But if you do that, it would require couple other changes:

-Tunnel won't report back ports if count is 1 (can happen if 2xreal player 1x AI in original game, then 1x Real and 1x AI in loaded), so you'd need to generate a port yourself.
-Update the loop in LoadGame/GameLoadingLobbybase to `for (int i = 1; i <= Players.Count; i++)` so it still yeets the player section if there's only 1 real player in the loaded game.
-Maybe other things too


One other thing I noticed is if you have for example a game with 2 real players and 1 AI, if the other real player isn't present during the load, the spawner takes the Skirmish path rather than Internet path. I notice in the Internet path, you do Reconcile Players which turns missing players over to the AI (I assume this means the AI will build as the player?) and that doesn't happen here - the game loads fine, but the other player doesn't build or anything. Also the game speed is set to skirmish's default of 30fps.

PR #883, opened 2025-11-17
Add a web label

11EJDE11:
This PR adds a label that can query JSON data from a datasource. We can use this for the online player count, for ladder ranking, or news. INI datasources are also supported.

Datasources are specified in `ClientDefinitions.ini`

```
[DataSources]
DataSourceID=URL,RefreshIntervalSeconds,TimeoutSeconds,Format ;example: TopPlayers=https://ladder.cncnet.org/api/v1/qm/ladder/rankings,300,10,json|ini
```

Labels then point to a datasource and can query the JSON.

```
[ExtraControls]
1=lblTopPlayer1:XNAClientJSONLabel

[lblTopPlayer1]
$Type=XNAClientJSONLabel
DataSourceID=TopPlayers
Template=1. {YR[0].player_name} - {YR[0].points:N0} pts
LoadingText=Loading...
FallbackText=---
FontIndex=1
RemapColor=230,230,230
Location=360,166
```

See `INISytem.md` for more info.

Excuse the crappy screenshot.
image


ClientGUI/XNAClientJSONLabel.cs
```diff
@@ -0,0 +1,191 @@
...
+                        results.Add(token.ToString());
+
+            return results.Count > 0 ? results : null;
+        }
+        catch (Exception ex)
+        {
+            Logger.Log($"JSONPath error: {ex.Message}");
+            return null;
+        }
+    }
```
Metadorius: I think this should happen way earlier at manager stage, and the control should only receive a parsed json and do template substitution, otherwise we're duplicating work.

Also we already use System.Text.Json, I see that it has no support for JSON Path for some reason, maybe this extension to System.Text.Json would do better? https://www.nuget.org/packages/JsonPath.Net Just so we don't introduce two similar libraries.

ClientGUI/XNAClientJSONLabel.cs
```diff
@@ -0,0 +1,191 @@
...
+
+namespace ClientGUI;
+
+public class XNAClientJSONLabel : XNALabel
+{
+    public string DataSourceID { get; set; }
+    public string Template { get; set; }
+    public string LoadingText { get; set; }
+    public int MaxResults { get; set; } = 0;
+    public string FallbackText { get; set; } = "N/A";
```
Metadorius: missing .L10N call

ClientCore/JSONDataSourceManager.cs
```diff
@@ -0,0 +1,339 @@
...
+            }
+
+            _lastJson = data;
+            NotifySubscribers(data, false);
+        }
+
+        private string ConvertIniToJson(string iniContent)
+        {
+            var iniFile = new IniFile();
+            using (var reader = new StringReader(iniContent))
```
Metadorius: There is a constructor in IniFile that takes a stream, no need to parse it by hand, just take a string and make a stream out of it 😄

ClientCore/JSONDataSourceManager.cs
```diff
@@ -0,0 +1,339 @@
...
+                        string key = line.Substring(0, equalIndex).Trim();
+                        string value = line.Substring(equalIndex + 1).Trim();
+                        iniFile.SetStringValue(currentSection, key, value);
+                    }
+                }
+            }
+
+            var sb = new StringBuilder();
+            sb.Append("{");
+            bool firstSection = true;
```
Metadorius: The aforementioned bringing of JSON parsing earlier than that will allow you to construct the programmatic representation via a few foreaches, iterating sections and then keys to set values and greatly simplifying the code 🙂

ClientCore/JSONDataSourceManager.cs
```diff
@@ -0,0 +1,339 @@
...
+        const string sectionName = "DataSources";
+        var section = iniFile.GetSection(sectionName);
+
+        foreach (var key in section.Keys)
+        {
+            string value = section.GetStringValue(key.Key, string.Empty);
+            if (string.IsNullOrEmpty(value))
+                continue;
+
+            // format: URL,RefreshIntervalSeconds,TimeoutSeconds,Format
```
Metadorius: URL could have commas in it, I think instead should do something like

```ini
[DataSourceName]
URL=https://...
RefreshIntervalSeconds=
```

Also I think individual data source parsing should be in the DataSource itself?

ClientCore/JSONDataSourceManager.cs
```diff
@@ -0,0 +1,339 @@
...
+namespace ClientCore;
+
+/// <summary>
+/// Manages shared JSON data sources for XNAClientJSONLabel.
+/// Data sources are defined in INI files under [DataSources] section.
+/// </summary>
+public class JSONDataSourceManager
+{
+    private static JSONDataSourceManager _instance;
+    public static JSONDataSourceManager Instance => _instance ??= new JSONDataSourceManager();
```
Metadorius: We have a DI container in use, I think you could use that? The codebase haven't fully migrated though, so tell me if you encounter any issues with that

ClientCore/JSONDataSourceManager.cs
```diff
@@ -0,0 +1,339 @@
...
+            dataSource.Unsubscribe(callback);
+    }
+
+    private class DataSource
+    {
+        private readonly string _id;
+        private readonly string _url;
+        private readonly int _refreshIntervalSeconds;
+        private readonly int _timeoutSeconds;
+        private readonly string _format;
```
Metadorius: I think this `_format` should be either an enum and/or maybe even a tiny simple interface/class with one method that converts input value (stream?) and gives JSON for arbitrary data representation.

ClientCore/JSONDataSourceManager.cs
```diff
@@ -0,0 +1,339 @@
...
+
+        if (_dataSources.TryGetValue(dataSourceId, out var dataSource))
+            dataSource.Unsubscribe(callback);
+    }
+
+    private class DataSource
+    {
+        private readonly string _id;
+        private readonly string _url;
+        private readonly int _refreshIntervalSeconds;
```
Metadorius, line 147: does it run even in background or when controls using it is not active?

although I realise now that it will probably complicate the code a lot, so I don't think there's a big need in doing it exactly. but if there is *a lot* of players that don't go online -- maybe autorefresh isn't something we may want to do? it will strain our servers quite a bit, maybe we should update just when the control is shown for the first time? (or sth in this direction)

ClientGUI/XNAClientJSONLabel.cs
```diff
@@ -0,0 +1,191 @@
...
+using Rampastring.Tools;
+using System;
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
+using ClientCore;
+using ClientCore.Extensions;
+
+namespace ClientGUI;
+
+public class XNAClientJSONLabel : XNALabel
```
Metadorius: Perhaps let's name it `XNAClientWebLabel`, because it's not just JSON anymore?

Metadorius: also maybe it could be good to have it as a descendant of XNALinkLabel? in case the users want to see details

Metadorius (changes requested):
Nice! Gave you some feedback.

Generally though while this approach definitely has it's place, I am not sure if we should do an explicit separate data source manager? My thoughts were more about something like an implicit manager class; different labels could have content URL endpoints specified right in place, and the response will be shared between them by that class (say, if the requests came to the manager in span of half a minute, then give the same data to every subscriber; if something updates the URL -- then all of the consumers of that URL will receive that). 

Not fully set on the above, what are your thoughts?

cc @SadPencil @MahBoiDeveloper @Starkku @Rampastring

ClientGUI/XNAClientJSONLabel.cs
SadPencil: "N/A" is short for not applicable, instead of "not available". So I don't think we should use "N/A" here

ClientCore/JSONDataSourceManager.cs
SadPencil: I agree with an interface. So we can decouple some string parsing logic from the UI logic. [My recent editing experience on LAN lobby](https://github.com/CnCNet/xna-cncnet-client/pull/826#issuecomment-3758744082) shows that it might waste a huge amount of time to maintain codes if different kinds of logics are not decoupled

ClientGUI/XNAClientJSONLabel.cs
Metadorius: it's used interchangeably as either of those, so I think it's fine?

pzhlkj6612: The example in "Docs/INISystem.md" uses `---`. I think using `---` as the default value is better because it exactly means "nothing" with placeholders. Then we can change the example value to `N/A`.

11EJDE11: Changed to "---" [65229d1](https://github.com/CnCNet/xna-cncnet-client/pull/883/commits/65229d1294d1a9ddc4a1cfd53be6ab2e3fe8583f)

ClientGUI/XNAClientJSONLabel.cs
11EJDE11: Renamed [65229d1](https://github.com/CnCNet/xna-cncnet-client/pull/883/commits/65229d1294d1a9ddc4a1cfd53be6ab2e3fe8583f)
Will look into the LinkLabel.

ClientCore/JSONDataSourceManager.cs
11EJDE11: Looks good, works well. [a19f21c](https://github.com/CnCNet/xna-cncnet-client/pull/883/commits/a19f21c2cc42ff42d123962573f16137c79dedb6)

ClientCore/JSONDataSourceManager.cs
11EJDE11: [65229d1](https://github.com/CnCNet/xna-cncnet-client/pull/883/commits/65229d1294d1a9ddc4a1cfd53be6ab2e3fe8583f)

ClientCore/JSONDataSourceManager.cs
11EJDE11: [65229d1](https://github.com/CnCNet/xna-cncnet-client/pull/883/commits/65229d1294d1a9ddc4a1cfd53be6ab2e3fe8583f)

ClientCore/JSONDataSourceManager.cs
11EJDE11: [65229d1](https://github.com/CnCNet/xna-cncnet-client/pull/883/commits/65229d1294d1a9ddc4a1cfd53be6ab2e3fe8583f)

ClientGUI/XNAClientJSONLabel.cs
11EJDE11: What do you mean by "in case the users want to see details"? @Metadorius

ClientGUI/XNAClientJSONLabel.cs
11EJDE11: JSON is now parsed to JToken in the manager and the label only does template substitution, so the duplication is sorted. Regarding JsonPath.NET, I couldn't get this working in an acceptable manner.

ClientCore/JSONDataSourceManager.cs
11EJDE11: Done in the latest commit - IDataSourceFormat.

ClientGUI/XNAClientJSONLabel.cs
Metadorius: I meant that, for example, there's ladder info like you've shown, and clicking on that info could open the full ladder info with details in browser, which is the details in question.

11EJDE11: Makes sense, sorted thanks.

PR #884, opened 2025-11-20
Try to avoid GitVersion race condition in writing log file

SadPencil:
PR reverted. See https://github.com/CnCNet/xna-cncnet-client/pull/885 instead


SadPencil:
PR reverted

PR #889, opened 2025-11-24
Switch to Xml Solution File (.slnx)

SadPencil:
This PR:

1. Use `.slnx` format for a cleaner and easy-to-maintain VS solution. This allows us to specify x86 platform only for XNA build (no matter "Any CPU" or "x86" platform is selected), enforcing AnyCPU platform for non-XNA builds even if the platform was accidently change to x86 in Visual Studio.

This change requires at least Visual Studio 2022 17.14 or Visual Studio 2026. We already using C# 12 for a long time and adopting C# 14 usage (which requires at least Visual Studio 2026) just now, so I think this slnx change should not bring problems.

See also: https://devblogs.microsoft.com/visualstudio/new-simpler-solution-file-format/

2. Make `ClientCore` and `ClientUpdater` unaware of GUI stuffs. Since `ClientCore` was decoupled from GUI stuffs recently in 4946616cd2550201dfc565d27e6ec8692cd82758 and 1283a39b1ee38a8924ddc8a07dab726055eb73df, finally we can move `ClientCore` outside of the `net8.0-windows` `<UseWindowsForms>` `<Platform Condition="$(Engine.Contains(XNA))">x86</Platform>` stuffs.


Metadorius (approved):
did a build on Rider, seems good

PR #893, opened 2025-12-07
Greatly reduce time cost in generating and loading custom map cache

SadPencil:
This PR upgrades the custom map cache feature, bumping version to v2.
- Optionally invalidate the map file only if the file size or last write time gets changed. The old recalculate SHA method is not used anymore since it is time consuming and also results in metadata mismatch.
- Improve the behavior with map name translation in cached map.

图片
图片

### DIFF
--- a/DXMainClient/Domain/Multiplayer/CustomMapCache.cs
+++ b/DXMainClient/Domain/Multiplayer/CustomMapCache.cs
@@ -47,17 +47,10 @@ namespace DTAClient.Domain.Multiplayer
                 }
             }
 
-            public void RefreshIfOutdated()
+            public bool IsOutdated()
             {
                 Item refreshedItem = new(Map);
-                bool recalculateSHA = refreshedItem.FileSize != FileSize || refreshedItem.LastWriteTimeUtc != LastWriteTimeUtc;
-                if (recalculateSHA)
-                {
-                    FileSize = refreshedItem.FileSize;
-                    LastWriteTimeUtc = refreshedItem.LastWriteTimeUtc;
-                }
-
-                Map.AfterDeserialize(recalculateSHA);
+                return refreshedItem.FileSize != FileSize || refreshedItem.LastWriteTimeUtc != LastWriteTimeUtc;
             }
         }
     }
--- a/DXMainClient/Domain/Multiplayer/CustomMapCache.cs
+++ b/DXMainClient/Domain/Multiplayer/CustomMapCache.cs
@@ -16,16 +16,16 @@ namespace DTAClient.Domain.Multiplayer
         [JsonPropertyName("maps")]
         public required ConcurrentDictionary<string, Item> Items { get; set; }
 
-        public class Item
+        public record Item
         {
             [JsonInclude]
             public required Map Map { get; init; }
 
             [JsonInclude]
-            public required long FileSize { get; init; }
+            public long FileSize { get; private set; }
 
             [JsonInclude]
-            public required DateTime LastWriteTimeUtc { get; init; }
+            public DateTime LastWriteTimeUtc { get; private set; }
 
             public Item() : base() { }
 
@@ -49,9 +49,15 @@ namespace DTAClient.Domain.Multiplayer
 
             public void RefreshIfOutdated()
             {
-                FileInfo fileInfo = new(Map.CompleteFilePath);
-                bool recalculateSHA = fileInfo.Exists && (fileInfo.Length != FileSize || fileInfo.LastWriteTimeUtc != LastWriteTimeUtc);
-                Map.AfterDeserialize(recalculateSHA);
+                Item refreshedItem = new(Map);
+                bool recalculateSHA = refreshedItem.FileSize != FileSize || refreshedItem.LastWriteTimeUtc != LastWriteTimeUtc;
+                if (recalculateSHA)
+                {
+                    FileSize = refreshedItem.FileSize;
+                    LastWriteTimeUtc = refreshedItem.LastWriteTimeUtc;
+
+                    Map.AfterDeserialize(recalculateSHA);
+                }
             }
         }
     }
--- a/DXMainClient/Domain/Multiplayer/CustomMapCache.cs
+++ b/DXMainClient/Domain/Multiplayer/CustomMapCache.cs
@@ -44,9 +44,6 @@ namespace DTAClient.Domain.Multiplayer
 
             public void RefreshIfOutdated()
             {
-                if (!File.Exists(Map.CompleteFilePath))
-                    return;
-
                 FileInfo fileInfo = new(Map.CompleteFilePath);
                 bool recalculateSHA = fileInfo.Exists && (fileInfo.Length != FileSize || fileInfo.LastWriteTimeUtc != LastWriteTimeUtc);
                 Map.AfterDeserialize(recalculateSHA);
--- a/DXMainClient/Domain/Multiplayer/CustomMapCache.cs
+++ b/DXMainClient/Domain/Multiplayer/CustomMapCache.cs
@@ -1,4 +1,7 @@
-﻿using System.Collections.Concurrent;
+﻿using System;
+using System.Collections.Concurrent;
+using System.Diagnostics.CodeAnalysis;
+using System.IO;
 using System.Text.Json.Serialization;
 
 namespace DTAClient.Domain.Multiplayer
@@ -7,10 +10,47 @@ namespace DTAClient.Domain.Multiplayer
     {
         [JsonInclude]
         [JsonPropertyName("version")]
-        public int Version { get; set; }
+        public required int Version { get; set; }
 
         [JsonInclude]
         [JsonPropertyName("maps")]
-        public ConcurrentDictionary<string, Map> Maps { get; set; }
+        public required ConcurrentDictionary<string, Item> Items { get; set; }
+
+        public class Item
+        {
+            [JsonInclude]
+            public required Map Map { get; init; }
+
+            [JsonInclude]
+            public required long FileSize { get; init; }
+
+            [JsonInclude]
+            public required DateTime LastWriteTimeUtc { get; init; }
+
+            public Item() : base() { }
+
+            [SetsRequiredMembers]
+            public Item(Map map)
+            {
+                Map = map;
+
+                FileInfo fileInfo = new(Map.CompleteFilePath);
+                if (fileInfo.Exists)
+                {
+                    FileSize = fileInfo.Length;
+                    LastWriteTimeUtc = fileInfo.LastWriteTimeUtc;
+                }
+            }
+
+            public void RefreshIfOutdated()
+            {
+                if (!File.Exists(Map.CompleteFilePath))
+                    return;
+
+                FileInfo fileInfo = new(Map.CompleteFilePath);
+                bool recalculateSHA = fileInfo.Exists && (fileInfo.Length != FileSize || fileInfo.LastWriteTimeUtc != LastWriteTimeUtc);
+                Map.AfterDeserialize(recalculateSHA);
+            }
+        }
     }
 }
--- a/DXMainClient/Domain/Multiplayer/CustomMapCache.cs
+++ b/DXMainClient/Domain/Multiplayer/CustomMapCache.cs
@@ -40,6 +40,11 @@ namespace DTAClient.Domain.Multiplayer
                     FileSize = fileInfo.Length;
                     LastWriteTimeUtc = fileInfo.LastWriteTimeUtc;
                 }
+                else
+                {
+                    FileSize = 0;
+                    LastWriteTimeUtc = DateTime.MinValue;
+                }
             }
 
             public void RefreshIfOutdated()
--- a/DXMainClient/Domain/Multiplayer/CustomMapCache.cs
+++ b/DXMainClient/Domain/Multiplayer/CustomMapCache.cs
@@ -55,9 +55,9 @@ namespace DTAClient.Domain.Multiplayer
                 {
                     FileSize = refreshedItem.FileSize;
                     LastWriteTimeUtc = refreshedItem.LastWriteTimeUtc;
-
-                    Map.AfterDeserialize(recalculateSHA);
                 }
+
+                Map.AfterDeserialize(recalculateSHA);
             }
         }
     }
--- a/DXMainClient/Domain/Multiplayer/Map.cs
+++ b/DXMainClient/Domain/Multiplayer/Map.cs
@@ -65,7 +65,7 @@ namespace DTAClient.Domain.Multiplayer
         /// The name of the map.
         /// </summary>
         [JsonIgnore]
-        public string Name => string.IsNullOrEmpty(UntranslatedName) || string.IsNullOrEmpty(BaseFilePath)
+        public string Name => !Official || string.IsNullOrEmpty(UntranslatedName) || string.IsNullOrEmpty(BaseFilePath)
             ? UntranslatedName
             : UntranslatedName.L10N($"INI:Maps:{BaseFilePath}:Description");
 
--- a/DXMainClient/Domain/Multiplayer/Map.cs
+++ b/DXMainClient/Domain/Multiplayer/Map.cs
@@ -637,7 +637,12 @@ namespace DTAClient.Domain.Multiplayer
         {
             Name = UntranslatedName;
             if (recalculateSHA)
+            {
+                // Instead of doing so, we should just remove the Map object from cache when the map file changes.
+                // Otherwise, the metadata can be out of date.
+                Debug.Assert(false, "The map SHA1 should not be recalculated after deserialization. Remove the Map object from cache when the map file changes instead.");
                 CalculateSHA();
+            }
         }
 
         private void ParseForcedOptions(IniFile iniFile, string forcedOptionsSection)
--- a/DXMainClient/Domain/Multiplayer/Map.cs
+++ b/DXMainClient/Domain/Multiplayer/Map.cs
@@ -64,25 +64,13 @@ namespace DTAClient.Domain.Multiplayer
         /// <summary>
         /// The name of the map.
         /// </summary>
-        [JsonIgnore]
-        public string Name => !Official || string.IsNullOrEmpty(UntranslatedName) || string.IsNullOrEmpty(BaseFilePath)
-            ? UntranslatedName
-            : UntranslatedName.L10N($"INI:Maps:{BaseFilePath}:Description");
+        [JsonInclude]
+        public string Name { get; private set; }
 
         /// <summary>
         /// The original untranslated name of the map.
         /// </summary>
-        [JsonInclude]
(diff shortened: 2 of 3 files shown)


DXMainClient/Domain/Multiplayer/Map.cs
```diff
@@ -641,10 +642,10 @@ public bool SetInfoFromCustomMap()
         }
 
         // Ran after the map has been loaded from cache if it is a custom map.
-        public void AfterDeserialize()
+        public void AfterDeserialize(bool recalculateSHA = true)
         {
```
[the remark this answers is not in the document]
SadPencil: Good catch!

DXMainClient/Domain/Multiplayer/CustomMapCache.cs
```diff
@@ -7,10 +10,47 @@ public class CustomMapCache
...
+                }
+            }
+
+            public void RefreshIfOutdated()
+            {
+                if (!File.Exists(Map.CompleteFilePath))
+                    return;
+
+                FileInfo fileInfo = new(Map.CompleteFilePath);
+                bool recalculateSHA = fileInfo.Exists && (fileInfo.Length != FileSize || fileInfo.LastWriteTimeUtc != LastWriteTimeUtc);
```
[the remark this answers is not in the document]
SadPencil: 0ff789b937d5bcba82867dffed39420d3d78a472

DXMainClient/Domain/Multiplayer/CustomMapCache.cs
```diff
@@ -7,10 +10,47 @@ public class CustomMapCache
...
+        public class Item
+        {
+            [JsonInclude]
+            public required Map Map { get; init; }
+
+            [JsonInclude]
+            public required long FileSize { get; init; }
+
+            [JsonInclude]
+            public required DateTime LastWriteTimeUtc { get; init; }
```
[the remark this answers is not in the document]
SadPencil: Good catch! Addressed in fad1c52be4550b4735e6ed7698d724c7bbbd1113

DXMainClient/Domain/Multiplayer/CustomMapCache.cs
```diff
@@ -7,10 +10,55 @@ public class CustomMapCache
...
+            public void RefreshIfOutdated()
+            {
+                Item refreshedItem = new(Map);
+                bool recalculateSHA = refreshedItem.FileSize != FileSize || refreshedItem.LastWriteTimeUtc != LastWriteTimeUtc;
+                if (recalculateSHA)
+                {
+                    FileSize = refreshedItem.FileSize;
+                    LastWriteTimeUtc = refreshedItem.LastWriteTimeUtc;
+
+                    Map.AfterDeserialize(recalculateSHA);
```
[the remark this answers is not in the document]
SadPencil: 0cec474477c65dedbcb321d739e6be8ca78a3f5d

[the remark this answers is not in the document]
SadPencil: Good catch. The 2nd issue is addressed in d5c41f3a98762a0f2d79fe34cfcdedd735b292f0

DXMainClient/Domain/Multiplayer/CustomMapCache.cs
```diff
@@ -7,10 +10,48 @@ public class CustomMapCache
...
+                {
+                    FileSize = 0;
+                    LastWriteTimeUtc = DateTime.MinValue;
+                }
+            }
+
+            public bool IsOutdated()
+            {
+                Item refreshedItem = new(Map);
+                return refreshedItem.FileSize != FileSize || refreshedItem.LastWriteTimeUtc != LastWriteTimeUtc;
```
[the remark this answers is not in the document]
SadPencil: ?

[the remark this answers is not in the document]
SadPencil: Wrong. Note the `!customMapCache.Items.ContainsKey(map.SHA1)` if condition

11EJDE11:
LGTM. If you can exclude Preview and PreviewPack in GetCustomMapIniFile, it will speed up the creation of the cache file. 2 minutes vs 25 seconds for my test case.

DXMainClient/Domain/Multiplayer/Map.cs
```diff
@@ -632,10 +633,16 @@ public bool SetInfoFromCustomMap()
...
-            CalculateSHA();
-            UntranslatedName = Name;
+            Name = UntranslatedName;
```
MahBoiDeveloper: Are you sure that previous `UntranslatedName = Name` is a bug?

SadPencil: 95% sure

SadPencil:
> LGTM. If you can exclude Preview and PreviewPack in GetCustomMapIniFile, it will speed up the creation of the cache file. 2 minutes vs 25 seconds for my test case.

Good idea! Can you push your code directly to this PR or do you prefer me implementing it later?

DXMainClient/Domain/Multiplayer/Map.cs
Metadorius: Could you explain?

SadPencil: > Could you explain?

`Name` is a localized map name. If it get cached, the next time the client reads the cache, `UntranslatedName = Name` will fill the `UntranslatedName` with the localized map name, which is wrong.

SadPencil:
> LGTM. If you can exclude Preview and PreviewPack in GetCustomMapIniFile, it will speed up the creation of the cache file. 2 minutes vs 25 seconds for my test case.

Impressive.

图片

11EJDE11 (approved):

SadPencil:
Will merge this PR after ~24 hours if there are no more comments

DXMainClient/Domain/Multiplayer/Map.cs
Metadorius: Oh you should've mentioned that you made Name not serialized and UntranslatedName serialized. I don't understand the reason behind this though (and map caching is done for custom maps wnyway which don't have localization anyways).

SadPencil: > Oh you should've mentioned that you made Name not serialized and UntranslatedName serialized. I don't understand the reason behind this though (and map caching is done for custom maps wnyway which don't have localization anyways).

I am not a fan of assuming a lot of rules, so would prefer lift such a `custom maps don't have localization anyways` restriction even if no one localize it

Metadorius: This isn't an assumption. Localizing custom maps in current system makes no sense. Our translation file is centralized and no way custom mappers will tell users to change it somehow, and I don't think custom mappers care about translations either way, so it doesn't make much sense in the current system.

SadPencil: > This isn't an assumption. Localizing custom maps in current system makes no sense. Our translation file is centralized and no way custom mappers will tell users to change it somehow, and I don't think custom mappers care about translations either way, so it doesn't make much sense in the current system.

This is a statement that can only be derived from the existing rules, requires some times and brains to get it -- while it's good to lower the 'thinking overhead' when writing codes for this project. So I am still not a fan of it.

Besides, there are no drawbacks for the `made Name not serialized and UntranslatedName serialized` change -- before we introduce localization, `Name` just refers to `UntranslatedName`, and therefore we should cache this `UntranslatedName` field to maintain the old behavior.

SadPencil: cb95c5db to completely remove `Name = UntranslatedName` or vice versa thing

DXMainClient/Domain/Multiplayer/Map.cs
```diff
@@ -64,12 +64,15 @@ public Map(string baseFilePath, bool isCustomMap)
         /// <summary>
         /// The name of the map.
         /// </summary>
-        [JsonInclude]
-        public string Name { get; private set; }
+        [JsonIgnore]
+        public string Name => string.IsNullOrEmpty(UntranslatedName) || string.IsNullOrEmpty(BaseFilePath)
+            ? UntranslatedName
+            : UntranslatedName.L10N($"INI:Maps:{BaseFilePath}:Description");
```
Metadorius: > This is a statement that can only be derived from the existing rules, requires some times and brains to get it -- while it's good to lower the 'thinking overhead' when writing codes for this project. So I am still not a fan of it.

IMO such things should be solved by leaving comments if not possible to easily figure that out without thinking.

I mean, this new change isn't obvious either. It is not obvious that missing `BaseFilePath` signifies that it is a custom map or not, and it is far from clear why in one case it's translated and in other it is not.

SadPencil: > > This is a statement that can only be derived from the existing rules, requires some times and brains to get it -- while it's good to lower the 'thinking overhead' when writing codes for this project. So I am still not a fan of it.
> 
> IMO such things should be solved by leaving comments if not possible to easily figure that out without thinking.
> 
> I mean, this new change isn't obvious either. It is not obvious that missing `BaseFilePath` signifies that it is a custom map or not, and it is far from clear why in one case it's translated and in other it is not.

The new change now completely lifted the restriction of translating custom maps. People can translate these maps if they want to, although no one would do so. It lifted this assumption/requirement/statement/whatever name so we don't have to leaving comments for this

SadPencil: Alternatively we can add `isCustomMap` to the if condition here, if we deliberately want to avoid this translation thing. Either way, I don't know why we should bring such a difference against whether it is a custom map when speaking of map caching

Metadorius: > Alternatively we can add `isCustomMap` to the if condition here, if we deliberately want to avoid this translation thing. Either way, I don't know why we should bring such a difference against whether it is a custom map when speaking of map caching

the map caching ONLY makes sense for custom maps, info about non-custom maps is listed in MPMaps.ini

SadPencil: > > Alternatively we can add `isCustomMap` to the if condition here, if we deliberately want to avoid this translation thing. Either way, I don't know why we should bring such a difference against whether it is a custom map when speaking of map caching
> 
> the map caching ONLY makes sense for custom maps, info about non-custom maps is listed in MPMaps.ini

this is not what I mean. For example we use UDP to carry DNS queries, but when implementing UDP, we don't think anything special of DNS things. Same logic applied to the caching here -- caching is a low-level thing, so why should it care if it's a custom map? `ONLY` is not a good answer here.

Metadorius: DNS and UDP are hardly comparable to our case. DNS and UDP are decoupled and abstracted completely from each other, and for a good reason: they are common, reusable protocols, and the more reusable the protocol is -- the higher the decoupling must be for it to be truly reusable.

Here we don't have such an issue. Our code is not intended to be reusable in this case because there's little benefit, hence why there's no point in decoupling it and abstracting away. Abstraction has it's costs, one of such costs is not being able to do optimizations depending on the interaction of the systems. Not using caching for maps that don't need caching is an example of such optimization.

SadPencil: > DNS and UDP are hardly comparable to our case. DNS and UDP are decoupled and abstracted completely from each other, and for a good reason: they are common, reusable protocols, and the more reusable the protocol is -- the higher the decoupling must be for it to be truly reusable.
> 
> Here we don't have such an issue. Our code is not intended to be reusable in this case because there's little benefit, hence why there's no point in decoupling it and abstracting away. Abstraction has it's costs, one of such costs is not being able to do optimizations depending on the interaction of the systems. Not using caching for maps that don't need caching is an example of such optimization.

So are there any cost in changing the translation name thing in this PR? Decoupling the caching system from being aware if the map is custom map or not does not prevent us from doing "Not using caching for official maps"

Metadorius: > So are there any cost in changing the translation name thing in this PR?

There is a cost:
1. Every change leads to unknown amount of degradations, even if we try to minimise it, because software engineering is a complex feat. When the change is warranted (there was a bug that needs to be fixed, or there is an objective need for improvement), we have to pay this cost, otherwise we won't be able to fulfill the objective.
2. It adds to the history of changes when going through it for reasons of debugging or whatever, adding cognitive load on the one who is doing that.

SadPencil: I'll try to say why this change is necessary:

Before this PR, the caching logic has an inconsistency issue (UntranslatedName = Name, https://github.com/CnCNet/xna-cncnet-client/pull/893#discussion_r2598132781), only if the map is cached and translated. Since the caching only works on custom map, and most people won't translate the custom map, this inconsistency issue never been triggered.

However, this issue is there, wasting some of the brains whenever one is maintaining either the map or the cache.

This PR fixes this inconsistency issue.

That's all. I don't think leaving this issue here has any advantages. The weaknesses of leaving an obvious but untriggerable bug is greater than the "benefits" of simplifying the code history.

Metadorius: > only if the map is cached and translated

This is impossible. Check the original code in more detail, I specifically didn't code translation for the custom map loading path.

SadPencil: > > only if the map is cached and translated
> 
> 
> 
> This is impossible. Check the original code in more detail, I specifically didn't code translation for the custom map loading path.

I already said untriggerable

if (false)
  a_confusing_statement;

This still confuses people, especially the if condition is so implicit and in a place very far away from a_confusing_statement;

Metadorius: So we go back again to "should have left a comment explaining the relation to reduce the overhead".

SadPencil: There will be numerous of comments then, shown below. I don't think it is better than just improve it.

1. A comment right after marking Name JsonInclude, as Name is a translated string so it is strange it is cached

2. A comment in marking UntranslatedName JsonExclude, as it is a raw data but being Excluded

3. A comment in assigning UntranslatedName = Name in AfterDeserialize method, reminding that despite the direction is reversed, it still hold because of the following 3 points:
(a) there is another line far away that sets both Name and UntranslatedName to an identical value
(b) there are in total two cases assigning Name
(c) only the first case mentioned in (a) can trigger AfterDeserialize method, concluding the soundness

Instead, by applying this PR we can just do it in a straightforward way -- save UntranslatedName and calculate Name based on UntranslatedName. This is self-explainary, without having to write that much details above. We can even remove the AfterDeserialize method.

Metadorius: Maybe, but I propose to do that as a separate PR since this has nothing to do with the scope of this PR.

SadPencil: > Maybe, but I propose to do that as a separate PR since this has nothing to do with the scope of this PR.

If so, that PR increases cache version to v2 and this PR increases it to V3

Metadorius: wait, why does this PR increase cache version?

Metadorius: and we don't see a release soon, so it shouldn't matter as much I think?

SadPencil: > and we don't see a release soon, so it shouldn't matter as much I think?

if we leave one commit to bump the cache version and does not touch the cache version in the other commit, this behavior introduces an implicit assumption that these two commits must bind to each other as a whole, and people cannot just cherry pick or revert one of the commit (more importantly, the commit that does not touch the cache version leaves people a FALSE information that the commit is a small change without invalidating anything, which is not true). This is worse than squashing them into one commit or bumping the version to v3 IMO.

SadPencil: > wait, why does this PR increase cache version?
> 
> 

Because this PR catches the file size and last modified time

Metadorius: > Because this PR catches the file size and last modified time

it seems trivial to read and upgrade the old format with those changes, isn't it?

Metadorius: > if we leave one commit to bump the cache version and does not touch the cache version in the other commit, this behavior introduces an implicit assumption that these two commits must bind to each other as a whole, and people cannot just cherry pick or revert one of the commit (more importantly, the commit that does not touch the cache version leaves people a FALSE information that the commit is a small change without invalidating anything, which is not true). This is worse than squashing them into one commit or bumping the version to v3 IMO.

That's not what I meant. I meant that we upgrade the version in both commits, but this should have not a lot of impact since we're not making a release anyways.

SadPencil: > > if we leave one commit to bump the cache version and does not touch the cache version in the other commit, this behavior introduces an implicit assumption that these two commits must bind to each other as a whole, and people cannot just cherry pick or revert one of the commit (more importantly, the commit that does not touch the cache version leaves people a FALSE information that the commit is a small change without invalidating anything, which is not true). This is worse than squashing them into one commit or bumping the version to v3 IMO.
> 
> 
> 
> That's not what I meant. I meant that we upgrade the version in both commits, but this should have not a lot of impact since we're not making a release anyways.

Upgrade the version in both commits? You mean these two commits share the same parent commit and use a merge commit to merge them? Otherwise if these two commits are sequentially ordered, only one commit can upgrade the version to v2

Metadorius: > Upgrade the version in both commits? You mean these two commits share the same parent commit and use a merge commit to merge them? Otherwise if these two commits are sequentially ordered, only one commit can upgrade the version to v2

why so complicated? one commit upgrades to v2, another to v3...

SadPencil: I said 

> This is worse than squashing them into one commit or bumping the version to v3 IMO.

which means I accept such an option.

SadPencil: > > Because this PR catches the file size and last modified time
> 
> 
> 
> it seems trivial to read and upgrade the old format with those changes, isn't it?

We have to discard v1 cache anyway so I don't think it is necessary to implement a migration. Besides, it's just a cache and we have greatly accelerated the speed and optimized the RAM consumption. The time cost to build a v3 cache is nearly the same as (in fact, even better than) loading a v1 cache, i.e., players won't experience a worse experience.

Metadorius: > We have to discard v1 cache anyway so I don't think it is necessary to implement a migration. Besides, it's just a cache and we have greatly accelerated the speed and optimized the RAM consumption. The time cost to build a v3 cache is nearly the same as (in fact, even better than) loading a v1 cache, i.e., players won't experience a worse experience.

Ok, that's fair.

PR #896, opened 2025-12-10
Use a shorter timeout for CnCNet player count retrieval

SadPencil:
This PR differs the initial GetCnCNetPlayerCount call (sync) with the upcoming GetCnCNetPlayerCount calls (async), using 1-second and 5-second timeouts respectively. Previously the timeout is 10 second (more previously it was 100 second xD), might be a little bit of long just retrieving a number and therefore unnecessary.


MahBoiDeveloper (approved):

11EJDE11:
Looks fine. I question why we're using a blocking method though. Perhaps we switch to the web label #883 when it's ready.

SadPencil:
> Looks fine. I question why we're using a blocking method though. Perhaps we switch to the web label #883 when it's ready.

Introduced since https://github.com/CnCNet/xna-cncnet-client/commit/4193389ba9980c09827551bdaf27c49829da6137. I don't know why

Metadorius:
I agree with switching to web label btw

PR #899, opened 2026-01-04
Add processor affinity argument for Ares

SadPencil:
Implements https://github.com/CnCNet/cncnet-yr-client-package/pull/793

Ref:

- what's broken now: https://github.com/CnCNet/cncnet-yr-client-package/pull/793#issuecomment-3706774421
> In CnCNetYR before version 9.0.0, ProcessorAffinity was set to 2 via SingleCoreAffinity to make it run on cpu1 (the second processor). Since version 9.0.0, the Ares extension has been introduced, which has a -AFFINITY parameter that also controls ProcessorAffinity and defaults to 1, meaning the game will run on cpu0, the first processor only. Therefore, here we explicitly set the value of this command-line parameter to make it consistent with the behavior before 9.0.0.
- why CPU 1: https://github.com/CnCNet/cncnet-yr-client-package/pull/793#issuecomment-3707370526
> The current issue: running game on CPU 0 exclusively results in a bad performance because of driver implementation and system interrupts etc., which varies from machines to machines
>
> Reference 1: <https://www.reddit.com/r/GlobalOffensive/comments/1dzci78/disabling_core_0_fps_boost/>
> Reference 2: ~3 years ago, my previous PC playing MO 3.3.6 with cnc-ddraw (old version), Windows 10, E5 2666v3, RX 580. With CPU 0 assigned: the game renderers literally like a PowerPoint slide show. With other CPU thread assigned (I made a tool to automatically do it), the game plays smoothly. Note that this observation might differs a lot upon different kinds of hardware.

### DIFF
--- a/ClientGUI/GameProcessLogic.cs
+++ b/ClientGUI/GameProcessLogic.cs
@@ -2,12 +2,15 @@
 using System.Diagnostics;
 using System.IO;
 using System.Runtime.InteropServices;
-using ClientCore;
-using Rampastring.Tools;
-using ClientCore.INIProcessing;
 using System.Threading;
-using Rampastring.XNAUI;
+
+using ClientCore;
+using ClientCore.Enums;
 using ClientCore.Extensions;
+using ClientCore.INIProcessing;
+
+using Rampastring.Tools;
+using Rampastring.XNAUI;
 
 namespace ClientGUI
 {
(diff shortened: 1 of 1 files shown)


ClientGUI/GameProcessLogic.cs
```diff
@@ -124,6 +159,12 @@ public static void StartGameProcess(WindowManager windowManager)
                 else
                     arguments = additionalExecutableName + "-SPAWN";
 
+                if (overrideProcessorAffinity && ClientConfiguration.Instance.ClientGameType == ClientType.Ares)
+                {
+                    // Ares defaults to use CPU 0 exclusively, unless explicitly overridden.
+                    arguments += " " + "-AFFINITY:" + processorAffinity.ToString(CultureInfo.InvariantCulture);
+                }
+
```
Metadorius, line 167: I am going to disappoint you: we can't really rely on this anymore with upcoming SyringeEx, because there command line params to be passed to the exe are defined as `--args="-command -line -params"`.

SadPencil:
Update: hold off this PR because it seems both the spawner and cnc-ddraw/ts-ddraw/ddrawcompat overrides the affinity mask. Currently this PR is only effective on Mental Omega 3.3.6 (Ares only, without phobos) and with old version ddrawcompat/without ddraw.

PR #719, opened 2025-05-16
Rework partial GameModeMap and GameLobby codes

SadPencil:
- Unified duplicated codes in `GameMode` and `Map`. For example, `MultiplayerOnly` is a setting that can be either defined in `GameMode` and `Map`. Now, the duplicated codes are merged. Priority: `Map` > `GameMode`.
- `ForceRandomStartLocations` setting can also be turned on with co-op missions (`IsCoop`)
- `Forced Random Teams` has been renamed to `Forced No Teams` in UI and source codes, because the original meaning seems wrong.
- Fix a bug where the team mapping (in player extra options panel) breaks the team setting with co-op missions (`IsCoop`)
- Support specifying non sequential locations by specifying `AllowedStartingLocations`. For example, in a co-op mission based on a 4P map, the players might be positioned in 1 and 3 instead of 1 and 2, while MaxPlayers=2. The AIs are provided in co-op related settings instead of regular AI players.
- Also contains and closes https://github.com/CnCNet/xna-cncnet-client/pull/519, supporting randomized game mode

Note: I can't separate the features, refactors, and bug fixes as separate PRs, because: (a) they rely on one others, and (b) we don't merge PRs in time -- so it will be much painful maintaining many PRs that depend on one others.

PR #900, opened 2026-01-10
Fix blocking of start positions

### LINKED ISSUE #796: "X" - in Auto Allying function doesn't work
When player activate "Auto Allying" function, he can define teams to start positions (A, B, C, D) and there are also "x" and "-".
As it says there "x" does not allow the player to occupy this position.
But this option does not work. Players still can occupy this position.

image
image

11EJDE11:
This PR enforces the auto ally "X" positions. If the host has set "X" on any starting locations, users will no longer be able to choose them from the dropdown or from the map position selectors. If a user is already in that spot, they will be reset to random [???]. 

Closes #796


SadPencil:
Seems like a (partial) overlap with https://github.com/CnCNet/xna-cncnet-client/pull/719/ which provides the `AllowedStartingLocations` feature

11EJDE11:
> Seems like a (partial) overlap with #719 which provides the `AllowedStartingLocations` feature

Didn't see that! Thanks, closed.

PR #907, opened 2026-01-20
Add "Show in folder" context menu items for maps

### LINKED ISSUE #906: "Open in Folder" feature for Maps
### Description

Client having a right click option to open the map file in its folder would make map sharing/locating a lot easier, Maybe we could also add/delete maps directly from the client too.

11EJDE11:
Adds a context menu item to the map list box and map thumbnail to open the map's folder. The map gets selected in explorer automatically. 

Relatively simple change - will merge in a couple days if no feedback.

Closes #906

### DIFF
--- a/DXMainClient/DXGUI/Multiplayer/GameLobby/MapPreviewBox.cs
+++ b/DXMainClient/DXGUI/Multiplayer/GameLobby/MapPreviewBox.cs
@@ -149,6 +149,7 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
         private XNAContextMenu mapContextMenu;
         private XNAContextMenuItem toggleFavoriteMapItem;
         private XNAContextMenuItem toggleExtraTexturesItem;
+        private XNAContextMenuItem openMapFolderItem;
         private XNAClientButton btnToggleFavoriteMap;
         private XNAClientButton btnToggleExtraTextures;
 
@@ -207,10 +208,17 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
                 SelectableChecker = () => GameModeMap != null,
                 VisibilityChecker = () => extraTextures.Any(x => x.Toggleable)
             };
+            openMapFolderItem = new XNAContextMenuItem()
+            {
+                Text = "Open Map Folder".L10N("Client:Main:OpenMapFolder"),
+                SelectAction = OpenMapFolder,
+                SelectableChecker = () => GameModeMap != null
+            };
             mapContextMenu = new XNAContextMenu(WindowManager);
             mapContextMenu.ClientRectangle = new Rectangle(0, 0, 120, 2);
             mapContextMenu.AddItem(toggleFavoriteMapItem);
             mapContextMenu.AddItem(toggleExtraTexturesItem);
+            mapContextMenu.AddItem(openMapFolderItem);
 
             btnToggleFavoriteMap = new XNAClientButton(WindowManager);
             btnToggleFavoriteMap.IdleTexture = AssetLoader.LoadTexture("favInactive.png");
@@ -268,6 +276,8 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
 
             RefreshExtraTexturesBtn();
         }
+
+        private void OpenMapFolder() => GameModeMap?.Map.OpenContainingFolder();
 
         private void ContextMenu_OptionSelected(int index)
         {
--- a/DXMainClient/DXGUI/Multiplayer/GameLobby/MapPreviewBox.cs
+++ b/DXMainClient/DXGUI/Multiplayer/GameLobby/MapPreviewBox.cs
@@ -149,7 +149,7 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
         private XNAContextMenu mapContextMenu;
         private XNAContextMenuItem toggleFavoriteMapItem;
         private XNAContextMenuItem toggleExtraTexturesItem;
-        private XNAContextMenuItem RevealInExplorerItem;
+        private XNAContextMenuItem showInFolderItem;
         private XNAClientButton btnToggleFavoriteMap;
         private XNAClientButton btnToggleExtraTextures;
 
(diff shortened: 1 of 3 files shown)


Metadorius (approved):
I propose to name it `Reveal in Explorer`, VSCode names it like so. LGTM otherwise

11EJDE11:
> I propose to name it `Reveal in Explorer`, VSCode names it like so. LGTM otherwise

[a4f949f](https://github.com/CnCNet/xna-cncnet-client/pull/907/commits/a4f949fd61f00e79c521469b202691f2eccd3499)

Good idea, thanks!

DXMainClient/Domain/Multiplayer/Map.cs
```diff
@@ -934,6 +934,19 @@ private static Point GetIsoTilePixelCoord(int isoTileX, int isoTileY, string[] a
...
+        /// <summary>
+        /// Opens the folder containing this map in Explorer and selects the map file.
+        /// </summary>
+        public void OpenContainingFolder()
+        {
+            FileInfo mapFileInfo = SafePath.GetFile(CompleteFilePath);
+            if (mapFileInfo.Exists)
+            {
+                // https://stackoverflow.com/questions/13680415/how-to-open-explorer-with-a-specific-file-selected
+                ProcessLauncher.StartShellProcess("explorer.exe", $"/select,\"{mapFileInfo.FullName}\"");
```
SadPencil, line 950: if windows

SadPencil (changes requested):
Also, I have a different opinion on the context menu name. VSCode etc. are more familiar to programmers so we should find similar words from other softwares. Maybe something like Open in Explorer

DXMainClient/DXGUI/Multiplayer/GameLobby/MapPreviewBox.cs
```diff
@@ -207,10 +208,17 @@ public override void Initialize()
                 SelectableChecker = () => GameModeMap != null,
                 VisibilityChecker = () => extraTextures.Any(x => x.Toggleable)
             };
+            RevealInExplorerItem = new XNAContextMenuItem()
+            {
+                Text = "Reveal in Explorer".L10N("Client:Main:RevealInExplorer"),
+                SelectAction = RevealInExplorer,
+                SelectableChecker = () => GameModeMap != null
```
SadPencil, line 215: also check is windows here (Runtime.IsWindowsPlatform? can't remember it)

11EJDE11:
> Also, I have a different opinion on the context menu name. VSCode etc. are more familiar with programmers so we should find similar words from other softwares. Maybe something like Open in Explorer

Fair point. 

Firefox and Chrome use "Show in Folder".
VS uses "Open Containing Folder"
Everything (Voidtools) uses "Open path"
Explorer uses "Open file location"

Seems like there's a bit of variation out there. I like "Open containing folder" best. 

Anyone else want to weigh in? @MahBoiDeveloper @CnCRAZER

DXMainClient/Domain/Multiplayer/Map.cs
11EJDE11: Thanks! [ed62eab](https://github.com/CnCNet/xna-cncnet-client/pull/907/commits/ed62eab57a3a90500c2c041ee7be052ed1fd5c35)

CnCRAZER:
> > Also, I have a different opinion on the context menu name. VSCode etc. are more familiar with programmers so we should find similar words from other softwares. Maybe something like Open in Explorer
> 
> 
> 
> Fair point. 
> 
> 
> 
> Firefox and Chrome use "Show in Folder".
> 
> VS uses "Open Containing Folder"
> 
> Everything (Voidtools) uses "Open path"
> 
> Explorer uses "Open file location"
> 
> 
> 
> Seems like there's a bit of variation out there. I like "Open containing folder" best. 
> 
> 
> 
> Anyone else want to weigh in? @MahBoiDeveloper @CnCRAZER 

Show in Folder and Open File Location is what I would pick but I'm pretty impartial on the matter

MahBoiDeveloper:
Is it possible not only open explorer also as select revealed map?

MahBoiDeveloper:
> > Also, I have a different opinion on the context menu name. VSCode etc. are more familiar with programmers so we should find similar words from other softwares. Maybe something like Open in Explorer
> 
> Fair point.
> 
> Firefox and Chrome use "Show in Folder". VS uses "Open Containing Folder" Everything (Voidtools) uses "Open path" Explorer uses "Open file location"
> 
> Seems like there's a bit of variation out there. I like "Open containing folder" best.
> 
> Anyone else want to weigh in? @MahBoiDeveloper @CnCRAZER

"Show in Folder" sounds more nice

MahBoiDeveloper:
> Is it possible not only open explorer also as select revealed map?

Oh, it actually does

MahBoiDeveloper (approved):
LGTM

DXMainClient/DXGUI/Multiplayer/GameLobby/MapPreviewBox.cs
```diff
@@ -149,6 +149,7 @@ public GameModeMap GameModeMap
         private XNAContextMenu mapContextMenu;
         private XNAContextMenuItem toggleFavoriteMapItem;
         private XNAContextMenuItem toggleExtraTexturesItem;
+        private XNAContextMenuItem RevealInExplorerItem;
```
MahBoiDeveloper:
```suggestion
        private XNAContextMenuItem revealInExplorerItem;
```

SadPencil:
Looks like the majority agree with "Show in Folder"?

Metadorius:
> Also, I have a different opinion on the context menu name. VSCode etc. are more familiar to programmers so we should find similar words from other softwares. Maybe something like Open in Explorer

why does it matter? "Reveal" is not a programmer word, there is nothing programmer-specific here

Metadorius:
> Looks like the majority agree with "Show in Folder"?

I do think that this is an incredibly minor/taste change. The initial change was made because "open containing folder" didn't mean what the action actually does, that is highlighting the file. Everything past that is just proposing variations. I don't mind either of the options voiced as long as it conveys the meaning properly, but I don't think any action is needed.

SadPencil:
> > Also, I have a different opinion on the context menu name. VSCode etc. are more familiar to programmers so we should find similar words from other softwares. Maybe something like Open in Explorer
> 
> why does it matter? "Reveal" is not a programmer word, there is nothing programmer-specific here
> 
> 

But this sentence is less familiar to regular users

11EJDE11:
I guess it isn't always Explorer. Could be Finder or Dolphin. We'll go with "Show in folder".

Metadorius:
> But this sentence is less familiar to regular users

this is incredibly minor and doesn't matter enough to be a point of discussions, also I disagree, there's no standard and it describes what it does just fine

PR #908, opened 2026-01-20
Improve startup performance by reducing file reads

11EJDE11:
For me, the client was reading ~1.4GB of data on startup because every custom map file had to be fully read to calculate its SHA1 hash just to check if it was cached, which kind of defeats the purpose of caching. Fix is to use the file path + last modified to determine cache validity instead of SHA1. SHA1 is still calculated and stored, but only when a map is new or changed. After this change, the client reads about 400mb (once the cache is warmed).


SadPencil:
LGTM

PR #910, opened 2026-01-26
Move `AdminRestarter` class from `ClientCore` to the `DXMainClient` namespace

MahBoiDeveloper:
I suggest that AdminRestarter class belongs to DXMainClient rather than to ClientCore because it requires more build information about which API is used (XNA, GL or DX). Also, previous AdminRestarter class location blocks the migration tool PR from build due to insufficient information about API.


SadPencil:
I am okay with this PR. @11EJDE11

PR #911, opened 2026-01-27
Fix loop of button sounds on multiple message boxes

### LINKED ISSUE #687: Multiple message boxes shown simultaneously results in a loop of button sound and unclickable buttons

SadPencil:
Fixes https://github.com/CnCNet/xna-cncnet-client/issues/687

See also https://github.com/CnCNet/Rampastring.XNAUI/commit/aa35903f75bb3d51dbb57c851e81b9f83d8474fa

图片

### DIFF
--- a/ClientGUI/XNAMessageBox.cs
+++ b/ClientGUI/XNAMessageBox.cs
@@ -233,8 +233,11 @@ namespace ClientGUI
         /// <param name="description">The description of the message box.</param>
         public static void Show(WindowManager windowManager, string caption, string description)
         {
-            var panel = new DarkeningPanel(windowManager);
-            panel.Focused = true;
+            var panel = new DarkeningPanel(windowManager)
+            {
+                Focused = true
+            };
+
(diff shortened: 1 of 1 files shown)


SadPencil:
will merge it in 7 days if there are no comments in https://github.com/Rampastring/Rampastring.XNAUI/pull/62

11EJDE11:
Works well, nice work!

PR #923, opened 2026-02-13
Use composition mode for GameMode and Map

SadPencil:
Previously on PR 719:

图片

### DIFF
--- a/DXMainClient/Domain/Multiplayer/CommonMapConfig.cs
+++ b/DXMainClient/Domain/Multiplayer/CommonMapConfig.cs
@@ -12,7 +12,7 @@ using Rampastring.Tools;
 
 namespace DTAClient.Domain.Multiplayer
 {
-    public abstract class GameModeMapBase
+    public class CommonMapConfig
     {
         public const int MAX_PLAYERS = 8;
 
@@ -82,7 +82,7 @@ namespace DTAClient.Domain.Multiplayer
         [JsonInclude]
         public bool? ForceNoTeams { get; private set; }
 
-        protected void InitializeBaseSettingsFromIniSection(IniSection section, bool isCustomMap)
+        public void InitializeBaseSettingsFromIniSection(IniSection section, bool isCustomMap)
         {
             // MinPlayers
             MinPlayers = section.GetIntValueOrNull(isCustomMap ? "MinPlayer" : "MinPlayers");
--- a/DXMainClient/Domain/Multiplayer/GameMode.cs
+++ b/DXMainClient/Domain/Multiplayer/GameMode.cs
@@ -21,8 +21,11 @@ namespace DTAClient.Domain.Multiplayer
             Initialize();
         }
 
+        // TODO: need to test whether the map cache is broken or not
+
         #region CommonMapConfig
 
+        [JsonInclude]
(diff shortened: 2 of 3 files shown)


SadPencil:
@Metadorius Is this what composition mode for? If so I think it is so dumb to copy-paste the boilerplate code twice

Metadorius:
> @Metadorius Is this what composition mode for? If so I think it is so dumb to copy-paste the boilerplate code twice

I didn't mean you should "forward" things like this. Just write `this.CommonConfig.XYZ` when you need to access something

SadPencil:
> > @Metadorius Is this what composition mode for? If so I think it is so dumb to copy-paste the boilerplate code twice
> 
> I didn't mean you should "forward" things like this. Just write `this.CommonConfig.XYZ` when you need to access something

This is ugly because we have `GameModeMap.MaxPlayers` (returns int), `GameMode.MaxPlayers` (returns int?), and `Map.MaxPlayers` (returns int?). Now two of the threes becomes `this.CommonConfig.XYZ` while the rest one stays in `this.XYZ`

Even defining an interface for CommonConfig does not solve this issue since `GameModeMap` has a different nullablity on these properties

https://github.com/CnCNet/xna-cncnet-client/blob/f1c21365735e66d7460f28dee9e0c4c97c6b434e/DXMainClient/Domain/Multiplayer/GameModeMap.cs#L76-L85

Metadorius:
> This is ugly because we have `GameModeMap.MaxPlayers` (returns int), `GameMode.MaxPlayers` (returns int?), and `Map.MaxPlayers` (returns int?). Now two of the threes becomes `this.CommonConfig.XYZ` while the rest one stays in `this.XYZ`
> 
> Even defining an interface for CommonConfig does not solve this issue since `GameModeMap` has a different nullablity on these properties
> 
> https://github.com/CnCNet/xna-cncnet-client/blob/f1c21365735e66d7460f28dee9e0c4c97c6b434e/DXMainClient/Domain/Multiplayer/GameModeMap.cs#L76-L85


solution: make a CommonConfig.Merge method that would return a merged config that you reference instead of individual things

SadPencil:
>  
> > This is ugly because we have `GameModeMap.MaxPlayers` (returns int), `GameMode.MaxPlayers` (returns int?), and `Map.MaxPlayers` (returns int?). Now two of the threes becomes `this.CommonConfig.XYZ` while the rest one stays in `this.XYZ`
> > 
> > Even defining an interface for CommonConfig does not solve this issue since `GameModeMap` has a different nullablity on these properties
> > 
> > https://github.com/CnCNet/xna-cncnet-client/blob/f1c21365735e66d7460f28dee9e0c4c97c6b434e/DXMainClient/Domain/Multiplayer/GameModeMap.cs#L76-L85
> 
> 
> solution: make a CommonConfig.Merge method that would return a merged config that you reference instead of individual things
> 
> 

What do you mean? A CommonConfig and a MergedCommonConfig class? if not, how to make a different nullability?

SadPencil:
Based on previous discussion shown below, this PR is now paused due to the nullability issue. This composition pattern brings some heavy usages of `!` operator and therefore is a heavy load to maintain and makes our codes (files with `#nullable enable`) less reliable against null reference exception.

Issue #940, opened 2026-03-04
FileSettingCheckBox stops working after migrating CampaignSelector to INItializableWindow


SadPencil:
### Description

it looks like after migrating CampaignSelector to INItializableWindow, the component such as FileSettingCheckBox will stop working as expected, probably due to FileSettingCheckBox.GetAttributes() not being called anymore (not sure). 

### XNA Client Version

3e52e8be3ef16f866b6d84dbc814b8e3b543d6b1@develop

### Steps To Reproduce

1. Apply this file [CampaignSelector.ini.txt](https://github.com/user-attachments/files/25732343/CampaignSelector.ini.txt) to cncnet yr, with latest client binaries in develop branch
2. Run the client. Tick `chkCustomMission` in campaign selector
3. Observe whether `EnabledFile0` and `EnabledFile1` are copied or not

### Expected Behaviour

The files should be copied.

### Actual Behaviour

The files are not copied.

### Checklist

- [x] The issue happens on the **latest official** version of XNA Client and wasn't fixed yet.
- [x] I agree to elaborate the details if requested and provide thorough testing if the bugfix is implemented.
- [x] I added a very descriptive title to this issue.
- [x] I used the GitHub search and read the issue list to find a similar issue and didn't find it.
- [x] I have attached as much information as possible *(screenshots, gifs, videos, client logs, etc)*.

SadPencil:
Close by #941. Note that we still need to handle this issue when we migrate CampaignSelector to INItializableWindow in the future

PR #941, opened 2026-03-04
Revert migrating CampaignSelector to INItializableWindow

SadPencil:
This PR reverts the INItializableWindow migration in #922 if we can't solve issue https://github.com/CnCNet/xna-cncnet-client/issues/940


SadPencil:
I'll merge this PR in about a week perhaps if there is not a clear path to directly handle the issue 940

SadPencil:
According to the latest conversation in Discord, this revert is now performed immediately

PR #943, opened 2026-03-06
Support settings keys and hotkey config embedded in settings INI for RA1

CO2-code:
Fixes an issue where hotkeys and client settings overwrite each other when using same INI file.
Updated UserINISettings key mapping to match RA1's expected keys.
Made the hotkey section configurable (allowing RA1 to use WinHotkeys while other games continue using Hotkey).

### DIFF
--- a/ClientCore/ClientConfiguration.cs
+++ b/ClientCore/ClientConfiguration.cs
@@ -367,7 +367,11 @@ namespace ClientCore
 
         public bool SettingsIniAsKeyboardIni => SettingsIniName == KeyboardINI;
         
-        public string KeyboardHotkeySection => clientDefinitionsIni.GetStringValue(SETTINGS, "KeyboardHotkeySection", "Hotkey");
+        public string KeyboardHotkeySection => clientDefinitionsIni.GetStringValue(
+            SETTINGS,
+            "KeyboardHotkeySection",
+            ClientGameType == ClientType.RA ? "WinHotKeys" : "Hotkey"
+        );
 
         public int MinimumIngameWidth => clientDefinitionsIni.GetIntValue(SETTINGS, "MinimumIngameWidth", 640);
 
--- a/ClientCore/ClientConfiguration.cs
+++ b/ClientCore/ClientConfiguration.cs
@@ -365,6 +365,8 @@ namespace ClientCore
 
         public string KeyboardINI => clientDefinitionsIni.GetStringValue(SETTINGS, "KeyboardINI", "Keyboard.ini");
 
+        public bool SettingsIniAsKeyboardIni => SettingsIniName == KeyboardINI;
+        
         public string KeyboardHotkeySection => clientDefinitionsIni.GetStringValue(SETTINGS, "KeyboardHotkeySection", "Hotkey");
 
         public int MinimumIngameWidth => clientDefinitionsIni.GetIntValue(SETTINGS, "MinimumIngameWidth", 640);
--- a/ClientCore/ClientConfiguration.cs
+++ b/ClientCore/ClientConfiguration.cs
@@ -365,6 +365,8 @@ namespace ClientCore
 
         public string KeyboardINI => clientDefinitionsIni.GetStringValue(SETTINGS, "KeyboardINI", "Keyboard.ini");
 
+        public string KeyboardHotkeySection => clientDefinitionsIni.GetStringValue(SETTINGS, "KeyboardHotkeySection", "Hotkey");
+
         public int MinimumIngameWidth => clientDefinitionsIni.GetIntValue(SETTINGS, "MinimumIngameWidth", 640);
 
         public int MinimumIngameHeight => clientDefinitionsIni.GetIntValue(SETTINGS, "MinimumIngameHeight", 480);
--- a/ClientCore/ClientConfiguration.cs
+++ b/ClientCore/ClientConfiguration.cs
@@ -366,12 +366,11 @@ namespace ClientCore
         public string KeyboardINI => clientDefinitionsIni.GetStringValue(SETTINGS, "KeyboardINI", "Keyboard.ini");
 
         public bool SettingsIniAsKeyboardIni => SettingsIniName == KeyboardINI;
-        
+
         public string KeyboardHotkeySection => clientDefinitionsIni.GetStringValue(
             SETTINGS,
             "KeyboardHotkeySection",
-            ClientGameType == ClientType.RA ? "WinHotKeys" : "Hotkey"
-        );
+            ClientGameType == ClientType.RA ? "WinHotKeys" : "Hotkey");
 
         public int MinimumIngameWidth => clientDefinitionsIni.GetIntValue(SETTINGS, "MinimumIngameWidth", 640);
 
--- a/ClientCore/Settings/UserINISettings.cs
+++ b/ClientCore/Settings/UserINISettings.cs
@@ -88,97 +88,6 @@ namespace ClientCore
         {
             SettingsIni = iniFile;
 
-           if (ClientConfiguration.Instance.ClientGameType == ClientType.RA)
-            {
-                // ====================
-                // Display (Resolution)
-                // ====================
-                IngameScreenWidth = new IntSetting(iniFile, OPTIONS, "Width", 1024);
-                IngameScreenHeight = new IntSetting(iniFile, OPTIONS, "Height", 768);
-                ClientTheme = new StringSetting(iniFile, MULTIPLAYER, "Theme", ClientConfiguration.Instance.GetThemeInfoFromIndex(0).Name);
-                Translation = new StringSetting(iniFile, OPTIONS, "Translation", I18N.Translation.GetDefaultTranslationLocaleCode());               
-                Renderer = new StringSetting(iniFile, COMPATIBILITY, "Renderer", string.Empty);               
-                BorderlessWindowedClient = new BoolSetting(iniFile, VIDEO, "BorderlessWindowedClient", ClientConfiguration.Instance.UserDefault_BorderlessWindowedClient);
-                IntegerScaledClient = new BoolSetting(iniFile, VIDEO, "IntegerScaledClient", ClientConfiguration.Instance.UserDefault_IntegerScaledClient);
-                ClientFPS = new IntSetting(iniFile, VIDEO, "ClientFPS", 60);
-                DisplayToggleableExtraTextures = new BoolSetting(iniFile, VIDEO, "DisplayToggleableExtraTextures", true);
-
-                // ====================
-                // GAME
-                // ====================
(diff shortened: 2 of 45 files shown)


ClientCore/Settings/UserINISettings.cs
Metadorius, line 1: why are you duplicating ALL of the settings? you should surgically change only the ones which are different for RA1

ClientGUI/HotkeyConfigurationWindow.cs
```diff
@@ -15,8 +15,7 @@ namespace ClientGUI
     /// </summary>
     public class HotkeyConfigurationWindow : XNAWindow
     {
-        private readonly string HOTKEY_TIP_TEXT = "Press a key...".L10N("Client:DTAConfig:PressAKey");
-        private const string HOTKEY_INI_SECTION = "Hotkey";
+        private readonly string HOTKEY_TIP_TEXT = "Press a key...".L10N("Client:DTAConfig:PressAKey");
```
Metadorius: what changed in this specific line? did line ending change? that should be reverted as it litters the history with unnecessary changes

ClientCore/Settings/UserINISettings.cs
```diff
@@ -200,6 +300,7 @@ protected UserINISettings(IniFile iniFile)
         /* AUDIO */
         /*********/
 
+        public DoubleSetting MultiplayerScoreVolume { get; private set; }
```
Metadorius: not sure what is this for, some other score volume?

Metadorius (commented):
The keyboard INI change is good and if you split it out it can be merged even now, the rest needs more refinement

Metadorius:
> The keyboard INI change is good

although keep in mind that this is a basic review of only changed files, I suspect there may be more places to handle this

Metadorius:
have you handled "reset to defaults" part of the keyboard INI to work with RA1 setup and your changes?

CO2-code:
> have you handled "reset to defaults" part of the keyboard INI to work with RA1 setup and your changes?

I think it is already done as it uses WriteKeyboardINI() , right ?

ClientCore/Settings/UserINISettings.cs
CO2-code: > not sure what is this for, some other score volume?

yes ScoreVolume and MultiplayerScoreVolume are the game's "Music volume"
image

ClientCore/Settings/UserINISettings.cs
CO2-code: i didn't want to use a lot of ifs 
and  i didn't know any other way :)
do u want me to do as u said (you should surgically change only the ones which are different for RA1)?

Metadorius: yes, of course
shouldn't worry about the amount of ifs. also use ternary operators for the key names. also you can do something like `bool isRA = client type == RA1;` (pseudocode) and then `isRA ? "RAKeyName" : "UsualKeyName"` inside

ClientCore/Settings/UserINISettings.cs
Metadorius: should be done in another way probably, not sure. @SadPencil @11EJDE11 opinions?

ClientGUI/HotkeyConfigurationWindow.cs
CO2-code: I deleted (private const string HOTKEY_INI_SECTION = "Hotkey";) only

ClientCore/Settings/UserINISettings.cs
CO2-code: > should be done in another way probably, not sure. @SadPencil @11EJDE11 opinions?

can you re-review please

Metadorius (commented):
looks much better now! remind me to review in more details on weekend

CO2-code:
> looks much better now! remind me to review in more details on weekend

for me the weekend is Saturday and Sunday , which starts tomorrow :)

Metadorius:
yea correct, I'll just forget by tomorrow likely 🤣

ClientCore/Settings/UserINISettings.cs
SadPencil: > > not sure what is this for, some other score volume?
> 
> yes ScoreVolume and MultiplayerScoreVolume are the game's "Music volume" image

Can we assume ScoreVolume and MultiplayerScoreVolume always have the same values?

CO2-code: > > > not sure what is this for, some other score volume?
> > 
> > 
> > yes ScoreVolume and MultiplayerScoreVolume are the game's "Music volume" image
> 
> Can we assume ScoreVolume and MultiplayerScoreVolume always have the same values?

I am 100% sure that they should always have the same values , according to Iran's RAConfigTool ,

 double MusicVolume = ((double)this.Slider_MusicVolume.Value) / 1000;
Files.RedAlertINI.setStringValue("Options", "ScoreVolume", MusicVolume.ToString());
Files.RedAlertINI.setStringValue("Options", "MultiplayerScoreVolume", MusicVolume.ToString());

CO2-code:
> yea correct, I'll just forget by tomorrow likely 🤣

Reminder :)
also don't forget to check why hotkeys getting deleted after Clickin "Save" button in the options window 
Edit : I think I have fixed that tho , just let me know if thats the correct way :)
Edit 2 : It only works if  hotkeys section exists before firstluanch :( , but assigning doesn't work :(

ClientCore/Settings/UserINISettings.cs
```diff
@@ -48,6 +48,12 @@ public static void Initialize(string userIniFileName)
 
             var userIni = new IniFile(SafePath.CombineFilePath(ProgramConstants.GamePath, userIniFileName));
 
+            if (ClientConfiguration.Instance.ClientGameType == ClientType.RA)
+            {
+                _instance = new UserINISettings(userIni);
+                return;
+            }
```
Metadorius: what is this for?

DXMainClient/Resources/ClientDefinitions.ini
```diff
@@ -104,6 +104,9 @@ RecommendedResolutions=1280x720	;,1360x768,1366x768,2560x1440
 ; in the options menu
 KeyboardINI=Keyboard.ini
 
+; Hotkey section in KeyboardINI , for RA = WinHotkeys
```
Metadorius:
```suggestion
; Hotkey section in KeyboardINI, for RA this should be WinHotkeys
```
also copy this to docs

ClientCore/Settings/UserINISettings.cs
```diff
@@ -493,7 +520,7 @@ private void ResetGameOptionFilters()
         /// </summary>
         private void CleanUpLegacySettings()
             => SettingsIni.GetSection(GAME_FILTERS).RemoveKey("SortAlpha");
-
+
```
Metadorius:
```suggestion

```

ClientGUI/HotkeyConfigurationWindow.cs
```diff
@@ -7,6 +7,7 @@
 using Rampastring.XNAUI.XNAControls;
 using System;
 using System.Collections.Generic;
+using ClientCore.Enums;
```
11EJDE11: Remove this, it's unused.

ClientGUI/HotkeyConfigurationWindow.cs
```diff
@@ -482,13 +487,18 @@ private KeyModifiers GetCurrentModifiers()
...
             foreach (var command in gameCommands)
             {
-                keyboardIni.SetStringValue("Hotkey", command.ININame, command.Hotkey.GetTSEncoded().ToString());
+                keyboardIni.SetStringValue(ClientConfiguration.Instance.KeyboardHotkeySection, command.ININame, command.Hotkey.GetTSEncoded().ToString());
             }
 
-            keyboardIni.WriteIniFile(SafePath.CombineFilePath(ProgramConstants.GamePath, ClientConfiguration.Instance.KeyboardINI));
+            // TODO should it be like this and main settings window will handle it, or should we flush even here, saving all settings?
+            if (!ClientConfiguration.Instance.SettingsIniAsKeyboardIni)
+                keyboardIni.WriteIniFile(SafePath.CombineFilePath(ProgramConstants.GamePath, ClientConfiguration.Instance.KeyboardINI));
```
11EJDE11: For consistency I'd say flush here. The hotkey window has its own save button.

```c#

  if (ClientConfiguration.Instance.SettingsIniAsKeyboardIni)
       keyboardIni.WriteIniFile();
  else
       keyboardIni.WriteIniFile(SafePath.CombineFilePath(ProgramConstants.GamePath, ClientConfiguration.Instance.KeyboardINI));
```

Metadorius: just as you said this I realized it's not gonna work. Imagine a situation when the user pressed save in the hotkey config window, then decided they don't want config changes they did in the options.

If we don't flush here, everything (including hotkeys) can be restored by hitting a cancel. 

If we flush here - you can't cancel anymore at all.

The hotkeys erased/not is a discrepancy in behavior, though, but I'd argue it's not as critical and the behavior with unified INI might be more logical

11EJDE11: Makes sense, what a pain. I guess just add a comment explaining that for now.

ClientCore/ClientConfiguration.cs
```diff
@@ -365,6 +365,10 @@ private List<TranslationGameFile> ParseTranslationGameFiles()
 
         public string KeyboardINI => clientDefinitionsIni.GetStringValue(SETTINGS, "KeyboardINI", "Keyboard.ini");
 
+        public bool SettingsIniAsKeyboardIni => SettingsIniName == KeyboardINI;
+        
+        public string KeyboardHotkeySection => clientDefinitionsIni.GetStringValue(SETTINGS, "KeyboardHotkeySection", "Hotkey");
+
```
SadPencil, line 374: Consider that we already introduced so many hard coded ini key names for RA, when providing the default value of `KeyboardHotkeySection`, provide "WinHotKeys" as the default value when it's RA.

ClientGUI/HotkeyConfigurationWindow.cs
CO2-code: Done

SadPencil:
I am working on hotkey related codes. Will check your PR too

ClientCore/Settings/UserINISettings.cs
```diff
@@ -63,7 +63,9 @@ public static void Initialize(string userIniFileName)
 
             // Combine userIni and userDefaultIni
             foreach (string sectionName in userIni.GetSections())
-            {
+            {   
+                // TODO should we somehow handle hotkey section specially here when settings ini = keyboard ini?
```
SadPencil: I think not? I mean I think we can just leave the code as is

SadPencil (approved):

PR #948, opened 2026-03-14
Add TTF/OTF support

11EJDE11:
Add documentation for TTF/OTF font support (Docs/Fonts.md).


SadPencil:
This PR can be named as "Add TTF/OTF support". Note that you have to upgrade the git submodule to upgrade XNAUI

Metadorius (approved):

Metadorius:
docs aside, I think fallbacks should be unique per font and instead be a single number pointing to another font index, for the reasons of being able to define unique fallbacks for different style fonts and different sizings between different fonts. I don't remember the code around that part tho, would that be complicated to change?

Metadorius:
single number because we could recursively check the other font for characters

SadPencil:
@11EJDE11 I agree with the per-font fallback font thing.

Besides, I find it very difficult to find the correct font size to make the rendering **pixel-perfect**. Below is the **only** pixel-perfect combination I have found among ~10 fonts, but the font is still a little bit wider than expected.

```ini
Path=unifont-17.0.04.otf
Size=16
```

Note: to observe the pixel-perfectness from the screenshot below, one need to download the png file and use `mspaint` or other photo viewers supporting an exact 100% scaling, instead of the browser.

无标题

SadPencil:
Given an arbitrary TTF file (especially for some fonts already optimized with pixel), is it easy to determine the `Size` property to make the text pixel-perfect, or can the client automatically select a size to make it pixel-perfect?

SadPencil:
Now I find a better TTF font:

WenQuanYi.Bitmap.Song.16px.ttf with size **18**

but why 18 makes the font pixel-perfect? What's the unit of `Size`, and does it refer to the width pixel, the height pixel, or pt? It is very confusing when only a word `Size` is given (Edit: the doc says the unit is "pixel" without mentioning it is the width or the height)

I manually tested in MS Word and specifying 13.5 pt makes the font WenQuanYi.Bitmap.Song.16px.ttf pixel-perfect, which means 18 px (on 96 DPI). But this procedure heavily relies on human eyes.

SadPencil:
Below is the number 2 (the third one) TTF file extracted from `wqy-zenhei.ttc`. This file is even weirder: even MS word shows the font is pixel perfect with 12 pt (note: 50% scaling in MS word and 200% DPI for my monitor), setting the size to 16 px (which corresponds to 12 pt) does not get a pixel perfect result in the client.

[wqy-zenhei.ttc#2.zip](https://github.com/user-attachments/files/26168559/wqy-zenhei.ttc.2.zip)

SadPencil:
I also tried `simsun#0.ttf` from `simsun.ttc` and it results the same with `wqy-zenhei.ttc`.

One common thing: these two fonts contains bitmap fonts specialized for common font sizes and it looks like FontStash won't use them

图片

As a result, we won't get a pixel-perfect result when using these fonts for now

SadPencil:
Yes. The original font stash project does not contain bitmap font handling. Only a fork adds such a support:

https://github.com/akrinke/Font-Stash

And even so, the fork does not support directly reading the bitmap font from the TTF file.

It is a severe issue that many fonts (especially fonts with CJK support) can't be rendered in a pixel-perfect way even the font itself supports it -- this makes the TTF font get a worse rendering than the sprite font. Due to this limitation, I think we might need to either delay introducing TTF support or deprecating using TTF fonts for now.

See also: https://github.com/itouhiro/sbitget for some tech details

SadPencil (changes requested):
I'd like a conversation in Discord

SadPencil:
Another thought: for a ttf font with bitmap fonts embedded, try to follow the same way as how WenQuanYi.Bitmap.Song.16px.ttf file was created.

https://github.com/AmusementClub/WenQuanYi-Bitmap-Song-TTF

java -jar BitsNPicas.jar convertbitmap -f ttf -o "wenquanyi_12px.ttf" "wenquanyi_12px.bdf"

SadPencil:
Sample text:
> 憂鬱的臺灣烏龜

This text is included in `simsun14.ttf`, not included in `SpriteFont0/1.xnb` file. However, the fallback font is not correctly selected in this case

图片

图片

See DXMainClient\Resources\DTA\Fonts.ini

Docs/Fonts.md
```diff
@@ -0,0 +1,317 @@
...
+
+### SpriteFont (legacy)
+
+```ini
+[Fonts]
+Count=4
+
+[Font0]
+Type=SpriteFont
+Path=SpriteFont0
```
SadPencil, line 267: should the extension .xnb be kept or not?

SadPencil:
> Sample text:
> 
> > 憂鬱的臺灣烏龜
> 
> 
> 
> This text is included in `simsun14.ttf`, not included in `SpriteFont0/1.xnb` file. However, the fallback font is not correctly selected in this case
> 
> 
> 
> 图片
> 
> 
> 
> 图片
> 
> 
> 
> See DXMainClient\Resources\DTA\Fonts.ini

I just realized that fallback for a SpriteFont is not feasible. Probably needs to mention that in doc and logger

Metadorius:
why is it not feasible?

11EJDE11:
Fallback is for when there are no characters in the primary font that match so it falls back to one of the other fonts. It's all handled by FSS, so adding a SpriteFont into the mix just isn't really possible, two different systems. And not sure why you would want to anyway, TTFs should have all the characters you need. I see it more like Arial would be your main font but you'd have Segoe UI Emoji as a fallback so emojis get rendered.

SadPencil:
@11EJDE11 

> so adding a SpriteFont into the mix just isn't really possible

Yes, this is what I realized last week after going deep to the technical details. Even *I* didn't realize it for a long time, when my mind is occupied selecting different fonts. This indicates we must clearly write the documentation and *remove any technical assumption that the user must know*. If the doc does not say we can't mix them, then the client should allow such a thing -- therefore, we need to enhance the documentation 

> not sure why you would want to anyway

Such a thought is quite natural, and it’s a very likely approach many would take.​ The reason is that instead of building from scratch, many client users are upgrading from older versions, where all their UIs and existing sprite fonts are already optimized. They’d think: “Okay, the TTF feature improves compatibility for languages I don’t really care about yet. So, *what’s the fastest, compatible way to adopt TTF without spending time selecting, testing, and verifying that new fonts won’t break my current layout?*” The answer they’d land on is: use my existing sprite font as the primary font, and use any TTF font as a fallback.

To serve these users, we should:
1.    Explicitly state that sprite font and TTF font cannot be mixed in documentation.
2.    Provide a recommended English TTF font that looks very close to our existing sprite font.
3.    Provide a recommended fallback TTF font (I would recommend either the simsun14.ttf file I made, or unifont-17.0.04.otf)

11EJDE11:
I've added per-font fallbacks. Sorry - committed to master - please double check:
https://github.com/CnCNet/Rampastring.XNAUI/commit/deb7fa656ec60893789d9ba63f1ff57d598a1699

Docs have been update to address new fallback method and all feedback posted.

SadPencil:
It seems that even if the TTF font does not have an embedded bitmap font, the rendering is still notably blurry compared with a "preprocessed" font, especially consider that we have many users with a High DPI screen so it's often that the client chooses a 200% scaling. See the screenshot below. The `simsun14.ttf` contains English characters from `arial.ttf`.

Note: the screenshot might not be able to show the feeling. It's better running the client to observe

图片
图片

I also tried using arial.ttf with size 14, 15, or 16 but none of them provides the same sharp looking (like our current sprite font). Therefore, this means ALL of TTF fonts need to be "preprocessed"

SadPencil:
I am working on the Forme implementation (it still lacks a cache and lacks harfbuzz) but your FontStashSharp PR should likely to finish before mine

remaining points to be done for FontStashSharp based TTF:
(1) revert some doc changes (I'll do it later) to not recommending "preprocess" the fonts. But we still need to admit the current engine is not perfect and bad result does not indicate a font problem
(2) allow KernelWidth/KernalHeight/ the factor to be customized in ini. Default the factor to 2 or even 3. Per-font perhaps (not sure)
(3) "Update the common assembly list" https://github.com/CnCNet/xna-cncnet-client/blob/develop/Scripts/README.md (update: https://github.com/CnCNet/xna-cncnet-client/commit/a88b449dfe3eea53210db08fbd9e985025e9b78c)
(4) swtich to `stbtt_ScaleForMappingEmToPixels` and observe if the text in the dropdown has the correct vertical position (using multiple fonts to observe, say Arial and Noto Sans)

PR #962, opened 2026-03-28
Fix/color selection

CO2-code:
fixes color selection.
if player X picks yellow then player Y won't be able to pick yellow too 

Before:
image

After:
image


CnCRAZER:
IMO:
The color box **and** the color text within color dropdown need to be grey'd out incase clients use one or the other(if not both)

Should also consider adding this to the start location as well since multiple players can not start at same locations.

SadPencil:
Gray? Imagine a modder select using pure colors instead of color + text for the color selection dropdown

Besides, imagine players wants to defeat AIs, or there are fan-made campaign maps that requires multiple AIs using same colors. These are possible (although rare) situation where same colors need to be specified

CO2-code:
> Gray? Imagine a modder select using pure colors instead of color + text for the color selection dropdown
> 
> Besides, imagine players wants to defeat AIs, or there are fan-made campaign maps that requires multiple AIs using same colors. These are possible (although rare) situation where same colors need to be specified

I think removing the taken color from the dropdown list is better

PR #979, opened 2026-04-12
Fix intermittent crash on exit

11EJDE11:
Crash dump analysis showed the faulting thread was an XAUdio2 graph worker:
```
xaudio2_9!CGraphManager::GraphThreadProc
xaudio2_9!CX2Engine::OnProcessingPassStart
clr!UMThunkStub
clr!UMThunkStubRareDisableWorker      <-- raises 0xC0020001
```

XAudio2 could invoke managed callbacks after CLR teardown began, causing a native crash (0xC0020001) on a background thread (didn't appear in client log).

Goes with: https://github.com/CnCNet/Rampastring.XNAUI/pull/14

### DIFF
--- a/DXMainClient/DXGUI/Generic/MainMenu.cs
+++ b/DXMainClient/DXGUI/Generic/MainMenu.cs
@@ -1136,10 +1136,6 @@ namespace DTAClient.DXGUI.Generic
             Logger.Log("Exiting.");
             WindowManager.CloseGame();
             themeSong?.Dispose();
-#if !XNA
-            Thread.Sleep(1000);
-            Environment.Exit(0);
-#endif
         }
 
         public void SwitchOn()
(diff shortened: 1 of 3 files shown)


SadPencil:
@11EJDE11 is it tested in all 3 engines and both net48 and net8?

SadPencil:
i tested on ugl, dx net48, gl net48 in vs and this pr works like a charm

SadPencil:
I feel the word intermittent hard to understand what is wrong. is there a better pr title?

Metadorius:
what's wrong with intermittent?

SadPencil:
> what's wrong with intermittent?

seems like an "advanced" or "high-level" English word to say "occasional" (we use "occasional" 3 times in commit messages and 2 times in the source)

Metadorius:
I don't think this needs dumbing down, and it's not an uncommon word really

SadPencil:
> I don't think this needs dumbing down, and it's not an uncommon word really

fine

PR #980, opened 2026-04-14
Add CustomDragDistance setting which overrides DragDistance

11EJDE11:
Based on feedback from `skylegend` on Discord, TS requires a custom drag distance. Rampastring was of the opinion it would be bloat to add something to the GUI so I have just made it a setting. 

```ini
[Options]
CustomDragDistance=<value> ;0 will retain the normal calculation. Anything else will change DragDistance to whatever is specified.
```

The user must manually add the value in whatever the TS equivalent of `RA2MD.ini` is (rules.ini ?)

### DIFF
--- a/ClientCore/Settings/UserINISettings.cs
+++ b/ClientCore/Settings/UserINISettings.cs
@@ -240,6 +240,7 @@ namespace ClientCore
 
         public IntSetting ScrollRate { get; private set; }
         public IntSetting DragDistance { get; private set; }
+        // When > 0, overrides the auto-scaled DragDistance. Allows players to set a fixed pixel threshold regardless of resolution.
         public IntSetting CustomDragDistance { get; private set; }
         public IntSetting DoubleTapInterval { get; private set; }
         public StringSetting Win8CompatMode { get; private set; }
--- a/ClientCore/Settings/UserINISettings.cs
+++ b/ClientCore/Settings/UserINISettings.cs
@@ -139,6 +139,7 @@ namespace ClientCore
 
             ScrollRate = new IntSetting(iniFile, OPTIONS, "ScrollRate", 3);
             DragDistance = new IntSetting(iniFile, OPTIONS, "DragDistance", 4);
+            CustomDragDistance = new IntSetting(iniFile, OPTIONS, "CustomDragDistance", 0);
             DoubleTapInterval = new IntSetting(iniFile, OPTIONS, "DoubleTapInterval", 30);
             Win8CompatMode = new StringSetting(iniFile, OPTIONS, "Win8Compat", "No");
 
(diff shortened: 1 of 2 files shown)


SadPencil:
what does DragDistance do btw?

11EJDE11:
No idea. @Rampastring ?
https://github.com/CnCNet/ts-patches/blob/7bb632d4b7014a4aef11d7461aec2f20f45fa472/src/mouse_behavior.asm#L10

Rampastring:
When you press down the left mouse button, as a player, you might be either performing a click, or about to do drag selection. The game could differentiate between these two just based on whether the mouse moved between mouse-down and mouse-up.

However, such an approach would be error-prone. When people press down the left mouse button, they often accidentally also move the mouse by 1-2 pixels.

To prevent accidental drag selection inputs instead of clicks in such cases, the game has a threshold. An input is not considered a drag-selection unless the mouse cursor has moved at least 4 pixels between the mouse-down point and the current mouse location. Some newer "C&C-like" RTS games, like Tempest Rising, lack this threshold, and as a player it's noticeable that you often make accidental drag-selection inputs instead of ordering units in those games.

These 4 pixels, hardcoded in the original game, were enough of a safeguard when the game was released in '99. The most common resolution to play with was 640x480, some people used 800x600.

Nowadays, people use much higher resolutions. 1280x720, 1920x1080, some even higher. What was a 1-2 pixel accidental movement around the turn of the millenium, is nowadays easily 4+ pixels. This causes people to make accidental drag selection inputs in hectic high-APM gameplay, making them not perform clicks as they intended.

DragDistance is the fix. ts-patches allows customizing the threshold. The client was made to scale DragDistance based on the resolution's width following the original design: if 4 pixels was enough for 640x480, then 8 pixels should be good for 1280x720, for 1920x1080 you'd want 12 pixels, and so on. The modding community has relied on this solution for a decade.

Apparently FunkyFr3sh's CnCNet client made it customizable instead of scaling it automatically, and some in the TS "pro" community are used to having it as high as 32 pixels, regardless of their resolution.

11EJDE11:
Great explanation, thanks. 32 is wild.

Metadorius:
I kind of feel that the resolution is the wrong scale factor, and instead DPI or DPI scale should be the one to control that, but I digress.

Rampastring:
I can tell you that if we based it on DPI instead, most people wouldn't find it sufficient. The primary reason for lower resolutions back in the days wasn't small display size, but performance. Many people would have used higher resolutions if PCs back then could have handled them smoothly. I believe most people still use 96 DPI today, the same they used back in Windows 9x, despite resolutions having multiplied over the last 30 years. They're just enjoying the extra clarity and desktop work space.

Resolution makes it scale naturally for modern desktop resolutions anyway as the game is upscaled to fullscreen by renderers such as CnC-DDRAW. Effectively, a 8-pixel threshold at 1280x720 becomes 12 pixels when that 720p is stretched to cover a 1080p area. When the user instead runs 1080p, and the client sets the DragDistance to 12, they keep the same physical threshold for activating drag selection.

People can also choose to scale elements based on other methods rather than DPI. I have a 32-inch 4K display and I run it at 100% DPI. Instead I've increased font scale in Windows and sometimes use zoom levels in applications that support them (like browsers, Discord, WAE etc.) to keep things readable.

SadPencil:
Perhaps we need to summarize the key points as code comments

PR #986, opened 2026-04-18
Allow overriding CustomIniPath when loading game mode map code

11EJDE11:
This changes GameMode to use `CustomIniPath` from `MPMaps.ini` when loading the game mode's map-code INI, instead of
always falling back to `INI/Map Code/<GameModeName>.ini`.

Why this is needed:

- Some game modes define a custom map-code filename that does not match the mode name.
- Blitz 2v2 is one example: it uses Blitz_2v2.ini, not Blitz 2v2.ini.
- Before this change, the loader ignored CustomIniPath, so those modes could silently miss their intended map rules.

What changed:

- Read CustomIniPath in GameMode.Initialize().
- Use it as the primary map-code path.
- Keep the existing MapCodeININame behavior as a fallback when CustomIniPath is not set.

### DIFF
--- a/DXMainClient/Domain/Multiplayer/GameMode.cs
+++ b/DXMainClient/Domain/Multiplayer/GameMode.cs
@@ -90,7 +90,7 @@ namespace DTAClient.Domain.Multiplayer
 
             forcedOptionsSection = section.GetStringValue("ForcedOptions", string.Empty);
             string customIniPath = section.GetStringValue("CustomIniPath", string.Empty);
-            mapCodeININame = section.GetStringValue("MapCodeIniName", section.GetStringValue("MapCodeININame", Name + ".ini"));
+            string mapCodeIniName = section.GetStringValue("MapCodeIniName", section.GetStringValue("MapCodeININame", Name + ".ini"));
             mapCodeIniPath = string.IsNullOrWhiteSpace(customIniPath)
                 ? SafePath.CombineFilePath(BASE_INI_PATH, mapCodeIniName)
                 : customIniPath;
--- a/DXMainClient/Domain/Multiplayer/GameMode.cs
+++ b/DXMainClient/Domain/Multiplayer/GameMode.cs
@@ -62,7 +62,7 @@ namespace DTAClient.Domain.Multiplayer
 
         public int? MaxPlayersOverride { get; private set; }
 
-        private string mapCodeININame;
+        private string mapCodeIniPath;
         private List<string> randomizedMapCodeININames;
         private int randomizedMapCodesCount;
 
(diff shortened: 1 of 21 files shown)


DXMainClient/Domain/Multiplayer/GameMode.cs
```diff
@@ -89,7 +89,11 @@ public void Initialize()
             MaxPlayersOverride = section.GetIntValueOrNull("MaxPlayersOverride");
 
             forcedOptionsSection = section.GetStringValue("ForcedOptions", string.Empty);
-            mapCodeININame = section.GetStringValue("MapCodeININame", Name + ".ini");
+            string customIniPath = section.GetStringValue("CustomIniPath", string.Empty);
+            string mapCodeIniName = section.GetStringValue("MapCodeININame", Name + ".ini");
+            mapCodeIniPath = string.IsNullOrWhiteSpace(customIniPath)
```
SadPencil, line 94: Wait. We have Custom**Ini**Path and MapCode**INI**Name at the same time?

SadPencil: I suggest: Use `Ini`. Read "MapCodeIniName" first and fallback to "MapCodeININame"

11EJDE11: Yep - both are in use.
Now we have a third "MapCodeIniName" in [f4f015d](https://github.com/CnCNet/xna-cncnet-client/pull/986/commits/f4f015d0a824d2b4df925a23d0a3684e9e43c934).

DXMainClient/Domain/Multiplayer/GameMode.cs
```diff
@@ -89,7 +89,11 @@ public void Initialize()
             MaxPlayersOverride = section.GetIntValueOrNull("MaxPlayersOverride");
 
             forcedOptionsSection = section.GetStringValue("ForcedOptions", string.Empty);
+            string customIniPath = section.GetStringValue("CustomIniPath", string.Empty);
             mapCodeININame = section.GetStringValue("MapCodeIniName", section.GetStringValue("MapCodeININame", Name + ".ini"));
```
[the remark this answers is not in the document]
SadPencil: oops. my mistake

SadPencil:
Do we really need this "CustomIniPath" value while we can already override the ini name via "MapCodeININame"? @11EJDE11

11EJDE11:
> Do we really need this "CustomIniPath" value while we can already override the ini name via "MapCodeININame"? @11EJDE11

Yeah, good call. I had read it wrong and it should have been a YR package change as that is using the wrong values. PR to revert: #1012

Issue #790, opened 2025-07-15
Add option to disable client hotkeys


MahBoiDeveloper:
### Description

For example this option should allow users disable client's main menu hotkeys.

SadPencil:
https://github.com/CnCNet/xna-cncnet-client/pull/988

PR #988, opened 2026-04-22
Add DisableMainMenuHotkeys setting with default true

TheBirdGang:
Resolves issue #790.

### DIFF
--- a/DXMainClient/DXGUI/Generic/OptionPanels/CnCNetOptionsPanel.cs
+++ b/DXMainClient/DXGUI/Generic/OptionPanels/CnCNetOptionsPanel.cs
@@ -96,7 +96,7 @@ namespace DTAClient.DXGUI.Generic.OptionPanels
             chkDisablePrivateMessagePopup.Name = nameof(chkDisablePrivateMessagePopup);
             chkDisablePrivateMessagePopup.ClientRectangle = new Rectangle(
                 chkPingUnofficialTunnels.X,
-                chkNotifyOnUserListChange.Bottom + 12, 0, 0);
+                chkNotifyOnUserListChange.Bottom + 8, 0, 0);
             chkDisablePrivateMessagePopup.Text = "Disable popups from private messages".L10N("Client:DTAConfig:DisablePMPopup");
 
             AddChild(chkDisablePrivateMessagePopup);
--- a/DXMainClient/DXGUI/Generic/OptionPanels/CnCNetOptionsPanel.cs
+++ b/DXMainClient/DXGUI/Generic/OptionPanels/CnCNetOptionsPanel.cs
@@ -34,7 +34,6 @@ namespace DTAClient.DXGUI.Generic.OptionPanels
         XNAClientCheckBox chkSteamIntegration;
         XNAClientCheckBox chkAllowGameInvitesFromFriendsOnly;
         XNAClientCheckBox chkDisablePrivateMessagePopup;
-
         XNAClientCheckBox chkDisableMainMenuHotkeys;
 
         XNAClientDropDown ddAllowPrivateMessagesFrom;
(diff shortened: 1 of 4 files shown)


DXMainClient/DXGUI/Generic/OptionPanels/CnCNetOptionsPanel.cs
```diff
@@ -35,6 +35,8 @@ public CnCNetOptionsPanel(WindowManager windowManager, UserINISettings iniSettin
         XNAClientCheckBox chkAllowGameInvitesFromFriendsOnly;
         XNAClientCheckBox chkDisablePrivateMessagePopup;
 
+        XNAClientCheckBox chkDisableMainMenuHotkeys;
```
11EJDE11: Remove the blank line before this line so it falls in with the other XNAClientCheckBox's

DXMainClient/DXGUI/Generic/OptionPanels/CnCNetOptionsPanel.cs
```diff
@@ -94,12 +96,21 @@ private void InitOptions()
...
             chkDisablePrivateMessagePopup.Text = "Disable Popups from Private Messages".L10N("Client:DTAConfig:DisablePMPopup");
 
             AddChild(chkDisablePrivateMessagePopup);
 
+            chkDisableMainMenuHotkeys = new XNAClientCheckBox(WindowManager);
+            chkDisableMainMenuHotkeys.Name = nameof(chkDisableMainMenuHotkeys);
+            chkDisableMainMenuHotkeys.ClientRectangle = new Rectangle(
+                chkPingUnofficialTunnels.X,
+                chkDisablePrivateMessagePopup.Bottom + 12, 0, 0);
+            chkDisableMainMenuHotkeys.Text = "Disable Main Menu Hotkeys".L10N("Client:DTAConfig:DisableMainMenuHotkeys");
```
11EJDE11: I don't like the capitalisation here. "Disable main menu hotkeys" would be my preference. And also change the private message one to "Disable popups from private messages".

TheBirdGang:
Updated it, how does it look now? Is this good to merge?

SadPencil:
The code change is small and looks good to me. But can you briefly explain what is main menu hot key? And maybe we can provide a tooltip explaining it to players too

Metadorius:
> The code change is small and looks good to me. But can you briefly explain what is main menu hot key? And maybe we can provide a tooltip explaining it to players too

You can hit some keys and open some menus like LAN or Skirmish. Most doesn't know about that and gets confused, AND they also don't properly get disabled in submenus IIRC, I yeeted myself out of the lobby multiple times by accident when trying to type something and not focusing the field, or sth like that.

SadPencil:
> > The code change is small and looks good to me. But can you briefly explain what is main menu hot key? And maybe we can provide a tooltip explaining it to players too
> 
> You can hit some keys and open some menus like LAN or Skirmish. Most doesn't know about that and gets confused, AND they also don't properly get disabled in submenus IIRC, I yeeted myself out of the lobby multiple times by accident when trying to type something and not focusing the field, or sth like that.

Huh? So it's a client hot key instead of an in-game hotkey? Well we must have a tooltip to explain this feature.

TheBirdGang:
Yeah, I honestly didn't know these hotkeys existed when I found this issue. Maybe this could be conveyed by having the hotkey listed next to the main menu buttons in parentheses when they are enabled? I wanted to see if I could do that originally when I took this on, but I don't know where the main menu button text is located.

SadPencil:
btw,



why did we put one sound option in audio tab while the other sound option in cncnet tab?

in my opinion we can move the "Play sound when a game is hosted" into Audio tab, and let the new "Disable main menu hotkeys" option reuse the location of "Play sound when a game is hosted" in CnCNet tab

SadPencil:
图片
图片

SadPencil:
图片
图片

SadPencil:
图片

SadPencil (approved):

PR #991, opened 2026-04-23
Launch RA executable directly from Campaign button (no client wrapper)

CO2-code:
Before: https://youtu.be/dtjBkAuyp2s


Metadorius:
this is not needed, you can just make a simple button from INI config that launches whatever EXE you want

PR #992, opened 2026-04-25
Fix undefined hotkey configurations being lost on startup and save

Flactine:
This PR fixes a bug where hotkey configuration could be lost when the game starts.

In particular, custom hotkey entries added by other extension platforms, such as the hotkeys provided by [ObjectInfo.dll](https://github.com/handama/ObjectInfo), could disappear because KeyboardMD.ini was rewritten during startup with only the hotkeys known to the launcher.

This PR changes the save behavior so that existing hotkey entries are preserved instead of being unintentionally dropped.


SadPencil:
The client erases unrecognized hotkeys whenever the Save button is clicked. This issue has been existed for 9 years. However, before the smart hotkey PR https://github.com/CnCNet/xna-cncnet-client/pull/964, the client didn't automatically save the hotkey on start, so this issue did not get noticed. Now, since the client automatically handles recommended default hotkeys on start, it is necessary to fix this issue.

SadPencil (approved):

PR #1002, opened 2026-05-05
Various fixups - Clamp skill level, thread-safe map updates, dispose map preview images

11EJDE11:
Couple of fixups that are needed before our next release.

-Fixes an issue where a lobby's skill level being out of bounds could crash the client
-Fixes a race condition when reloading maps while they are open in lobby
-Fix a memory leak with preview images

Don't merge just yet, need to do some more checking on the map stuff tomorrow.

### DIFF
--- a/DXMainClient/DXGUI/Multiplayer/GameInformationPanel.cs
+++ b/DXMainClient/DXGUI/Multiplayer/GameInformationPanel.cs
@@ -15,7 +15,6 @@ using Microsoft.Xna.Framework.Graphics;
 
 using Rampastring.XNAUI;
 using Rampastring.XNAUI.XNAControls;
-using Image = SixLabors.ImageSharp.Image;
 
 namespace DTAClient.DXGUI.Multiplayer
 {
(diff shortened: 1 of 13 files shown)


DXMainClient/DXGUI/Multiplayer/GameInformationPanel.cs
```diff
@@ -255,9 +256,16 @@ public void SetInfo(GenericHostedGame game)
...
+                    ? mapLoader.GetCachedPreviewImageFromMap(map, syncLoadOnCacheMiss: false)
+                    : null;
 
-                if (mapPreviewImage != null)
+                if (map?.IsImmediatePreviewImageAvailable() ?? false)
+                {
+                    mapPreviewTexture = AssetLoader.LoadTextureUncached(map.PreviewPath);
+                    mapPreviewTextureNeedsDispose = true;
+                }
+                else if (mapPreviewImage != null)
```
SadPencil: Can you comment why manually calling `LoadTextureUncached` directly from file system instead of getting the image object? performance?

11EJDE11: Yep. LoadTextureUncached uses Texture2D.FromStream so PNG decodes straight to a GPU texture - much faster than the alternative (GetCachedPreviewImageFromMap > Image.Load > TextureFromImage > decodes to Image<Rgba31> > allocates Color[imageSize] > PreMultiply > SetData.

DXMainClient/Domain/Multiplayer/MapPreviewCacheManager.cs
```diff
@@ -22,4 +22,9 @@ public MapPreviewCacheManager(int capacity) : base(capacity) { }
         Image? image = map.GetNonImmediatePreviewImage();
         return image;
     }
+
+    protected override void OnOutputRemoved(Image? output)
+    {
+        output?.Dispose();
```
SadPencil, line 28: This method adds a strong assumption that the caller must not preserve the Image object, which is the main reason I didn't call Dispose when making this cache manager. If you want the manager to dispose the Image, you must repeat this assumption in the Request method's XML comment and almost every occurrence of the method

SadPencil (changes requested):

DXMainClient/Domain/Multiplayer/CacheManagerBase.cs
```diff
@@ -186,6 +190,10 @@ public void Clear()
     /// <returns>The output.</returns>
     protected abstract TOutput? ComputeOutputForInput(TInput input);
 
+    protected virtual void OnOutputRemoved(TOutput? output)
```
SadPencil, line 193: This method adds a strong assumption that the caller must not preserve the `TOutput` object, which is the main reason I didn't call Dispose when making this cache manager. If you want the manager to dispose the `TOutput` object, you must repeat this assumption in the Request method's XML comment and almost every occurrence of the method

SadPencil (changes requested):

DXMainClient/Domain/Multiplayer/CacheManagerBase.cs
11EJDE11: [97475eb](https://github.com/CnCNet/xna-cncnet-client/pull/1002/commits/97475eb6ff457c6c31e2d893fd3a7b9ea7c8674d)

DXMainClient/Domain/Multiplayer/MapPreviewCacheManager.cs
11EJDE11: [97475eb](https://github.com/CnCNet/xna-cncnet-client/pull/1002/commits/97475eb6ff457c6c31e2d893fd3a7b9ea7c8674d)

DXMainClient/Domain/Multiplayer/CacheManagerBase.cs
SadPencil: > [97475eb](https://github.com/CnCNet/xna-cncnet-client/pull/1002/commits/97475eb6ff457c6c31e2d893fd3a7b9ea7c8674d)

This commit does not modify CacheManagerBase.cs

SadPencil:
I don't agree with the [97febbf](https://github.com/CnCNet/xna-cncnet-client/pull/1002/commits/97febbf7f1cfcfc056b4d91eea2a5f554f181969) commit. Can we discuss the borrow-check pattern in discord?

Issue #653, opened 2025-02-03
Allowing `networkdefinitions.local.ini` file to override `networkdefinitions.ini` file


SadPencil:
Some mods are distributing `networkdefinitions.ini` file https://github.com/CnCNet/cncnet-yr-client-package/pull/407 in response to the removal of GameSurge IPs #489. This file will be overriden during an update. Therefore, if a user with a poor Internet condition want to override the default network definitions that comes with a mod, he/she should copy `networkdefinitions.ini` file as `networkdefinitions.local.ini` file, the latter of which should be guaranteed not being overwritten by an update by convention.

Codes should be added to the client, when about reading `networkdefinitions.ini` file, find `networkdefinitions.local.ini` file first.

SadPencil:
https://github.com/CnCNet/xna-cncnet-client/pull/661

PR #1003, opened 2026-05-06
Add "Use legacy fonts" option to allow users to revert to SpriteFonts

11EJDE11:
Adds a checkbox to the DisplayOptionsPanel to enable legacy fonts.

Requires https://github.com/CnCNet/Rampastring.XNAUI/pull/21

I think we should offer this to users, at least initially. Currently the only way to go back to SpriteFonts would be to change your Fonts.ini which would get overwritten later with updates. This way, it is nice and easy for the user and will save a few headaches when we inevitably get some complaints.


SadPencil:
btw Should we offer a Fonts.local.ini as well?

11EJDE11:
How do you mean?

SadPencil:
> How do you mean?

i mean like https://github.com/CnCNet/xna-cncnet-client/issues/653

PR #1022, opened 2026-05-22
Add [RecommendedOptions] section for maps and game modes

11EJDE11:
Maps and game modes can now suggest default option values that apply when the map is selected, without locking the controls.

[RecommendedOptions]
chkCrates=true
chkMultiEng=true

Works identically to [ForcedOptions] syntax but leaves checkboxes and dropdowns editable. Values persist when switching to a map that has no recommendation for that option - the last-set value is kept.

Supports both official maps (via MPMaps.ini) and custom maps.

(revival of #643 with the RecommendedOptions as suggested by kerbiter.)

### DIFF
--- a/DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
+++ b/DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
@@ -2482,6 +2482,12 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
             foreach (var dd in dropDownListClone)
                 dd.SelectedIndex = dd.HostSelectedIndex;
 
+            ApplyRecommendedCheckBoxOptions(GameMode.RecommendedCheckBoxValues);
+            ApplyRecommendedCheckBoxOptions(Map.RecommendedCheckBoxValues);
+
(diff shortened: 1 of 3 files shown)


Metadorius:
> Works identically to [ForcedOptions] syntax but leaves checkboxes and dropdowns editable. Values persist when switching to a map that has no recommendation for that option - the last-set value is kept.

I think this behavior will be confusing to players. Imagine tweaking settings, changing the map to something else and back, and now your settings are screwed.

Also have you thought of any way to do some indication that this setting was changed? With forced options it's obvious as they are blocked, but not here.

CnCRAZER:
> > Works identically to [ForcedOptions] syntax but leaves checkboxes and dropdowns editable. Values persist when switching to a map that has no recommendation for that option - the last-set value is kept.
> 
> I think this behavior will be confusing to players. Imagine tweaking settings, changing the map to something else and back, and now your settings are screwed.
> 
> Also have you thought of any way to do some indication that this setting was changed? With forced options it's obvious as they are blocked, but not here.

I share the same thoughts regarding this. As a map maker it would be cool to suggest settings instead of forcing them but I can already imagine the amount of posts or comments saying their options keep changing and it being a nuisance to them

PR #1027, opened 2026-06-05
Fix issues with unicode paths and Windows 7 DLL loading

11EJDE11:
Fix an issue with loading HarfBuzz when path contains unicode characters.
Fix an issue with loading DLLs on unpatched Windows 7 machines.

### DIFF
--- a/DXMainClient/Program.cs
+++ b/DXMainClient/Program.cs
@@ -58,12 +58,11 @@ namespace DTAClient
 #endif
 
 #if NETFRAMEWORK
-            // Native libs (e.g. libHarfBuzzSharp.dll from HarfBuzzSharp.NativeAssets.Win32)
-            // ship under either SPECIFIC_LIBRARY_PATH/{x64|x86|arm64}/
-            // or COMMON_LIBRARY_PATH/{x64|x86|arm64}/. The .NET Framework runtime
-            // does not search those subfolders by default, and HarfBuzzSharp's resolver looks
-            // beside the EXE rather than beside its managed wrapper - so without help, P/Invoke
-            // calls fail with "Unable to load library 'libHarfBuzzSharp'".
+            // Native libs (e.g. SDL2.dll, soft_oal.dll) ship under SPECIFIC_LIBRARY_PATH/{x64|x86|arm64}/
+            // or COMMON_LIBRARY_PATH/{x64|x86|arm64}/. The .NET Framework runtime does not search those
+            // subfolders by default, so register them with the OS DLL search path. When the secure
+            // DLL-directory APIs are missing (older Windows 7 SP1 without KB2533623), fall back to
+            // pre-loading the natives so the client still starts instead of throwing here.
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
                 static bool areSecureDllLoadingAPIsAvailable()
@@ -82,11 +81,6 @@ namespace DTAClient
                     return true;
                 }
 
-                if (!areSecureDllLoadingAPIsAvailable())
-                    throw new PlatformNotSupportedException("This application requires at least Windows 7 SP1 with KB4457144 (alternatively, KB2533623 or KB3063858) installed.");
-
-                SetDefaultDllDirectories(LOAD_LIBRARY_SEARCH_APPLICATION_DIR | LOAD_LIBRARY_SEARCH_USER_DIRS | LOAD_LIBRARY_SEARCH_DEFAULT_DIRS);
-
                 string archSubfolder = RuntimeInformation.ProcessArchitecture switch
                 {
                     Architecture.X64 => "x64",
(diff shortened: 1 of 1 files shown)


SadPencil:
Do we really need to add such a lot of codes to support these machines? Personally I think the current error message is sufficient.

```cs
if (!areSecureDllLoadingAPIsAvailable())
    throw new PlatformNotSupportedException("This application requires at least Windows 7 SP1 with KB4457144 (alternatively, KB2533623 or KB3063858) installed.");
```

DXMainClient/Program.cs
```diff
@@ -140,6 +166,114 @@ static void addDllDirectoryIfExists(string path)
...
+
+        private static bool IsAnsiRepresentable(string text)
+        {
+            if (string.IsNullOrEmpty(text))
+                return true;
+
+            // Encoding.Default is the system ANSI code page (CP_ACP) on .NET Framework - the same
+            // code page HarfBuzzSharp's ANSI LoadLibrary marshals through. If the round-trip is lossy,
+            // the path cannot be represented and the native load would fail.
+            Encoding ansi = Encoding.Default;
```
SadPencil, line 243: Use EncodingExt.ANSI here -- without using EncodingExt class in ClientCore the .net 8 client does not guarantee returning ANSI

Metadorius:
> Do we really need to add such a lot of codes to support these machines?

Are we sure this is the code of kind "once written and never needing to be touched again"?
- "Yes, absolutely" => we could leave it in still
- "Maybe" or "No" => a error asking the user to update is better since we're not sure if this would need to be maintained, users better install some updates

DXMainClient/Program.cs
```diff
@@ -140,6 +166,114 @@ static void addDllDirectoryIfExists(string path)
...
+        /// relaunching from the directory's all-ASCII 8.3 short path makes the native load succeed
+        /// with no change to how the libraries are deployed. This only ever triggers for the rare user
+        /// whose install path is not ANSI-representable; normal installs return immediately. If an
+        /// ANSI-safe short path is unavailable (8.3 short names disabled on the volume), we simply do
+        /// not relaunch: Rampastring.XNAUI's FontManager detects that the native shaper cannot load
+        /// and falls back to unshaped rendering on its own, so the client still starts.
+        /// </para>
+        /// </summary>
+        [SupportedOSPlatform("windows")]
+        private static void RelaunchFromAnsiSafePathIfNeeded(string[] args)
```
SadPencil, line 187: Personally I would prefer just crash the client if the path is not ANSI safe. There are game executables too and I don't know if they can fully support a non ANSI path

DXMainClient/Program.cs
```diff
@@ -140,6 +166,114 @@ static void addDllDirectoryIfExists(string path)
...
+            // Encoding.Default is the system ANSI code page (CP_ACP) on .NET Framework - the same
+            // code page HarfBuzzSharp's ANSI LoadLibrary marshals through. If the round-trip is lossy,
+            // the path cannot be represented and the native load would fail.
+            Encoding ansi = Encoding.Default;
+            return ansi.GetString(ansi.GetBytes(text)) == text;
+        }
+
+        private static string GetShortPath(string longPath)
+        {
+            var buffer = new StringBuilder(longPath.Length + 16);
```
SadPencil, line 249: what does number 16 refer to here

DXMainClient/Program.cs
Metadorius: I disagree with a crash. a warning is fine. but we need to make a small framework piece of code for warnings/messages that can be dismissed

SadPencil: > I disagree with a crash. a warning is fine. but we need to make a small framework piece of code for warnings/messages that can be dismissed

by crashing it we can immediately dropping all the diff starting from line 169 to line 322. And, non ANSI paths are not that common. It means the folder name contains a language that the user is not familiar with (e.g., a Chinese character is considered as ANSI in a Chinese user's machine but not considered as ANSI in an English user's machine), or contains emojis.

SadPencil: Also, look at the code. The code uses 8.3 short names to support such a non ANSI path. So we use a DOS-era technique to support a modern emoji name. Well, this is kind of weird to me.

Metadorius: > And, non ANSI paths are not that common

wrong. some people like to use desktop folder and give the user account non-ANSI names

Metadorius: > e.g., a Chinese character is considered as ANSI in a Chinese user's machine but not considered as ANSI in an English user's machine

I don't know why you folks still stick to the outdated encodings, we've abandoned encodings for Cyrillics a long time ago, for example, and all Cyrillics are Unicode nowadays since I think Windows Vista

SadPencil: > > e.g., a Chinese character is considered as ANSI in a Chinese user's machine but not considered as ANSI in an English user's machine
> 
> I don't know why you folks still stick to the outdated encodings, we've abandoned encodings for Cyrillics a long time ago, for example, and all Cyrillics are Unicode nowadays since I think Windows Vista

for a modern support we should fix HarfBuzzSharp instead, to make sure it's Unicode all the way, rather than making a workaround in the client by using an ASCII 8.3 path to support a full Unicode path.

MahBoiDeveloper:
> Fix an issue with loading DLLs on unpatched Windows 7 machines.

Better to please user to update Win7 or switch to the newer version of the OS Windows.

SadPencil (changes requested):
Can we directly fix the issue in HarfBuzzSharp instead of introducing the 8.3 filename workaround?

PR #1032, opened 2026-06-14
Fix GameListBox row height to fit icons and center content

11EJDE11:
Rows in the GameListBox now size to the tallest icon, or the font height if the text is taller. All icons and text are vertically centered within each row.

Screenshot (contains Coolisy's recent font change):
image


SadPencil (approved):
Looks like the current client mixed too much GUI code in DXMainClient project instead of ClientGUI. Whatever. Wait for the MVVM migration to solve this mess.

MahBoiDeveloper (approved):

Metadorius (approved):

PR #1035, opened 2026-06-19
Add dropped file map handler

11EJDE11:
[-] Updates XNAUI
[-] Adds a handler for drag/dropped map files

Now when drag/dropping map files onto the client, it will automatically rename (.yrm > .map for example), move them to the Custom maps folder, then add them to the map listing. Supports dropping multiple files, zip files, as well as games that have two map files (for instance .ini and .bin with TD).

### DIFF
--- a/DXMainClient/DXGUI/GameClass.cs
+++ b/DXMainClient/DXGUI/GameClass.cs
@@ -241,6 +241,9 @@ namespace DTAClient.DXGUI
 
             buildServiceProviderTask.GetAwaiter().GetResult();
 
+            // Enable drag-and-drop import of map files onto the client window.
+            serviceProvider.GetService<DroppedMapHandler>().Initialize();
+
             Logger.Log("Initializing loading screen.");
             LoadingScreen ls = serviceProvider.GetService<LoadingScreen>();
             wm.AddAndInitializeControl(ls);
@@ -275,6 +278,7 @@ namespace DTAClient.DXGUI
                             .AddSingleton<DiscordHandler>()
                             .AddSingleton<PrivateMessageHandler>()
                             .AddSingleton<MapLoader>()
+                            .AddSingleton<DroppedMapHandler>()
                             .AddSingleton<Random>(GetRandom())
                             .AddSingleton<DirectDrawWrapperManager>();
 
--- a/DXMainClient/DXGUI/GameClass.cs
+++ b/DXMainClient/DXGUI/GameClass.cs
@@ -241,9 +241,6 @@ namespace DTAClient.DXGUI
 
             buildServiceProviderTask.GetAwaiter().GetResult();
 
-            // Enable drag-and-drop import of map files onto the client window.
-            serviceProvider.GetService<DroppedMapHandler>().Initialize();
-
             Logger.Log("Initializing loading screen.");
             LoadingScreen ls = serviceProvider.GetService<LoadingScreen>();
             wm.AddAndInitializeControl(ls);
--- a/DXMainClient/Domain/Multiplayer/DroppedMapHandler.cs
+++ b/DXMainClient/Domain/Multiplayer/DroppedMapHandler.cs
@@ -27,13 +27,11 @@ namespace DTAClient.Domain.Multiplayer
     /// </summary>
     public class DroppedMapHandler
     {
-        private const string CUSTOM_MAPS_DIRECTORY = "Maps/Custom";
-
         /// <summary>
         /// File extensions that are treated as map files. These are renamed to the
         /// configured <see cref="ClientConfiguration.MapFileExtension"/> on import
-        /// (e.g. YR's ".yrm"/".mpr" become ".map"). The configured extension is
-        /// always included.
+        /// (e.g. YR's ".yrm" or RA2's ".mpr" become ".map"). The configured extension
+        /// is always included.
         /// </summary>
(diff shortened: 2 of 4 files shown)


DXMainClient/DXGUI/GameClass.cs
```diff
@@ -241,6 +241,9 @@ protected override void Initialize()
 
             buildServiceProviderTask.GetAwaiter().GetResult();
 
+            // Enable drag-and-drop import of map files onto the client window.
+            serviceProvider.GetService<DroppedMapHandler>().Initialize();
```
SadPencil: Do we really need to initialize it before the loading screen?

DXMainClient/Domain/Multiplayer/DroppedMapHandler.cs
```diff
@@ -0,0 +1,340 @@
...
+        /// A single dropped file (loose or a zip entry) that is a candidate for import,
+        /// together with how to write its contents to a destination path.
+        /// </summary>
+        private sealed class DroppedFile
+        {
+            public DroppedFile(string baseName, string extension, bool isMap, Action<string> extractTo)
+            {
+                BaseName = baseName;
+                Extension = extension;
+                IsMap = isMap;
```
SadPencil: it is not clear from the variable name what the file it is if `IsMap is false`

DXMainClient/Domain/Multiplayer/DroppedMapHandler.cs
```diff
@@ -0,0 +1,340 @@
...
+            {
+                Logger.Log($"DroppedMapHandler: Failed to import map {map.BaseName}: {ex.Message}");
+                result.Failed++;
+            }
+            finally
+            {
+                if (tempPath != null)
+                {
+                    try { File.Delete(tempPath); }
+                    catch (Exception ex) { Logger.Log($"DroppedMapHandler: Failed to clean up temp file {tempPath}: {ex.Message}"); }
```
SadPencil: write "temporary" instead of "temp" since it's more formal?

DXMainClient/Domain/Multiplayer/DroppedMapHandler.cs
```diff
@@ -0,0 +1,340 @@
...
+                            IsMapFileExtension(extension),
+                            destination => capturedEntry.ExtractToFile(destination, overwrite: false)));
+                    }
+                }
+
+                ImportGroups(files, importedThisDrop, result);
+            }
+            catch (Exception ex)
+            {
+                Logger.Log($"DroppedMapHandler: Failed to import zip {zipPath}: {ex.Message}");
```
SadPencil: write "archive" instead of "zip" since it's more formal?

DXMainClient/Domain/Multiplayer/DroppedMapHandler.cs
```diff
@@ -0,0 +1,340 @@
...
+    /// group is imported under a single shared, unique base name to keep them linked.
+    /// </summary>
+    public class DroppedMapHandler
+    {
+        private const string CUSTOM_MAPS_DIRECTORY = "Maps/Custom";
+
+        /// <summary>
+        /// File extensions that are treated as map files. These are renamed to the
+        /// configured <see cref="ClientConfiguration.MapFileExtension"/> on import
+        /// (e.g. YR's ".yrm"/".mpr" become ".map"). The configured extension is
```
SadPencil: IIRC ".mpr" is not YR's but the original RA2's multiplayer map extension

DXMainClient/Domain/Multiplayer/DroppedMapHandler.cs
```diff
@@ -0,0 +1,340 @@
...
+    ///
+    /// A map may consist of more than one file: a primary map file plus one or more
+    /// supplemental files (see <see cref="ClientConfiguration.SupplementalMapFileExtensions"/>).
+    /// These are associated purely by sharing a base file name (e.g. "mymap.map" +
+    /// "mymap.bin"), so dropped files are grouped by base name and every file in a
+    /// group is imported under a single shared, unique base name to keep them linked.
+    /// </summary>
+    public class DroppedMapHandler
+    {
+        private const string CUSTOM_MAPS_DIRECTORY = "Maps/Custom";
```
SadPencil: Is there already a CUSTOM_MAPS_DIRECTORY constant? if so maybe we can assign this constant by using that constant as value

DXMainClient/Domain/Multiplayer/DroppedMapHandler.cs
SadPencil: @11EJDE11 I mean "temp" in the string. "temp" is fine as a code name.

> Failed to clean up **temp** file
>
> Failed to clean up **temporary** file

DXMainClient/Domain/Multiplayer/DroppedMapHandler.cs
```diff
@@ -323,7 +321,7 @@ public DroppedFile(string baseName, string extension, bool isMap, Action<string>
             /// <summary>The original file extension, without a leading dot.</summary>
             public string Extension { get; }
 
-            public bool IsMap { get; }
+            public bool IsPrimaryMapFile { get; }
```
SadPencil, line 324: Did you mean if `IsPrimaryMapFile is false` the file is a supplemental file? Better to explain it in XML doc since it can't be directly inferred from the variable name

PR #1052, opened 2026-07-29
Add DisableModifierKeys INI option for keyboard commands

### LINKED ISSUE #1051: Allow disabling modifier keys for some keyboard commands
### Description

RA2/YR does not allow setting a hotkey with modifier combined, Ctrl+V for example, as the hotkey for waypoint mode, while the client allows. Setting such a key results in (1) when holding Ctrl+V to enter the waypoint mode, the force-attack cursor may be activated (2) when releasing Ctrl+V, there is a chance that the game is still in the waypoint mode.

The client should be aware of such a limitation and support this restriction by adding a new ini key `DisableModifierKeys` in `KeyboardCommands.ini`:
```ini
[PlanningMode]
UIName=Waypoint Mode
Category=Interface
Description=Enable waypoint mode.
DefaultKey=90
DisableModifierKeys=true
```

When this option is set, only single key can be set for that keyboard command.

SadPencil:
## Feature Pull Request

**Note for whoever merges it: delete all squash messages, including the co-author field! This modification is NOT done by claude, but DeepSeek v4 pro. Therefore, claude should not take this credit.**

> [!WARNING]
> A linked feature issue is required. Pull requests that do not link an existing feature issue—or link an issue that has not been approved by the maintainers—may be closed.

### Related Issue

Closes https://github.com/CnCNet/xna-cncnet-client/issues/1051

### What Changed

Added a new ini key `DisableModifierKeys` in `KeyboardCommands.ini`. 

Example in CnCNet YR:
```ini
[PlanningMode]
UIName=Waypoint Mode
Category=Interface
Description=Enable waypoint mode.
DefaultKey=90
DisableModifierKeys=true
```

When this option is set, only single key can be set for that keyboard command.

### Breaking Changes

- [ ] This pull request introduces a breaking change
- [X] This pull request does not introduce a breaking change

If checked, describe the breaking change and migration impact:

No breaking change.

### Documentation

- [X] Documentation update is needed and has been included
- [ ] Documentation update is not needed

### Checklist

- [X] I linked the corresponding feature issue above
- [X] This pull request is scoped to one feature only
- [x] I verified the implementation by running the client

### DIFF
--- a/ClientGUI/HotkeyConfigurationWindow.cs
+++ b/ClientGUI/HotkeyConfigurationWindow.cs
@@ -336,6 +336,9 @@ namespace ClientGUI
                 if (tsHotkey.HasValue)
                 {
                     Hotkey hotkey = new(tsHotkey.Value);
+                    if (command.DisableModifierKeys && hotkey.Modifier != KeyModifiers.None)
+                        hotkey = new Hotkey(hotkey.Key, KeyModifiers.None);
+
                     bool isDuplicate = false;
                     if (hotkey != Hotkey.None)
                         isDuplicate = !assignedHotkeys.Add(hotkey);
@@ -439,6 +442,8 @@ namespace ClientGUI
             }
 
             var command = (GameCommand)lbHotkeys.GetItem(0, lbHotkeys.SelectedIndex).Tag;
+            if (command.DisableModifierKeys && pendingHotkey.Modifier != KeyModifiers.None)
+                pendingHotkey = new Hotkey(pendingHotkey.Key, KeyModifiers.None);
             command.Hotkey = pendingHotkey;
             RefreshHotkeyList();
             pendingHotkey = Hotkey.None;
@@ -465,13 +470,6 @@ namespace ClientGUI
             }
 
             var currentModifiers = GetCurrentModifiers();
-
-            if (lbHotkeys.SelectedIndex >= 0 && lbHotkeys.SelectedIndex < lbHotkeys.ItemCount)
-            {
-                var selectedCommand = (GameCommand)lbHotkeys.GetItem(0, lbHotkeys.SelectedIndex).Tag;
-                if (selectedCommand.DisableModifierKeys)
-                    currentModifiers = KeyModifiers.None;
-            }
 
             // The XNA keys seem to match the Windows virtual keycodes! This saves us some work
             pendingHotkey = new Hotkey(e.PressedKey, currentModifiers);
@@ -649,10 +647,13 @@ namespace ClientGUI
                 Description = iniSection.GetStringValue("Description", "Unknown description")
                     .L10N($"INI:Hotkeys:{ININame}:Description");
 
+                DisableModifierKeys = iniSection.GetBooleanValue("DisableModifierKeys", false);
+
                 int? defaultTSKey = iniSection.GetIntValueOrNull("DefaultKey");
                 DefaultHotkey = defaultTSKey.HasValue ? new Hotkey(defaultTSKey.Value) : null;
 
-                DisableModifierKeys = iniSection.GetBooleanValue("DisableModifierKeys", false);
+                if (DefaultHotkey != null && DisableModifierKeys && DefaultHotkey.Modifier != KeyModifiers.None)
+                    DefaultHotkey = new Hotkey(DefaultHotkey.Key, KeyModifiers.None);
 
                 // Note: currently, we treat Hotkey.None as null for default hotkeys, since it doesn't make much sense to have a default hotkey that is explicitly "no hotkey" -- Hotkey.None prevents automatically setting a new hot key via DefaultHotkey from a future update
                 if (DefaultHotkey == Hotkey.None)
--- a/ClientGUI/HotkeyConfigurationWindow.cs
+++ b/ClientGUI/HotkeyConfigurationWindow.cs
@@ -252,7 +252,18 @@ namespace ClientGUI
                     bool isDuplicate = !defaultHotkeys.Add(gameCommand.DefaultHotkey);
 
                     if (isDuplicate)
-                        throw new Exception("The default hotkey " + gameCommand.DefaultHotkey.ToString() + " for command " + gameCommand.UIName + " is duplicated with another command's default hotkey. Please make sure all default hotkeys in " + KEYBOARD_COMMANDS_INI + " are unique.");
+                        throw new Exception(
+                            string.Format
+                            (
+                                (
+                                    "The default hotkey {0} for command {1} is duplicated with another command's default hotkey." +
+                                    "Please make sure all default hotkeys in {2} are unique."
+                                ).L10N("Client:Config:ExceptionDuplicatHotkeys"),
+                                gameCommand.DefaultHotkey.ToString(),
+                                gameCommand.UIName,
+                                KEYBOARD_COMMANDS_INI
+                            )
+                        );
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


Coronia:
The fix works properly, Planning mode hotkey can no longer be set to Ctrl+Z with DisableModifierKeys enabled now

11EJDE11:
Haven't yet tested, but looks like if a user already has Ctrl+Z for planning mode, then we ship a build with DisableModifierKeys=true, it will change to Z. But the user may already have Z pointing to something. The duplicate check then runs and removes the first one it finds (may not be the one that had ctrl+Z). Doubt that would affect many so not a biggie.

SadPencil:
> Haven't yet tested, but looks like if a user already has Ctrl+Z for planning mode, then we ship a build with DisableModifierKeys=true, it will change to Z. But the user may already have Z pointing to something. The duplicate check then runs and removes the first one it finds (may not be the one that had ctrl+Z). Doubt that would affect many so not a biggie.

Well... there should be only a few people doing so so I prefer not handling such a case, to avoid introducing more codes

ClientGUI/HotkeyConfigurationWindow.cs
```diff
@@ -632,9 +638,20 @@ public GameCommand(IniSection iniSection)
...
                 int? defaultTSKey = iniSection.GetIntValueOrNull("DefaultKey");
                 DefaultHotkey = defaultTSKey.HasValue ? new Hotkey(defaultTSKey.Value) : null;
 
+                if (DefaultHotkey != null && DefaultHotkey != Hotkey.None
+                    && DisableModifierKeys && DefaultHotkey.Modifier != KeyModifiers.None)
+                {
+                    throw new Exception("The default hotkey " + DefaultHotkey.ToString()
+                        + " for command " + UIName + " has modifier keys but DisableModifierKeys is set to true."
+                        + " Please remove the modifier from the default hotkey or disable DisableModifierKeys in "
+                        + KEYBOARD_COMMANDS_INI + ".");
```
MahBoiDeveloper: How about to make string with format and then .L10N?

Docs/INISystem.md
```diff
@@ -749,6 +749,43 @@ The tag name in `ButtonTag_{TagName}` is matched against mission `Tags` values.
...
+
+In `KeyboardCommands.ini`:
+```ini
+[CommandName]
+UIName=Display name       ; string,  display name for the command in the hotkey configuration UI.
+Category=CategoryName     ; string,  grouping category used in the hotkey configuration UI dropdown.
+Description=Description   ; string,  description text shown in the hotkey configuration UI.
+DefaultKey=0              ; integer, the default TS-encoded key value (low byte = key code, high byte = modifier flags).
+                          ;          Use 0 for commands with no default hotkey.
+DisableModifierKeys=false ; boolean, whether to prevent modifier keys (Ctrl, Shift, Alt) from being combined with this command.
```
MahBoiDeveloper, line 766: > whether to prevent modifier keys (Ctrl, Shift, Alt) 

What about Left/Right Ctrl, Shift, Alt and also Win, Esc, Backspace, Enter and combination of LeftCtrl+LeftShift+S for example?

ClientGUI/HotkeyConfigurationWindow.cs
SadPencil: this is a message for modders only, so maybe not necessary?

Docs/INISystem.md
SadPencil: > > whether to prevent modifier keys (Ctrl, Shift, Alt) 
> 
> What about Left/Right Ctrl, Shift, Alt and also Win, Esc, Backspace, Enter and combination of LeftCtrl+LeftShift+S for example?

The client (and the game) only supports detecting Ctrl, Shift, and Alt. Windows key is not handled.

Esc, Backspace, Enter -- these are single keys and therefore does not require a special treat like this.

If you have found another key that (1) the game forbids, (2) the client allows, and (3) setting such a forbidden key causes in-game issues, feel free to report.

ClientGUI/HotkeyConfigurationWindow.cs
MahBoiDeveloper: > this is a message for modders only, so maybe not necessary?

Some of the modders messages are [translated](https://github.com/CnCNet/xna-cncnet-client/blob/1aa467822626ef27ded89b96ac7bbe6fc3ff591e/DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs#L276)

SadPencil: > > this is a message for modders only, so maybe not necessary?
> 
> Some of the modders messages are [translated](https://github.com/CnCNet/xna-cncnet-client/blob/1aa467822626ef27ded89b96ac7bbe6fc3ff591e/DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs#L276)

Can you help me adding L10N calls on the whole HotkeyConfigurationWindow.cs file? and submit directly to this PR

MahBoiDeveloper:
image

Doesn't work

SadPencil:
@MahBoiDeveloper works. Shift+Ctrl+A becomes A

图片

MahBoiDeveloper:
> @MahBoiDeveloper works. Shift+Ctrl+A becomes A

Sorry, my bad. I've used wrong config. Yes. PR feature works.

ClientGUI/HotkeyConfigurationWindow.cs
MahBoiDeveloper:
```suggestion
                        throw new Exception(
                            string.Format((
                                "The default hotkey {0} for command {1} is duplicated with another command's default hotkey. " +
                                "Please make sure all default hotkeys in {2} are unique.").L10N("Client:DTAConfig:KeyHasModifierWithEnabledDisableModifierKeys"),
                            gameCommand.DefaultHotkey.ToString(),
                            gameCommand.UIName,
                            KEYBOARD_COMMANDS_INI));
```

MahBoiDeveloper (changes requested):

ClientGUI/HotkeyConfigurationWindow.cs
SadPencil: The HotkeyConfigurationWindow.cs file already contains a string for modders that DOES NOT have L10N().

If we only introduce L10N for this sentence only, it introduces inconsistency. This is why I asked you

> Can you help me adding L10N calls on the whole HotkeyConfigurationWindow.cs file? and submit directly to this PR

before.

Besides, please do not introduce placeholder `.L10N("")` when using suggestion feature. Suggestion feature can be used only if one can directly press "Apply suggestion" button to merge your change, not if the suggestion itself must be edited.

SadPencil: @MahBoiDeveloper 
Received your `"KeyHasModifierWithEnabledDisableModifierKeys"`. How about the other L10N call?

MahBoiDeveloper: > @MahBoiDeveloper Received your `"KeyHasModifierWithEnabledDisableModifierKeys"`. How about the other L10N call?

@SadPencil check d686e6b77571f903b06e2b2553419124105d526f.

SadPencil: thanks. a space character after the dot `.` is missing. Updated

SadPencil: also, copilot said we never used `Client:Config:` before. Update to `Client:DTAConfig:`

SadPencil: also, copilot said the localization key "Client:DTAConfig:ExceptionDuplicatHotkeys" contains a typo ("Duplicat")

SadPencil:
> ## Pull request overview
> 
> Copilot reviewed 3 out of 3 changed files in this pull request and generated no new comments.
> Suppressed comments (2)
> 
> **ClientGUI/HotkeyConfigurationWindow.cs:450**
> 
>     * In the Assign flow, stripping modifiers from pendingHotkey means a user pressing (e.g.) Ctrl+V can unintentionally rebind plain V, and the code will also unbind any other command already using V. For DisableModifierKeys=true commands, modifier-combined input should be rejected (no rebinding) rather than silently converted.
> 
> 
> ```
>             var command = (GameCommand)lbHotkeys.GetItem(0, lbHotkeys.SelectedIndex).Tag;
>             if (command.DisableModifierKeys && pendingHotkey.Modifier != KeyModifiers.None)
>                 pendingHotkey = new Hotkey(pendingHotkey.Key, KeyModifiers.None);
> ```
> 
> **ClientGUI/HotkeyConfigurationWindow.cs:354**
> 
>     * For commands with DisableModifierKeys=true, silently stripping modifiers during INI load can steal an existing single-key binding from another command (depending on section iteration order) and then clear the other command as a duplicate. Safer behavior is to treat modifier-combined values as invalid for these commands: remove the key from the INI and leave the command unbound so defaults (or the user's next assignment) can apply without creating cross-command conflicts.
> 
> 
> This issue also appears on line 447 of the same file.
> 
> ```
>                 if (tsHotkey.HasValue)
>                 {
>                     Hotkey hotkey = new(tsHotkey.Value);
>                     if (command.DisableModifierKeys && hotkey.Modifier != KeyModifiers.None)
>                         hotkey = new Hotkey(hotkey.Key, KeyModifiers.None);
> ```

You are right. Should just assign the key as none instead

SadPencil:
> Haven't yet tested, but looks like if a user already has Ctrl+Z for planning mode, then we ship a build with DisableModifierKeys=true, it will change to Z. But the user may already have Z pointing to something. The duplicate check then runs and removes the first one it finds (may not be the one that had ctrl+Z). Doubt that would affect many so not a biggie.

Updated. If a user specified, e.g., `Ctrl+Z`, the client will not recognize it as `Z` but none now.

PR #1070, opened 2026-09-04
Fix LAN host listener startup race

wcfj78:
## Summary

Fixes a race condition when hosting LAN games where the local client can attempt to connect to the LAN lobby TCP listener before the listener has actually started.

## Problem

On Linux using the .NET 8 UniversalGL client, creating a LAN game consistently failed with:

    System.Net.Sockets.SocketException: Connection refused
    127.0.0.1:1233

`SetUp()` starts `ListenForClients()` on a new thread and immediately attempts to connect to `127.0.0.1:1233`. The listener thread is not guaranteed to have called `TcpListener.Start()` before that connection attempt occurs.

## Fix

Added a `ManualResetEventSlim` that is signaled immediately after `listener.Start()` succeeds. The host waits for that signal before attempting the local TCP connection.

A 5-second timeout is included so the client does not wait indefinitely if the listener fails to start.

## Testing

Tested with:

- Ubuntu 26.04.1
- .NET 8 UniversalGL client
- Yuri's Revenge / CnCNet client 2.13.3
- LAN host on Linux
- Remote Windows client

Before the change, port 1233 never reached LISTEN before the connection attempt and hosting consistently failed.

After the change:

- LAN game creation succeeds
- Local client connects successfully
- Remote Windows client joins successfully
- Game launches and plays normally

Firewall was also disabled during troubleshooting to rule out packet filtering as the cause.

### DIFF
--- a/DXMainClient/DXGUI/Multiplayer/GameLobby/LANGameLobby.cs
+++ b/DXMainClient/DXGUI/Multiplayer/GameLobby/LANGameLobby.cs
@@ -107,6 +107,7 @@ namespace DTAClient.DXGUI.Multiplayer.GameLobby
 
         private TcpListener listener;
         private TcpClient client;
+        private readonly ManualResetEventSlim listenerReady = new(false);
         private volatile bool leaving;
         private int sessionId;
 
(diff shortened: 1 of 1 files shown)


SadPencil:
The PR template says:

> Please click the `Preview` tab and select a PR template:
> 
> - Bug fix
> - Feature
> - Miscellaneous
> 
> Failed to execute the instructions may result in your pull request being closed directly.

SadPencil (commented):
Your changes looks good, but please follow our template and report a bug issue first before submitting a PR

Issue #1021, opened 2026-05-21
Clean up debug folder


11EJDE11:
### Description

Currently when a desync happens, we copy the debug.log, syringe log, sync file, and snapshot to the \debug folder, but no cleanup happens. We should only keep the last 10 or limit the amount of space.

11EJDE11:
Turns out we already have this, runs on startup and cleans older than 7 days. We had a user on Discord saying their folder was 500mb+ so perhaps we change to keep 10 files instead? @SadPencil

SadPencil:
> Turns out we already have this, runs on startup and cleans older than 7 days. We had a user on Discord saying their folder was 500mb+ so perhaps we change to keep 10 files instead? @SadPencil 

100 files perhaps? how big is a log file?

11EJDE11:
It varies. 100 at 20mb wouldn't be ideal.

<img width="605" height="872" alt="Image" src="https://github.com/user-attachments/assets/cb448abb-dac6-41e9-9eb7-0e03b4ae0402" />

SadPencil:
I prefer limiting the amount of space

PR #1073, opened 2026-09-07
Add Storage tab to Options window for log and save game rotation

11EJDE11:
Adds a **Storage** tab to the options window, and replaces the single `client_previous.log` backup with proper log rotation.

Today the client keeps one old log: `client.log` is copied over `client_previous.log` on every start. Logs are now renamed to a timestamped backup and pruned to a configured file count and total folder size

(`[ClientLogs]` `MaxKeptLogFiles`, `MaxLogFolderSizeMB`), which addresses #1021.

Retention is read straight from the settings INI rather than through `UserINISettings`, because rotation runs before that is initialised.

image

Note: Screenshots show the replay options - these will be added in a later PR and are opt-in. For now, it's just the client logs.

### DIFF
--- a/ClientCore/SinglePlayerSavedGameManager.cs
+++ b/ClientCore/SinglePlayerSavedGameManager.cs
@@ -1,3 +1,5 @@
+#nullable enable
+
 using System;
 using System.Collections.Generic;
 using System.IO;
--- a/ClientCore/SinglePlayerSavedGameManager.cs
+++ b/ClientCore/SinglePlayerSavedGameManager.cs
@@ -1,0 +1,70 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+using Rampastring.Tools;
+
+namespace ClientCore
+{
+    /// <summary>Applies optional storage limits to single-player saved games.</summary>
+    public static class SinglePlayerSavedGameManager
+    {
+        public static void PruneSavedGames()
+        {
+            PruneSavedGames(SafePath.GetDirectory(ProgramConstants.GamePath, "Saved Games"),
+                UserINISettings.Instance.MaxKeptSavedGames.Value,
+                UserINISettings.Instance.MaxSavedGameFolderSizeMB.Value);
+        }
+
+        internal static void PruneSavedGames(DirectoryInfo directory, int maxKeptSavedGames, int maxFolderSizeMB)
+        {
+            if (maxKeptSavedGames <= 0 && maxFolderSizeMB <= 0)
+                return;
+
+            try
+            {
+                if (!directory.Exists)
+                    return;
+
+                // Match only single-player saves, leaving multiplayer saves and metadata alone.
+                List<FileInfo> saves = directory.EnumerateFiles("*", SearchOption.TopDirectoryOnly)
+                    .Where(file => string.Equals(file.Extension, ".SAV", StringComparison.OrdinalIgnoreCase))
+                    .OrderByDescending(file => file.LastWriteTimeUtc)
+                    .ThenBy(file => file.Name, StringComparer.Ordinal)
+                    .ToList();
+
+                int remainingCount = saves.Count;
+                long totalSize = saves.Sum(file => file.Length);
+                long maxFolderSizeBytes = maxFolderSizeMB * 1024L * 1024L;
+
+                // Always preserve the newest save, even when it alone exceeds the size limit.
+                for (int i = saves.Count - 1; i > 0; i--)
+                {
+                    if ((maxKeptSavedGames <= 0 || remainingCount <= maxKeptSavedGames) &&
+                        (maxFolderSizeMB <= 0 || totalSize <= maxFolderSizeBytes))
+                        break;
+
+                    FileInfo save = saves[i];
+                    try
+                    {
+                        long size = save.Length;
+                        save.Delete();
+                        remainingCount--;
+                        totalSize -= size;
+                        Logger.Log("Deleted old single-player saved game: " + save.Name);
+                    }
+                    catch (Exception ex)
+                    {
+                        // Failed deletions still count towards both limits and can be retried next time.
+                        Logger.Log("Could not delete single-player saved game " + save.Name + ": " + ex.Message);
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                Logger.Log("Could not prune single-player saved games: " + ex.Message);
+            }
+        }
+    }
+}
--- a/DXMainClient/DXGUI/Generic/OptionPanels/StorageOptionsPanel.cs
+++ b/DXMainClient/DXGUI/Generic/OptionPanels/StorageOptionsPanel.cs
@@ -1,0 +1,117 @@
+#nullable enable
+
+using System;
+
+using ClientCore;
+using ClientCore.Extensions;
+
+using ClientGUI;
+
+using Microsoft.Xna.Framework;
+
+using Rampastring.XNAUI;
+using Rampastring.XNAUI.XNAControls;
+
+namespace DTAClient.DXGUI.Generic.OptionPanels;
+
+/// <summary>Configures how much disk space client logs and (if supported) replays are allowed to use.</summary>
+class StorageOptionsPanel : XNAOptionsPanel
+{
+    private const int TEXT_BOX_WIDTH = 70;
+    private const int TEXT_BOX_HEIGHT = 21;
+    private const int TEXT_BOX_X = 170;
+    private const int ROW_SPACING = 30;
+    private const int MAX_KEPT_FILES_LIMIT = 100000;
+    private const int MAX_FOLDER_SIZE_LIMIT_MB = 1024 * 1024;
+
+    public StorageOptionsPanel(WindowManager windowManager, UserINISettings iniSettings)
+        : base(windowManager, iniSettings)
+    {
+    }
+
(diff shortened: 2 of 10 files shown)


DXMainClient/DXGUI/Generic/OptionsWindow.cs
```diff
@@ -85,6 +89,10 @@ public override void Initialize()
                 componentsPanel
             };
 
+            panels.Add(new StorageOptionsPanel(WindowManager, UserINISettings.Instance));
```
SadPencil: Why not just put this item inside the `optionsPanels` initializer?

DXMainClient/PreStartup.cs
```diff
@@ -285,6 +279,91 @@ public static void HandleException(object sender, Exception ex)
             MainClientConstants.DisplayErrorAction("KABOOOOOOOM".L10N("Client:Main:FatalErrorTitle"), error, true);
         }
 
+        private const int DEFAULT_MAX_KEPT_LOG_FILES = 5;
```
SadPencil: I'd prefer at least 20 -- we have maximum 50 MB capacity so no need for a too small number of files restriction

DXMainClient/PreStartup.cs
```diff
@@ -285,6 +279,91 @@ public static void HandleException(object sender, Exception ex)
...
+            try
+            {
+                if (backupFile.Exists)
+                    backupFile.Delete();
+
+                File.Move(clientLogFile.FullName, backupFile.FullName);
+            }
+            catch
+            {
+                return;
```
SadPencil: Are you sure we return here?

DXMainClient/DXGUI/Generic/OptionsWindow.cs
```diff
@@ -41,7 +42,7 @@ public OptionsWindow(WindowManager windowManager, GameCollection gameCollection,
         public override void Initialize()
         {
             Name = "OptionsWindow";
-            ClientRectangle = new Rectangle(0, 0, 576, 475);
+            ClientRectangle = new Rectangle(0, 0, 576 + UIDesignConstants.BUTTON_WIDTH_92, 475);
```
SadPencil, line 45: This is a breaking change. Modders *may* have to adjust the location of existing elements, e.g., elements right aligned or in the 2nd column. Need to edit the documentation for such a migration

SadPencil: Alternatively, maybe we can change it to a button (like the hotkey setting) and open a new window for these storage options. This way it will not be a breaking change anymore

DXMainClient/PreStartup.cs
```diff
@@ -285,6 +279,91 @@ public static void HandleException(object sender, Exception ex)
...
+            {
+                file.Delete();
+            }
+            catch
+            {
+                // Ignored - the file will simply be considered again on the next startup.
+            }
+        }
+
+        private static (int maxKeptLogFiles, long maxFolderSizeBytes) ReadLogRetentionSettings()
```
SadPencil: We use `int ParseLimit(string? text, int previousValue, int maximum)` to process the maximum folder size bytes in options window. Why `long` here?

MahBoiDeveloper:
Log and replays ok. Would you kindly add this feature and for game saves too?

DXMainClient/DXGUI/Generic/OptionsWindow.cs
11EJDE11: Sorry, leftover from testing. Fixed: [473159c](https://github.com/CnCNet/xna-cncnet-client/pull/1073/commits/473159c5fa4475263f276c061145ca3e82fd8505)

DXMainClient/PreStartup.cs
11EJDE11: [32ae73b](https://github.com/CnCNet/xna-cncnet-client/pull/1073/commits/32ae73b5ad2475e13c5c23a00e17a31aecf910da)

DXMainClient/PreStartup.cs
11EJDE11: Just preference I guess. I've removed it. [32ae73b](https://github.com/CnCNet/xna-cncnet-client/pull/1073/commits/32ae73b5ad2475e13c5c23a00e17a31aecf910da)

DXMainClient/DXGUI/Generic/OptionsWindow.cs
11EJDE11: Docs updated [5beb37c](https://github.com/CnCNet/xna-cncnet-client/pull/1073/commits/5beb37cae803d4899385f877d70b1293dd275d4f).
I'm against the hotkey idea personally.

11EJDE11:
> Log and replays ok. Would you kindly add this feature and for game saves too?

I've added that. With 0/unlimited so no saved games get deleted unless the user changes the setting.


Also updated the docs as I missed a bit when changing to 20 instead of 5 files.

ClientCore/SinglePlayerSavedGameManager.cs
```diff
@@ -0,0 +1,70 @@
+using System;
```
SadPencil, line 1: Can you add `#nullable enable` here?

DXMainClient/Startup.cs
SadPencil, line 72: btw: maybe we should also let the storage setting manage this pruning as well. Can share the same setting with the client log.

Update: oops, maybe we can't share the same setting with the client log as the Ares/Phobos logs are huge. Might worth a new setting

ClientCore/SinglePlayerSavedGameManager.cs
11EJDE11: Yes sir [8c74034](https://github.com/CnCNet/xna-cncnet-client/pull/1073/commits/8c74034a2ab086858c5d0b80c2a08ca491af4e9f)

DXMainClient/Startup.cs
11EJDE11: I've added the pruning for them. Slightly bigger commit as I've switched to @Metadorius' scrollpanel [8bb69bc](https://github.com/CnCNet/xna-cncnet-client/pull/1073/commits/8bb69bc10f4a23b11a25354fa48399282e7614d0).

SadPencil (approved):

DXMainClient/Startup.cs
```diff
@@ -105,6 +105,8 @@ public void Execute()
                 }
             }
 
+            SinglePlayerSavedGameManager.PruneSavedGames();
```
SadPencil, line 108: Is it possible to put it in another thread? Just a thought, because I don't know whether doing this will have a negative effect on the load game window or not

Alternatively, put this call in LoadingScreen?

SadPencil:
图片
Do we really need to explain that much to players? I prefer just saying nothing

MahBoiDeveloper:
> Do we really need to explain that much to players? I prefer just saying nothing

Better to keep the description

DXMainClient/Startup.cs
11EJDE11: It's possible but doesn't seem to warrant it in my opinion. I just ran some timings on my potato PC with an HDD and it was <10ms.

11EJDE11:
How about change to `Oldest saves are permanently deleted.`? I'd like to keep "permanently deleted" in there somewhere.

SadPencil:
> How about change to `Oldest saves are permanently deleted.`? I'd like to keep "permanently deleted" in there somewhere.

Sounds good. The original text seems hard to understand (I simulated regular players with a tired mind and found I can't read that sentence)

DXMainClient/DXGUI/Generic/OptionPanels/StorageOptionsPanel.cs
```diff
@@ -0,0 +1,295 @@
...
+        var lblGameLogFolderSizeSuffix = new XNALabel(WindowManager);
+        lblGameLogFolderSizeSuffix.Name = nameof(lblGameLogFolderSizeSuffix);
+        lblGameLogFolderSizeSuffix.Text = "MB  (0 = no limit)".L10N("Client:DTAConfig:StorageMaxSizeSuffix");
+        lblGameLogFolderSizeSuffix.ClientRectangle = new Rectangle(
+            tbMaxGameLogFolderSize.Right + 8, lblGameLogFolderSize.Y, 0, 0);
+
+        var lblGameLogRetentionHint = new XNALabel(WindowManager);
+        lblGameLogRetentionHint.Name = nameof(lblGameLogRetentionHint);
+        lblGameLogRetentionHint.ClientRectangle = new Rectangle(12, lblGameLogFolderSize.Y + ROW_SPACING, 0, 0);
+        lblGameLogRetentionHint.Text = ("The game's debug folder: its logs and the snapshots saved for crashes and desyncs.\n" +
```
SadPencil: @11EJDE11 Can you also shorten this text? This should be my last review comment.

11EJDE11: Good call [81d402f](https://github.com/CnCNet/xna-cncnet-client/pull/1073/commits/81d402fb0510e4acce1776ba55db4ac5325a0176)

SadPencil:
I remember I deleted all squashed commit messages. Why they are retained? Is it a bug of GitHub's "stack" feature?

11EJDE11:
I have no idea. For 1074 and 1075 I can squash each branch to a single commit before merging. Did you have more feedback for those ones, btw?

PR #1074, opened 2026-09-07
Add local-only lobby checkboxes/dropdowns

11EJDE11:
Adds `LocalGameLobbyCheckBox` and `LocalGameLobbyDropDown`: lobby options that write to `spawn.ini` and are remembered per user, but are never broadcast to other players.

A `GameLobbyCheckBox` is part of the shared game options - it is sent in game option messages, only the host can change it. Some options are purely local, such as whether *this* player records a replay. Those need to be settable by non-hosts, must not be broadcast, and should be remembered between sessions.

Any `GameSessionCheckBox` or `GameSessionDropDown` can now persist through a `UserSettingKey`, which stores the value in `[LocalGameOptions]`. That is deliberately independent of `SaveSkirmishGameOptions` / `SaveCampaignGameOptions`, which persist whole lobbies under the control's own name and are loaded after the control initialises - setting both for one control would let the older mechanism overwrite the remembered value.


Metadorius:
What is the use case for such? Currently we do local options through user settings (via literal settings menu), this kind of blurs the line (and honestly calls for a greater restructure I outlined, with shared game options/settings that can be mixed and matched).

SadPencil:
Can you introduce more about the `UserSettingKey`, probably with an example? Also, whether it is possible to automatically determine such a key without letting modders naming it?

11EJDE11:
> What is the use case for such? Currently we do local options through user settings (via literal settings menu), this kind of blurs the line (and honestly calls for a greater restructure I outlined, with shared game options/settings that can be mixed and matched).

Use case is for a "Record replay" checkbox shown in the lobby. It needs to write to spawn.ini and not be broadcasted.
Where had you outlined a restructure? I can take a look.

Docs/INISystem.md
```diff
@@ -399,6 +399,10 @@ EnabledIcon=                               ; string,  texture name for the icon
 DisabledIcon=                              ; string,  texture name for the icon when setting is disabled.
 SortOrder=0                                ; integer, display order for icons in GameInformationPanel and GameListBox. 
                                            ;          Lower values appear first.
+Persistent=false                           ; boolean, remember this checkbox's value per user in the `[LocalGameOptions]`
+                                           ;          section of the user's settings INI, keyed by the checkbox's own
+                                           ;          section name (`SOMEGAMESESSIONCHECKBOX` above). Without it, `Checked`
+                                           ;          applies on every client start.
```
SadPencil, line 405: The comment here seems have different wording than the corresponding ones in LocalGameLobbyCheckBox below. Is it because the functionality in these two checkboxes are different, or is it because AI rewrote these words? Use the same text if the functionality are same

DXMainClient/DXGUI/Multiplayer/GameLobby/LocalGameLobbyCheckBox.cs
```diff
@@ -0,0 +1,32 @@
...
+namespace DTAClient.DXGUI.Multiplayer.GameLobby;
+
+/// <summary>A persisted lobby option that is not broadcast to other players.</summary>
+public class LocalGameLobbyCheckBox : GameSessionCheckBox
+{
+    public LocalGameLobbyCheckBox(WindowManager windowManager) : base(windowManager) { }
+
+    public override void Initialize()
+    {
+        // Register separately from broadcast game options.
```
SadPencil, line 17: I can't get what this comment says

DXMainClient/DXGUI/Multiplayer/GameLobby/LocalGameLobbyDropDown.cs
```diff
@@ -0,0 +1,32 @@
...
+namespace DTAClient.DXGUI.Multiplayer.GameLobby;
+
+/// <summary>A persisted lobby option that is not broadcast to other players.</summary>
+public class LocalGameLobbyDropDown : GameSessionDropDown
+{
+    public LocalGameLobbyDropDown(WindowManager windowManager) : base(windowManager) { }
+
+    public override void Initialize()
+    {
+        // Register separately from broadcast game options.
```
SadPencil, line 17: I can't get what this comment says

MahBoiDeveloper:
> > What is the use case for such? Currently we do local options through user settings (via literal settings menu), this kind of blurs the line (and honestly calls for a greater restructure I outlined, with shared game options/settings that can be mixed and matched).
> 
> Use case is for a "Record replay" checkbox shown in the lobby. It needs to write to spawn.ini and not be broadcasted. Where had you outlined a restructure? I can take a look.

I am curious if can we place such settings to the options menu for example?

Metadorius:
> I am curious if can we place such settings to the options menu for example?

my suggestion would be to implement both a spawn.ini "force on" key for things where we *need* replays as evidence, and also an option to always record replays in ra2md.ini. it both makes sense from a user standpoint (say they don't want to record replays against AI or in campaign, unless in a competition) and doesn't require client changes

DXMainClient/DXGUI/Multiplayer/GameLobby/LocalGameLobbyDropDown.cs
11EJDE11: removed. It was just mentioning that it's separate from the broadcasted values.

Docs/INISystem.md
11EJDE11: Updated in [33871c8](https://github.com/CnCNet/xna-cncnet-client/pull/1074/commits/33871c80d09a20c037a3cc5f47daf065e438ca16)

11EJDE11:
Closing this as I've gone with the settings approach instead. In the Game tab in options, there will be a new "Record replay" checkbox. QM will force recording through the ladder API.

PR #1075, opened 2026-09-07
Add replay recording and playback support

11EJDE11:
Adds replay recording and playback to the client, for packages whose spawner supports it.

A **Replays** tab in the Load Game window lists recorded games, info, and launches playback. Recording is switched on per game from the lobby or the campaign screen. The Storage tab gains replay retention controls.

Replay metadata is read from the `spawn.ini` embedded at the head of the replay file.

Enabled by setting `ReplaySupport=true` in `ClientDefinitions.ini`. Packages that do not opt in see no Replays tab, no replay storage controls, and no behaviour change.

Also writes the file hashes and client version into `spawn.ini` so a replay records what it was produced with.

image


DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
```diff
@@ -1700,6 +1700,7 @@ private PlayerHouseInfo[] WriteSpawnIni(Random pseudoRandom)
             settings.SetStringValue("Scenario", ProgramConstants.SPAWNMAP_INI);
             settings.SetStringValue("UIGameMode", GameMode.UntranslatedUIName);
             settings.SetStringValue("UIMapName", Map.UntranslatedName);
+            settings.SetStringValue("GameClientVersion", ReplayManager.GameClientVersion);
```
SadPencil: I can understand `GameVersion` and `ClientVersion`. What's `GameClientVersion`? A combination of both?

DXMainClient/Domain/ReplayManager.cs
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    public static DirectoryInfo GetDirectory()
+        => SafePath.GetDirectory(ProgramConstants.GamePath, DirectoryName);
+
+    public static FileInfo GetFile(string fileName)
+        => SafePath.GetFile(ProgramConstants.GamePath, DirectoryName, fileName);
+
+    public static string GetRelativePath(string fileName)
+        => SafePath.CombineFilePath(DirectoryName, fileName);
```
SadPencil: Given these method names, they are more like a "shortcut" way that are used internally in this class. I do not recommend exposing them as `public` methods

DXMainClient/Domain/ReplayManager.cs
```diff
@@ -0,0 +1,287 @@
...
+        // Use Windows restrictions on every platform because replay files are shared.
+        foreach (char character in "<>:\"/\\|?*")
+            invalid.Add(character);
+        for (char character = '\0'; character < ' '; character++)
+            invalid.Add(character);
+
+        return invalid;
+    }
+
+    private static string SanitizeForFileName(string name)
```
SadPencil: We have a method that allows getting a safe Win32 name. Do you want to invoke that method (and filter out additional unwanted characters)?

DXMainClient/Domain/ReplayManager.cs
```diff
@@ -0,0 +1,287 @@
...
+        }
+        catch (Exception ex)
+        {
+            Logger.Log("ReplayManager: could not open the replay directory: " + ex.Message);
+        }
+    }
+
+    private static readonly HashSet<char> invalidFileNameChars = BuildInvalidFileNameChars();
+
+    private static HashSet<char> BuildInvalidFileNameChars()
```
SadPencil: We have a method that allows getting a safe Win32 name. Do you want to invoke that method (and filter out additional unwanted characters)?

SadPencil (commented):
A partial review

DXMainClient/Domain/ReplayFileHashes.cs
```diff
@@ -0,0 +1,102 @@
...
+    public static void Write(IniFile spawnIni)
+    {
+        SortedDictionary<string, string> hashes = Collect();
+
+        if (hashes.Count == 0)
+            return;
+
+        var section = new IniSection(SECTION);
+
+        // Paths are values because INI keys cannot safely contain '='.
```
SadPencil: Didn't get it. What does this comment say?

DXMainClient/Domain/ReplayFileHashes.cs
```diff
@@ -0,0 +1,102 @@
...
+            return mismatches;
+
+        SortedDictionary<string, string> local = Collect();
+        var recordedPaths = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+
+        foreach (string key in keys)
+        {
+            string value = replaySpawnIni.GetStringValue(SECTION, key, string.Empty);
+
+            // Use the last separator so paths can contain '|'.
```
SadPencil: Maybe it's better to forbid characters like `|><` otherwise we would have to deal with some Bash/cmd escapes, which is a hell

DXMainClient/Domain/ReplayGame.cs
```diff
@@ -0,0 +1,386 @@
...
+public class ReplayGame
+{
+    public const int MAX_GAME_SPEED_INDEX = 6;
+
+    // 'YRRP' in file order.
+    private const uint REPLAY_MAGIC = 0x50525259;
+
+    // Keep both bounds so future formats can retain support for older replays.
+    private const uint MIN_SUPPORTED_REPLAY_FORMAT_VERSION = 1;
+    private const uint MAX_SUPPORTED_REPLAY_FORMAT_VERSION = 1;
```
SadPencil: Should the client or the modder be responsible for the valid format version?

DXMainClient/Domain/YRReplayGame.cs
```diff
@@ -0,0 +1,386 @@
...
+        return true;
+    }
+
+    private static uint ReadUInt32(byte[] header, int offset)
+        => BinaryPrimitives.ReadUInt32LittleEndian(new ReadOnlySpan<byte>(header, offset, sizeof(uint)));
+
+    private static ulong ReadUInt64(byte[] header, int offset)
+        => BinaryPrimitives.ReadUInt64LittleEndian(new ReadOnlySpan<byte>(header, offset, sizeof(ulong)));
+
+    private static DateTime FromUnixTime(ulong unixTime, DateTime fallback)
```
SadPencil, line 374: I didn't look this method in detail. Just a reminder to double check the time zone stuffs.

DXMainClient/Domain/YRReplayGame.cs
```diff
@@ -0,0 +1,386 @@
...
+            Logger.Log("An error occurred while reading the spawn files of " + FileName + ": " + ex.ToString());
+            spawnIni = string.Empty;
+            spawnMap = Array.Empty<byte>();
+            return false;
+        }
+    }
+
+    /// <summary>Must match SpeedLadder in the spawner's ReplayControls.h.</summary>
+    public static readonly int[] PlaybackSpeedLadder =
+        { 10, 12, 15, 20, 30, 45, 60, 90, 120, 180, 240, 300, 500, 1000, 2000 };
```
SadPencil, line 262: If these codes must match the spawner, whether this class is made for CnCNet YR only? Maybe we can claim an interface and let a YR class implement that interface

DXMainClient/Domain/ReplayManager.cs
SadPencil: The method name is `ToWin32FileName()` in ClientCore project

DXMainClient/DXGUI/Multiplayer/GameLobby/GameLobbyBase.cs
11EJDE11: I've changed it to `GamePackageVersion`.

DXMainClient/Domain/ReplayManager.cs
11EJDE11: Changed to `internal`, thanks.

DXMainClient/Domain/ReplayManager.cs
11EJDE11: Good call, sorted.

DXMainClient/Domain/ReplayManager.cs
11EJDE11: Done.

DXMainClient/Domain/ReplayFileHashes.cs
11EJDE11: `=` are allowed in a path which would mess with the INI parsing.

DXMainClient/Domain/ReplayFileHashes.cs
11EJDE11: Thanks, sorted.

DXMainClient/Domain/ReplayGame.cs
11EJDE11: The client should. I've simplified this to `SUPPORTED_REPLAY_FORMAT_VERSION`

DXMainClient/Domain/YRReplayGame.cs
11EJDE11: Already sorted.

DXMainClient/Domain/YRReplayGame.cs
11EJDE11: I agree and will do that when I add replays to OpenTS if that's OK.

SadPencil: Okay but let's leave a comment on this class file, mentioning what to do next

11EJDE11: Thanks. [51a6eba](https://github.com/CnCNet/xna-cncnet-client/pull/1075/commits/51a6eba93d3956e9d30c7d2adfebc0bc1a28c7de)

DXMainClient/Domain/ReplayManager.cs
SadPencil: I mean, why retaining this method instead of directly calling ToWin32FileName? Is there some special requirements for this feature, beyond a general requirement?

DXMainClient/Domain/ReplayManager.cs
SadPencil: Huh? Codes like `BuildInvalidFileNameChars` are still there.

DXMainClient/Domain/ReplayFileHashes.cs
SadPencil:
```
// paths can contain '|'
```

this comment is then obsolete

DXMainClient/Domain/ReplayFileHashes.cs
```diff
@@ -0,0 +1,110 @@
...
+            {
+                mismatches.Add(string.Format("{0} was not recorded with this replay".L10N("Client:Main:ReplayFileNotRecorded"), relativePath));
+            }
+        }
+
+        return mismatches;
+    }
+
+    /// <summary>Characters excluded from tracked paths because they conflict with the "path|hash" storage format.</summary>
+    private static readonly char[] disallowedPathChars = ['|', '<', '>'];
```
SadPencil: Why do we need `disallowedPathChars` if we can determine by `if (filepath=filepath.ToWin32FileName())`

DXMainClient/Domain/ReplayGame.cs
```diff
@@ -0,0 +1,387 @@
...
+}
+
+/// <summary>
+/// Reads the replay format used by the CnCNet YR spawner.
+/// Listing only reads the header and the embedded spawn.ini; the spawn files
+/// themselves are re-read on demand.
+/// </summary>
+// Written for a CnCNet YR-compatible spawner only. Redo with an interface for the
+// game-specific bits (header layout, speed/FPS tables) when other games are added.
+public class ReplayGame
```
SadPencil: maybe let's name this class as YRReplayGame for now?

DXMainClient/Domain/ReplayManager.cs
SadPencil: ummm... they are called from external classes? Then, maybe we use `public` back and use a longer name, like `GetReplayDirectory`, `GetReplayFile`, `GetReplayFileRelativePath` to prevent a possible misleading with the general utility function

DXMainClient/Domain/ReplayGame.cs
Metadorius: why though

SadPencil: to remind people this class is made for YR. Such a name also implies we can claim an `IReplayGame` interface to extend support for other games

DXMainClient/Domain/ReplayManager.cs
11EJDE11: Sorted.

DXMainClient/Domain/ReplayManager.cs
11EJDE11: Sorted. BuildRecordingPath now calls ToWin32FileName. ToWin32FileName was also missing `"` and control characters so I've added those to it. That also affects MapLoader/MapSharer/Mission, but only for names that couldn't be used anyway.

DXMainClient/Domain/ReplayManager.cs
11EJDE11: Gone, along with the rest of the old filter.

DXMainClient/Domain/ReplayFileHashes.cs
11EJDE11: Removed.

DXMainClient/Domain/ReplayFileHashes.cs
11EJDE11: Replaced.

DXMainClient/Domain/ReplayGame.cs
11EJDE11: Changed to YRReplayGame.

DXMainClient/Domain/ReplayFileHashes.cs
```diff
@@ -0,0 +1,112 @@
...
+
+            int separator = value.LastIndexOf('|');
+            if (separator <= 0)
+                continue;
+
+            string relativePath = FileHashCalculator.NormalizePath(value.Substring(0, separator));
+            string recordedHash = value.Substring(separator + 1);
+            recordedPaths.Add(relativePath);
+
+            // These only go to the log, so they are not translated.
```
SadPencil, line 66: What did this comment say? What "these" and "they" refers to?

DXMainClient/Domain/ReplayFileHashes.cs
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        return mismatches;
+    }
+
+    /// <summary>
+    /// Whether every segment of a tracked path is a valid Win32 file name.
+    /// </summary>
+    private static bool IsStorablePath(string relativePath)
+        => relativePath.Split('/').All(segment => segment == segment.ToWin32FileName());
```
SadPencil, line 92: Becauase this method has a very strong assumption (the input must be a result of `NormalizePath`, as you have assumed `/` as the separator), it is better to further reduce this method's scope -- inside the Collect() method

PR #1077, opened 2026-09-07
Add Tiberian Dawn game type and slightly refactor `ClientType` enum

MahBoiDeveloper:
### What's Changed

Improved `ClientType` enum client usage:

* Refactor getting the steam app id from game type
* Added `TD` option

### Breaking Changes
No

### Documentation

- [x] Documentation update is needed and has been included

### Checklist

- [x] This pull request is not a bug fix or a new feature
- [x] Verifier: I honestly verified the implementation by running the client with this PR. Verifier name: `mah_boi`. The verifier must be a human, not an AI. The verifier can be the same person as the PR proposer.

### DIFF
--- a/ClientCore/Enums/ClientType.cs
+++ b/ClientCore/Enums/ClientType.cs
@@ -2,9 +2,10 @@
 {
     public enum ClientType
     {
+        TD,
+        RA,
         TS,
         YR,
         Ares,
-        RA,
     }
 }
--- a/ClientCore/Enums/ClientTypeExtensions.cs
+++ b/ClientCore/Enums/ClientTypeExtensions.cs
@@ -1,9 +1,9 @@
 ﻿#nullable enable
 using System;
 
-using ClientCore.Enums;
+using ClientCore.Extensions;
 
-namespace ClientCore.Extensions;
+namespace ClientCore.Enums;
 
 public static class ClientTypeExtensions
 {
--- a/ClientCore/Enums/ClientTypeExtensions.cs
+++ b/ClientCore/Enums/ClientTypeExtensions.cs
@@ -36,7 +36,6 @@ public static class ClientTypeExtensions
             ClientType.RA => 2229840,
             ClientType.TS => 2229880,
             ClientType.YR or ClientType.Ares => 2229850,
-            // You may return null if a new ClientType does not have a corresponding Steam App ID
             _ => null,
         };
     }
--- a/ClientCore/Enums/ClientTypeExtensions.cs
+++ b/ClientCore/Enums/ClientTypeExtensions.cs
@@ -37,8 +37,7 @@ public static class ClientTypeExtensions
             ClientType.TS => 2229880,
             ClientType.YR or ClientType.Ares => 2229850,
             // You may return null if a new ClientType does not have a corresponding Steam App ID
-            // Throw on unexpected ClientType
-            _ => throw new Exception(unknownClientTypeErrorMsg),
+            _ => null,
         };
     }
 }
--- a/DXMainClient/Startup.cs
+++ b/DXMainClient/Startup.cs
@@ -167,21 +167,8 @@ namespace DTAClient
             {
                 try
                 {
-                    if (ClientConfiguration.Instance.ClientGameType == ClientType.Ares || ClientConfiguration.Instance.ClientGameType == ClientType.YR)
-                    {
-                        Logger.Log("Steam init called");
-                        SteamClient.Init(2229850);
-                    }
-                    else if (ClientConfiguration.Instance.ClientGameType == ClientType.TS)
-                    {
-                        Logger.Log("Steam init called");
-                        SteamClient.Init(2229880);
-                    }
-                    else if (ClientConfiguration.Instance.ClientGameType == ClientType.RA)
-                    {
-                        Logger.Log("Steam init called");
-                        SteamClient.Init(2229840);
-                    }
+                    Logger.Log("Steam init called");
+                    SteamClient.Init(ClientTypeHelper.ToSteamAppId(ClientConfiguration.Instance.ClientGameType));
                 }
                 catch (System.Exception e)
                 {
--- a/DXMainClient/Startup.cs
+++ b/DXMainClient/Startup.cs
@@ -21,7 +21,6 @@ using System.Runtime.Versioning;
 using ClientCore.Settings;
 using ClientGUI;
 using Steamworks;
-using ClientCore.Extensions;
 
 namespace DTAClient
 {
--- a/DXMainClient/Startup.cs
+++ b/DXMainClient/Startup.cs
@@ -166,10 +166,14 @@ namespace DTAClient
         {
             if (UserINISettings.Instance.SteamIntegration)
             {
+                uint? steamAppId = ClientConfiguration.Instance.ClientGameType.ToSteamAppId();
+                if (!steamAppId.HasValue)
+                    return;
+
                 try
                 {
                     Logger.Log("Steam init called");
-                    SteamClient.Init(ClientConfiguration.Instance.ClientGameType.ToSteamAppId());
+                    SteamClient.Init(steamAppId.Value);
                 }
                 catch (System.Exception e)
                 {
--- a/DXMainClient/Startup.cs
+++ b/DXMainClient/Startup.cs
@@ -21,6 +21,7 @@ using System.Runtime.Versioning;
 using ClientCore.Settings;
 using ClientGUI;
 using Steamworks;
+using ClientCore.Extensions;
 
 namespace DTAClient
 {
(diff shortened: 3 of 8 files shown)


SadPencil:
Please fix the build failure first

DXMainClient/Startup.cs
```diff
@@ -167,21 +167,18 @@ private void InitSteamworks()
...
+                        ClientType.TD => 2229830,
+                        ClientType.RA => 2229840,
+                        ClientType.TS => 2229880,
+                        ClientType.YR => 2229850,
+                        ClientType.Ares => 2229850,
+                        _ => null
+                    };
+
+                    Logger.Log("Steam init called");
+                    SteamClient.Init(steamAppId.Value);
```
SadPencil, line 175: when steamAppId is null, the behavior seems to be different compared with without this PR

MahBoiDeveloper: > when steamAppId is null, the behavior seems to be different compared with without this PR

It should crash. I thought it is Ok, If next time any developer would add new client game type but forgot about steam init part.

SadPencil: Why crashing it? The old client just do not call steam init in this situation

> If next time any developer would add new client game type but forgot about steam init part.

Didn't get the point. A silent try-catch block cannot remind the developer to add a steam app id

SadPencil:
The PR diff does not match the PR title now. What's changed? Is there any breaking change?

SadPencil:
Please fill the PR descrption using https://github.com/CnCNet/xna-cncnet-client/blob/develop/.github/PULL_REQUEST_TEMPLATE/misc.md template. From my understanding, there are two breaking changes introduced:
1. The client no longer supports YR legacy branch
2. Modders specifying Ares as the game type will get a crash telling them they need to follow a v2.12 change

MahBoiDeveloper:
Sorry for such ADHD changes in PR description and objectives. But these changes have to be done, because we have a technical debt from YRC migration to use Ares and Phobos, and I need TD clause in ClientType for TDC repo in next release.

SadPencil:
> we have a technical debt from YRC migration to use Ares and Phobos

Why not specifying `Ares` in CnCNet YR?

> and I need TD clause in ClientType for TDC repo in next release.

I recommend a simple PR that only adds TD, without touching anything else

MahBoiDeveloper:
> > we have a technical debt from YRC migration to use Ares and Phobos
> 
> Why not specifying `Ares` in CnCNet YR?
> 
> > and I need TD clause in ClientType for TDC repo in next release.
> 
> I recommend a simple PR that only adds TD, without touching anything else

Because there is no reason to do so if all clients for YR (mods included) use Ares nowdays.

SadPencil:
> > > we have a technical debt from YRC migration to use Ares and Phobos
> > 
> > 
> > Why not specifying `Ares` in CnCNet YR?
> > > and I need TD clause in ClientType for TDC repo in next release.
> > 
> > 
> > I recommend a simple PR that only adds TD, without touching anything else
> 
> Because there is no reason to do so if all clients for YR (mods included) use Ares nowdays.

I still maintain the legacy branch. For example, the v2.13.3 version fixes an important bug to prevent adversaries crashing all the clients. The YR legacy branch benifits from this update.

ClientCore/Enums/ClientType.cs
```diff
@@ -1,10 +1,10 @@
...
     {
-        TS,
-        YR,
-        Ares,
-        RA,
+        TD = 2229830,
+        RA = 2229840,
+        TS = 2229880,
+        YR = 2229850,
     }
```
SadPencil: Besides, this modification assumes each ClientType must corresponds a unique Steam app id, which is not always the case. For example, maybe, in the future, I could add an OpenTS type that shares the same steam app id with TS

SadPencil:
Above all, I recommend proposing a simple PR that only adds TD, without touching anything else. Then, for any other changes including any structural changes, open a new issue/PR and have a discussion

MahBoiDeveloper:
Docs updated

MahBoiDeveloper:
> > > > we have a technical debt from YRC migration to use Ares and Phobos
> > > 
> > > 
> > > Why not specifying `Ares` in CnCNet YR?
> > > > and I need TD clause in ClientType for TDC repo in next release.
> > > 
> > > 
> > > I recommend a simple PR that only adds TD, without touching anything else
> > 
> > 
> > Because there is no reason to do so if all clients for YR (mods included) use Ares nowdays.
> 
> I still maintain the legacy branch. For example, the v2.13.3 version fixes an important bug to prevent adversaries crashing all the clients. The YR legacy branch benifits from this update.

This PR doesn't change YR legacy branch behaviour from the current state

ClientCore/Enums/ClientType.cs
MahBoiDeveloper: > Besides, this modification assumes each ClientType must corresponds a unique Steam app id, which is not always the case. For example, maybe, in the future, I could add an OpenTS type that shares the same steam app id with TS

I suppose TS client with ts-patches, or with Vinifera, or OpenTS build wouldn't be so different in the future updates. Also OpenTS have planes to port all ts-patches and Vinifera features into main branch, so the future TSC would build with OpenTS, I believe.

SadPencil:
1. Even if you want to merge YR with Ares, we have only 1 repo specifying the game type as YR, while there are many mods using the game type Ares, then, why use the name YR instead of Ares? And why not let the client be compatible with both names?

2. Even if currently there might be no need to duplicate a game type for OpenTS, still, your modification adds unnecessary restriction such that each game type must have a unique steam id, disallowing a game does not have a steam id or shares the same steam id. For example a modder may choose to fork our client, and the ideal way to add its own feature is, adding a new game type -- this is better than directly modifying codes.

On the other hand, it is less straightforward to write `(uint)ClientConfiguration.Instance.ClientGameType` to fetch the Steam app id -- the statement does not imply anything related with Steam. A dictionary is better. And allow skipping initalizing Steam when app id is missing.

MahBoiDeveloper:
> 1. Even if you want to merge YR with Ares, we have only 1 repo specifying the game type as YR, while there are many mods using the game type Ares, then, why use the name YR instead of Ares? And why not let the client be compatible with both names?

Sounds reasonable. Ok, I'll rework the PR

ClientCore/Enums/ClientTypeExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+{
+    private static readonly string unknownClientTypeErrorMsg = string.Format((
+            "It seems the client configuration was not migrated to accommodate for the v2.12 changes. " +
+            "Please specify 'ClientGameType' in '[Settings]' section of the 'ClientDefinitions.ini' file " +
+            "(allowed options: {0}).\n\n" +
+            "Please refer to documentation of the client {1} for more details. This link can also be found in the log file.").L10N("Client:Main:ClientGameTypeNotFoundException"),
+            EnumExtensions.GetNames<ClientType>(),
+            "https://github.com/CnCNet/xna-cncnet-client/");
+
+    extension(ClientType)
```
SadPencil, line 18: @MahBoiDeveloper meet the new C# 14 syntax. It allows you to use `ClientType.FromString()`. It's awesome

SadPencil (approved):

ClientCore/Enums/ClientTypeExtensions.cs
MahBoiDeveloper: > @MahBoiDeveloper meet the new C# 14 syntax. It allows you to use `ClientType.FromString()`. It's awesome

Yes, I've read about it in MS's article, but XNA client not fully migrated to it, so I've decided to make extension like in good old days.

SadPencil: > but XNA client not fully migrated to it

Are you sure? We are using C# 14 for a relatively long time

MahBoiDeveloper: image

SadPencil: they need another refactor, which can either be refactored in another PR now (since they are straightforward) or delayed until someone make changes there

plus, the ClientGameTypeExtensions is very different from other extensions -- it's not actually an extension, but a, workaround for cannot defining methods for an enum. Therefore it should lies in the same namespace as ClientGameType, rather than the Extensions namespace which contains extension classes for *external* classes. Based on this consideration, it's more natural to use C# 14 syntax here, as if the enum supports new methods

DXMainClient/Startup.cs
```diff
@@ -165,23 +165,14 @@ private void InitSteamworks()
...
-                        Logger.Log("Steam init called");
-                        SteamClient.Init(2229880);
-                    }
-                    else if (ClientConfiguration.Instance.ClientGameType == ClientType.RA)
-                    {
-                        Logger.Log("Steam init called");
-                        SteamClient.Init(2229840);
-                    }
+                    Logger.Log("Steam init called");
+                    SteamClient.Init(steamAppId.Value);
```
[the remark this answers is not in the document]
SadPencil: @MahBoiDeveloper I compiled the client and the test result showed AI is right. Even if you throw an exception in ToSteamAppId method, the release version of the client will act as if the exception does not exist, because the whole Steam method is executed in another thread and of course the main thread cannot wait for it.

Therefore, I'll remove the throw statement in ToSteamAppId method since it's useless

SadPencil (approved):

PR #1084, opened 2026-09-10
Improve V3 tunnel ping metrics and hide P2P IPs

11EJDE11:
Some small fixups for V3 tunnels based on my experience with the latest release and feedback from users.

1) Remove the P2P IP from showing on tooltips. This is for showmatches/streams where you don't want to accidentally broadcast the IP.

2) Update the ping quality tiers. With V3 tunnels, the ping measurement is different than V2. The V2 system would measure ICMP ping to the tunnel. With V3, it is a round-trip from player to player. So the tiers need to be adjusted to account for this, as right now most people are showing with poor ping quality icons when it was fine before. 

3) Adjust the kick suggestion threshold from 300ms to 400ms.

### DIFF
--- a/DXMainClient/Domain/Multiplayer/CnCNet/P2PTunnel.cs
+++ b/DXMainClient/Domain/Multiplayer/CnCNet/P2PTunnel.cs
@@ -1,8 +1,9 @@
 #nullable enable
-using System;
 using System.Net;
 using System.Security.Cryptography;
 using System.Text;
+
+using Rampastring.Tools;
 
 namespace DTAClient.Domain.Multiplayer.CnCNet;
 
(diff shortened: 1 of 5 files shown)


DXMainClient/Domain/Multiplayer/CnCNet/P2PTunnel.cs
```diff
@@ -17,7 +17,7 @@ public class P2PTunnel : CnCNetTunnel
     public override bool IsDirect => true;
 
     public P2PTunnel(IPEndPoint peerEndpoint, string peerName)
-        : base(peerEndpoint.Address.ToString(), peerEndpoint.Port, $"Direct ({peerName} @ {peerEndpoint})", version: 3)
+        : base(peerEndpoint.Address.ToString(), peerEndpoint.Port, $"Direct ({peerName})", version: 3)
```
SadPencil: I don't know whether there is a unique requirement for this `peerName`. If so, I'd like to use

```cs
$"Direct ({peerName} @ {peerEndpointHash})"
```

where

```cs
string salt = "whatever fixed string"
string peerEndpointHash = SHA1($"{salt}|{peerEndpoint}")[0:6]
```

11EJDE11: [c475658](https://github.com/CnCNet/xna-cncnet-client/pull/1084/commits/c47565855ab55da014ee29ae88948654ede0d9ec)

